PR #8, opened 2019-08-22
chore: Add QA and RELEASING documents

msbrogli:


QA.md
```diff
@@ -0,0 +1,77 @@
...
+1. - [ ] Enter a valid address and click Next. You can paste the address copied before.
+1. - [ ] Click on the HTR to change the token. Then, select HTR.
+1. - [ ] Type 100 HTR, click Next, and check the insufficient funds error.
+1. - [ ] Type 2 HTR, click Next, and check the send summary.
+1. - [ ] Click on Send and type a wrong PIN Code.
+1. - [ ] Finally, type a correct PIN Code and check the confirmation.
+
+1. **Transaction History Tests**
+1. - [ ] Open HTR History. Your balance should be 8.00 HTR.
+1. - [ ] Check that you have two transaction.
```
pedroferreira1: transaction**s**

QA.md
```diff
@@ -0,0 +1,77 @@
...
+1. - [ ] Click on the configuration string to copy it.
+1. - [ ] Click on Share and cancel.
+1. - [ ] Close the token details.
+
+1. **Register a token Tests**
+1. - [ ] Go to the Dashboard Screen.
+1. - [ ] Select the Test Token.
+1. - [ ] Click on the Token Info and unregister it.
+1. - [ ] Click on Register Token, click on Manual Info, and paste the Configuration string.
+1. - [ ] Click on Register Token, and check that the Test Token is back and your balance is 100 TEST.
```
pedroferreira1: I think that we have a different token from HTR we should test the creation of a payment request of another token and Send a different token also

QA.md
```diff
@@ -0,0 +1,77 @@
...
+1. - [ ] Click on create token and check the error message.
+1. - [ ] Check that the token has not been created.
+
+1. **Receive Tests**
+1. - [ ] Go to the Receive Screen and check the QR Code.
+1. - [ ] Click on the address to copy it.
+1. - [ ] Click on New address and check that a new address was generated. The QRCode must update.
+1. - [ ] Click on Share and Cancel.
+1. - [ ] Click on Payment Request and generate a payment request of 10 HTR.
+1. - [ ] Pay the Payment Request using another wallet and check the confirmation message.
```
jansegre: The desktop wallet does not support reading QR codes nor payment requests. And this is really hard with a single phone. Should we improve our strategy?

obiyankenobi: You can use the emulator on the desktop.

QA.md
```diff
@@ -0,0 +1,77 @@
...
+
+1. **Send Tests**
+1. - [ ] Go to the Send Screen and check whether the camera loads correctly.
+1. - [ ] Click on Manual Info.
+1. - [ ] Type any random address and click Next. It must show an error message.
```
jansegre: "random text" instead of "any random address", because actual random addresses should not error, just likely burn funds

QA.md
```diff
@@ -0,0 +1,77 @@
...
+1. - [ ] Enter a valid address and click Next. You can paste the address copied before.
+1. - [ ] Click on the HTR to change the token. Then, select HTR.
+1. - [ ] Type 100 HTR, click Next, and check the insufficient funds error.
+1. - [ ] Type 2 HTR, click Next, and check the send summary.
+1. - [ ] Click on Send and type a wrong PIN Code.
+1. - [ ] Finally, type a correct PIN Code and check the confirmation.
+
+1. **Transaction History Tests**
+1. - [ ] Open HTR History. Your balance should be 8.00 HTR.
```
jansegre: Not if I'm transferring funds for myself, which is what happens when following the "You can paste de address copied before" advice.

QA.md
```diff
@@ -0,0 +1,77 @@
...
+1. - [ ] Check that you have two transaction.
+1. - [ ] Click on any transaction and check the details.
+
+1. **Settings Tests**
+1. - [ ] Go to the Settings Screen. Check that you are connected to our `testnet`.
+1. - [ ] Test Security > Lock Wallet.
+  1. - [ ] Enable TouchID.
```
jansegre: Should point out that this is an iOS only step.

QA.md
```diff
@@ -0,0 +1,77 @@
...
+1. - [ ] Click on Register Token, click on Manual Info, and paste the Configuration string.
+1. - [ ] Click on Register Token, and check that the Test Token is back and your balance is 100 TEST.
+
+1. **Reset Wallet**
+1. - [ ] Go to Settings and Reset your wallet.
+
+1. **Import a wallet**
+1. - [ ] Click on Import Wallet.
+1. - [ ] Type anything random and check the "Invalid words" error message.
+1. - [ ] Use a valid seed. For example: `enjoy better loop metal soft liquid express chaos energy slush kid truck before midnight squirrel brick bamboo dolphin donkey aware drill can hazard raise`.
```
jansegre: We could expand this to include a backup step in the beginning and testing restoring the backup now. Making sure that restoring a backup works is probably one of the most important checks IMO.

QA.md
msbrogli: Fixed. Thanks!

QA.md
msbrogli: Fixed. Thanks!

QA.md
msbrogli: You're right. I put a different address in the previous step. Now, it is 8.00 HTR. Thanks!

QA.md
msbrogli: Done. Thanks!

QA.md
pedroferreira1: You say that in the 'Simple Initialization' step we should save the generated words. And then here we import the same words and validate everything is loaded, right? If that's it I agree and will update the QA

QA.md
pedroferreira1: The emulator can't read the generated payment request QRCode. The idea is to try to execute all the actions with the mobile phone but I guess in this case we can test this payment request with the emulator and then read with the phone.

Is it ok for now this @jansegre?

QA.md
pedroferreira1: @msbrogli @obiyankenobi @jansegre do you agree with it? Since I am reviewing this PR now, need some feedback over my old comment.

RELEASING.md
```diff
@@ -0,0 +1,36 @@
...
+# Bumping up the version
+
+The following files must be updated: `android/app/build.gradle`, `ios/HathorMobile/Info.plist`, and `package.json`.
+
+In the `package.json`, the field `version` must be updated.
+
+In the `android/app/build.gradle`, the fields `versionCode` and `versionName` must be updated. The `versionCode` must always be increased because Google Play uses it to uniquelly identify the release.
+
+In the `ios/HathorMobile/Info.plist`, the fields `CFBundleShortVersionString` and `CFBundleVersion` must be updated. Usually, the `CFBundleVersion` is always `1`, while the `CFBundleShortVersionString` is updated with the new version. App Store uses the pair `(CFBundleShortVersionString, CFBundleVersion)` to identify a release.
+
```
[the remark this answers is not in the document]
pedroferreira1: Check, or create?

QA.md
obiyankenobi: Yes, I agree.

msbrogli: I agree.

QA.md
msbrogli: Can't we leave an image with the QR Code that will be used when running the QA?

QA.md
msbrogli: I agree. It is a critical operation that must be test before every release.

RELEASING.md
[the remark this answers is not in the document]
msbrogli: Create.

QA.md
pedroferreira1: I think the idea here is to validate two things: 

1. The QRcode is being correctly generated and if you read it, will be paid.
2. When the payment request is paid, it receives the payment from websocket and shows success on the screen.

If we have a fixed QRCode image neither of the points will be tested. I think for now the best option is to use the emulator.

QA.md
pedroferreira1: Added!

QA.md
pedroferreira1: Added

RELEASING.md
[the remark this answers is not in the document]
pedroferreira1: Added

QA.md
jansegre: I think that requiring the QA tester to run an emulator is a big burden for setting up a testing environment.

Personally I have somewhat worked this around another way, my phone has a dual app feature that enables two independent instances of the same app, I use that in addition to taking a picture with my webcam of the qr code while I switch the app instances. It is a complicated setup that only works because of the dual app feature.

This step is testing two "features" at once:

- Being able to read and execute payment requests. A pre-generated image would cover this. But this would be a good additional step to the **Send Tests**.
- Validating the generated payment requests. The pre-generated image cannot test this.

I think it's good that we test executing the payment request with a different (but working) version of an existing wallet. This step doesn't specify this, but if we were to test using the same version (like I do on my setup, or maybe ends up happening on the emulator) we could run into a problem where we introduce a breaking change to the qr code data format which breaks older versions but the version being testes can read the qr code it generated. This would only be caught if the "other" wallet is an older version.

I don't have a good suggestion, but I have a workaround for this, so I guess it's fine.

msbrogli: I think it is good for now. We can improve it later if necessary. Maybe we should consider that the QA must have access to two mobile devices to test them properly.

PR #9, opened 2019-08-26
feat: token deposit

obiyankenobi:
image
image
image
image
image
image

### DIFF
--- a/src/screens/CreateTokenDepositNotice.js
+++ b/src/screens/CreateTokenDepositNotice.js
@@ -13,6 +13,7 @@ import {
   View,
 } from 'react-native';
 
+import hathorLib from '@hathor/wallet-lib';
 import NewHathorButton from '../components/NewHathorButton';
 import HathorHeader from '../components/HathorHeader';
 import baseStyle from '../styles/init';
@@ -37,8 +38,11 @@ class CreateTokenDepositNotice extends React.Component {
         <View style={this.style.container}>
           <View>
             <Text style={this.style.text}>
-              When creating new tokens, a <Text style={this.style.link}>deposit of {1}%</Text> in
-              HTR is required. So if you create 1000 NewCoins, 10 HTR are needed as deposit.
+              When creating new tokens, a{' '}
+              <Text style={this.style.link}>
+                deposit of {hathorLib.tokens.getDepositPercentage() * 100}%{' '}
+              </Text>
+              in HTR is required - e.g. if you create 1000 NewCoins, 10 HTR are needed as deposit.
             </Text>
             <Text style={this.style.text}>
               If these tokens are later melted, the HTR deposit will be returned. Read more about it{' '}
--- a/src/screens/CreateTokenDepositNotice.js
+++ b/src/screens/CreateTokenDepositNotice.js
@@ -1,0 +1,61 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import {
+  SafeAreaView,
+  StyleSheet,
+  Text,
+  View,
+} from 'react-native';
+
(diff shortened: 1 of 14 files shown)


pedroferreira1:
Still missing error screenshot

src/screens/CreateTokenDepositNotice.js
```diff
@@ -0,0 +1,66 @@
...
+            <Text style={this.style.text}>
+              When creating new tokens, a{' '}
+              <Text style={this.style.link}>
+                deposit of {hathorLib.tokens.getDepositPercentage() * 100}%{' '}
+              </Text>
+              in HTR is required - e.g. if you create 1000 NewCoins, 10 HTR are needed as deposit.
+            </Text>
+            <Text style={this.style.text}>
+              If these tokens are later melted, the HTR deposit will be returned. Read more about it{' '}
+              <Link href='https://gitlab.com/HathorNetwork/rfcs/blob/master/text/0011-token-deposit.md'>here</Link>
```
msbrogli: Move the URL to the config file.

src/screens/CreateTokenName.js
```diff
@@ -46,11 +48,19 @@ class CreateTokenName extends React.Component {
...
-            <NewHathorButton
-              title='Next'
-              disabled={!this.state.name}
-              onPress={this.onButtonPress}
-            />
+            <View>
+              <InfoBox
+                items={[
+                  <Text>Token name should be the full name of the new token you are creating</Text>,
+                  <Italic>E.g. Hathor</Italic>
```
msbrogli: I would use another example. Something with two words and a space. Maybe "My Custom Token" or "Hathor Network".

src/screens/CreateTokenSymbol.js
```diff
@@ -38,28 +39,49 @@ class CreateTokenSymbol extends React.Component {
...
   }
 
+  isButtonDisabled = () => {
+    if (!this.state.symbol) {
+      return true;
+    }
+
+    if (this.state.symbol.length < 2) {
+      return true;
+    }
```
msbrogli: What about the max length of 5?

src/screens/CreateTokenName.js
obiyankenobi: The problem is that you're leading the person to use that model. I prefer without spaces: 'YanCoin' or 'MyCustomToken'. I used a single word to make sure we don't prime the user with a preference.

We can also give 2 examples, with and without space.

src/screens/CreateTokenSymbol.js
obiyankenobi: We already forbid that in the input field, bu I'll change to also enforce here.

src/screens/CreateTokenName.js
msbrogli: Seems good. Examples: "MyCustomToken", "My Custom Token", etc.

src/screens/CreateTokenName.js
```diff
@@ -46,11 +48,19 @@ class CreateTokenName extends React.Component {
               onChangeText={this.onNameChange}
               value={this.state.name}
```
pedroferreira1: Should we validate that this input has at most 30 chars? Just to keep it consistent with the full node validation

obiyankenobi: That's true, we're not enforcing it here. Will do it.

obiyankenobi: Did that and added a 'counter' for the name:
image

src/screens/CreateTokenSymbol.js
obiyankenobi: Done

src/screens/CreateTokenName.js
obiyankenobi: Done

src/screens/CreateTokenDepositNotice.js
obiyankenobi: Done

PR #11, opened 2019-09-22
feat: validate token info when registering it

pedroferreira1:


src/screens/RegisterTokenManual.js
```diff
@@ -57,12 +57,14 @@ class RegisterTokenManual extends React.Component {
       return;
     }
 
-    const ret = hathorLib.tokens.validateTokenToAddByConfigurationString(this.state.configString);
-    if (ret.success) {
-      this.setState({ token: ret.tokenData, errorMessage: '' });
-    } else {
-      this.setState({ errorMessage: ret.message });
-    }
+    const promise = hathorLib.tokens.validateTokenToAddByConfigurationString(
```
obiyankenobi: I wonder if we should show a spinner, as this calls the network and could take a bit. Otherwise, there's no feedback to the user.

pedroferreira1: Makes sense, added the loading.

PR #17, opened 2019-11-12
fix: enabling register token button only after api request is completed

pedroferreira1:


pedroferreira1:
Closes https://github.com/HathorNetwork/hathor-wallet-mobile/issues/10

Issue #19, opened 2019-11-12
iOS 13 dark mode forces status bar to be white. We should set it to black


pedroferreira1:
![File](https://user-images.githubusercontent.com/3298774/68772551-c7370c00-0608-11ea-9ad5-a3c533521a65.jpg)

PR #20, opened 2019-11-13
fix: setting status bar style to dark content by default

pedroferreira1:


pedroferreira1:
Closes https://github.com/HathorNetwork/hathor-wallet-mobile/issues/19

PR #25, opened 2019-12-04
Style/new brand

pedroferreira1:
image
image
image
image
image
image


src/components/TokenSelect.js
```diff
@@ -33,7 +34,7 @@ const TokenSelect = (props) => {
       <TouchableHighlight
         style={index === 0 ? styles.firstItemWrapper : null}
         onPress={() => { props.onItemPress(item); }}
-        underlayColor='rgba(227, 0, 82, 0.5)'
+        underlayColor='rgba(140, 70, 255, 0.3)'
```
msbrogli: Isn't it better to include as a constant as well?

src/screens/MainScreen.js
```diff
@@ -408,13 +409,13 @@ class BalanceView extends React.Component {
       marginBottom: 24,
     },
     networkView: {
-      backgroundColor: 'rgba(227, 0, 82, 0.1)',
+      backgroundColor: 'rgba(140, 70, 255, 0.1)',
```
msbrogli: Isn't it better to include it as a constant? It seems to be the same as in TokenSelect.js, isn't it?

src/screens/Settings.js
```diff
@@ -48,26 +50,26 @@ export class Settings extends React.Component {
       marginBottom: 24,
     },
     networkView: {
-      backgroundColor: 'rgba(227, 0, 82, 0.1)',
+      backgroundColor: 'rgba(140, 70, 255, 0.1)',
```
msbrogli: Again, the same color here.

src/components/TokenSelect.js
pedroferreira1: Created a method in constants to get this color depending on the alpha value.

src/screens/Settings.js
pedroferreira1: Created a method in constants to get this color depending on the alpha value.

src/screens/MainScreen.js
pedroferreira1: Created a method in constants to get this color depending on the alpha value.

PR #27, opened 2019-12-30
feat: prepare app to mainnet

pedroferreira1:
- Fix android app name
- Update initial screen
- Change address network byte
- Update server to mainnet
- Update wallet lib version

image

### DIFF
--- a/src/screens/InitWallet.js
+++ b/src/screens/InitWallet.js
@@ -58,7 +58,8 @@ class WelcomeScreen extends React.Component {
               This wallet is connected to the <Strong>mainnet</Strong>.
             </Text>
             <Text style={this.style.text}>
-              A mobile wallet is not the safest place to store your tokens. So, we advise you to keep only a small amount of tokens here, such as pocket money.
+              A mobile wallet is not the safest place to store your tokens.
+              So, we advise you to keep only a small amount of tokens here, such as pocket money.
             </Text>
             <Text style={this.style.text}>
               For further information, check our website{' '}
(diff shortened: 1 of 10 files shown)


src/screens/InitWallet.js
```diff
@@ -52,17 +52,13 @@ class WelcomeScreen extends React.Component {
...
-              This wallet is connected to a <Strong>testnet</Strong>.
+              This wallet is connected to the <Strong>mainnet</Strong>.
             </Text>
             <Text style={this.style.text}>
-              This means that{' '}
-              <Strong>your Hathor token (HTR) and any other token may be reset at any time.</Strong>
-            </Text>
-            <Text style={this.style.text}>
-              If someone offers to sell some tokens to you, that person is a scammer.
+              A mobile wallet is not the safest place to store your tokens, so we advise you to have only a small amount of them here.
```
msbrogli: Suggestion: A mobile wallet is not the safest place to store your tokens. So, we advise you to keep only a small amount of tokens here, such as pocket money.

msbrogli: Maybe we should include a page in our website about Token Custody to a longer explanation about how to store tokens [(hardware wallet, software wallet) x (desktop wallet, mobile wallet)]. What do you think?

obiyankenobi: I'd open an issue for this, as we don't have this page yet.

pedroferreira1: I agree with @obiyankenobi that we should do it later, maybe even a blog post would be better than a page on the website. I will create an issue for this.

I've update the text as you suggested.

PR #29, opened 2020-01-02
feat: forcing reset wallet when it was already loaded before for the …

pedroferreira1:
…testnet

### DIFF
--- a/src/screens/LoadHistoryScreen.js
+++ b/src/screens/LoadHistoryScreen.js
@@ -18,7 +18,7 @@ import { loadHistory, clearInitWallet } from '../actions';
 import {
   Strong, setSupportedBiometry, getSupportedBiometry, setBiometryEnabled, isBiometryEnabled,
 } from '../utils';
-import { KEYCHAIN_USER } from '../constants';
+import { KEYCHAIN_USER, NETWORK } from '../constants';
 import SimpleButton from '../components/SimpleButton';
 import Spinner from '../components/Spinner';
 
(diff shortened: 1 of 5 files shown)


src/screens/LoadHistoryScreen.js
```diff
@@ -113,6 +113,7 @@ class LoadHistoryScreen extends React.Component {
     }
     hathorLib.WebSocketHandler.on('addresses_loaded', this.addressesLoadedUpdate);
     this.cleanData();
+    hathorLib.storage.setItem('wallet:network', NETWORK);
```
msbrogli: I think we shouldn't set it in the `LoadHistoryScreen`. Maybe this can be verified and set in a single place. It must also be set in the last screen of the setup screens. What do you think?

pedroferreira1: I was doing this but the line above this is doing a `cleanData`. It clears all storage data, so I can't set it before, like I was doing and, since `LoadHistoryScreen` might be called many times during the use of the app (e.g. in case we lose connection to the ws) doing only one time wouldn't work.

I agree that clearing everything might not be the perfect solution, so we should change this behaviour before but I feel it's okay to continue with this for now.

obiyankenobi: I agree, we need a larger refactor. For now, it's ok for me.

msbrogli:
We cannot merge it as it is. I'm closing this pull request for now.

PR #31, opened 2020-01-07
feat(i18n): Add i18n support (ttag lib) and support for pt-br

msbrogli:
Screen Shot 2020-01-06 at 21 12 11


i18nInit.js
```diff
@@ -0,0 +1,20 @@
...
+  // https://stackoverflow.com/a/47349998
+  const deviceLanguage =
+        Platform.OS === 'ios'
+          ? NativeModules.SettingsManager.settings.AppleLocale ||
+            NativeModules.SettingsManager.settings.AppleLanguages[0] //iOS 13
+          : NativeModules.I18nManager.localeIdentifier;
+}
+
+const availableLocales = ['pt-br'];
+//const locale = getLocale();
```
msbrogli: fix: getLocale() is not working

msbrogli: Fixed.

msbrogli:
Maybe we should let the use choose the language. In this case, there would be a Language choice in both `Settings.js` and `InitWallet.js`; and, if the user has chosen a language, it must supersede OS's language. What do you think? @jansegre @obiyankenobi @pedroferreira1

pedroferreira1:
> Maybe we should let the use choose the language. In this case, there would be a Language choice in both `Settings.js` and `InitWallet.js`; and, if the user has chosen a language, it must supersede OS's language. What do you think? @jansegre @obiyankenobi @pedroferreira1

I think it's good for the user to choose their language but it would be something extra. Only having the OS language would already be ok but I guess it's not the complicated to add this.

locale/pt-br/texts.po
```diff
@@ -0,0 +1,62 @@
...
+msgid "No transactions"
+msgstr "Sem transações"
+
+#: src/screens/MainScreen.js:101
+msgid "Share your address"
+msgstr "Compartilhe seu endereço"
+
+#: src/screens/MainScreen.js:104
+msgid "with friends and start exchanging tokens"
+msgstr "com seus amigos e começe a trocar seus tokens"
```
pedroferreira1: comece*

msbrogli: Fixed. Thanks.

src/components/TextFmt.js
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import React from 'react';
+import { Text } from 'react-native';
+import { Strong } from '../utils';
+
+const TextFmt = (props) => {
```
obiyankenobi: Would be good to have a comment explaining this component.

locale/pt-br/texts.po
```diff
@@ -0,0 +1,734 @@
...
+msgstr "Endereço inválido"
+
+#: src/screens/About.js:43
+msgid "ABOUT"
+msgstr "SOBRE"
+
+#: src/screens/About.js:54
+msgid "This app is developed by Hathor Labs and is distributed for free."
+msgstr ""
+"Este aplicativo foi desenvolvido pela Hathor Labs e é distribuído de graça."
```
obiyankenobi: Why the empty string above?

locale/pt-br/texts.po
```diff
@@ -0,0 +1,734 @@
...
+
+#: src/screens/PinScreen.js:172 src/screens/Settings.js:128
+msgid "Reset wallet"
+msgstr "Resetar a wallet"
+
+#. *
+#.      * index {number} Selected index of the tab bar
+#.      
+#.  eslint thinks routes is not used, but TabView uses it
+#.  eslint-disable-next-line react/no-unused-state
```
obiyankenobi: What are these comments above?

locale/pt-br/texts.po
```diff
@@ -0,0 +1,734 @@
...
+msgid_plural "${ amountAndToken } available"
+msgstr[0] "${ amountAndToken } disponível"
+msgstr[1] "${ amountAndToken } disponíveis"
+
+#: src/screens/SendAmountInput.js:126 src/screens/SendConfirmScreen.js:133
+msgid "SEND ${ tokenNameUpperCase }"
+msgstr "ENVIAR ${ tokenNameUpperCase }"
+
+#.  show loading modal
+#.  eslint-disable-next-line react/jsx-indent
```
obiyankenobi: Why is this needed?

jansegre:
> Maybe we should let the use choose the language. In this case, there would be a Language choice in both `Settings.js` and `InitWallet.js`; and, if the user has chosen a language, it must supersede OS's language. What do you think? @jansegre @obiyankenobi @pedroferreira1

At least on Android I'm pretty used for apps that use the system language and offer no other choice. In fact I don't recall any app that gives me the option to choose.

src/components/TextFmt.js
msbrogli: Done. Thanks!

locale/pt-br/texts.po
msbrogli: In this case, it wasn't needed. Fixed. Thanks.

locale/pt-br/texts.po
msbrogli: I'm sure why the `npm run locale-update-pot` decided to get these comments from the source-code.

locale/pt-br/texts.po
msbrogli: The same here. I'm not sure why it was collected as a note for translators.

locale/pt-br/texts.po
obiyankenobi: Ah ok, this is auto generated. Weird.

PR #34, opened 2020-01-08
fix(init): Import wallet accepts both space and new line as separators

msbrogli:


src/screens/InitWallet.js
```diff
@@ -245,8 +245,9 @@ class LoadWordsScreen extends React.Component {
   }));
 
   onChangeText = (text) => {
-    const words = text.trim().split(' ');
+    const words = text.trim().split(/[ \n\t]/);
```
jansegre: Why not go all the way then, and accept any whitespace with `\s`.

msbrogli: Good one. Fixed. Thanks!

PR #42, opened 2020-01-27
Bump up to v0.8.3

msbrogli:


msbrogli:
> Now that there is this `MARKETING_VERSION` we should update the `RELEASING.md` file, I think.
> 
> > In the `ios/HathorMobile/Info.plist`, the fields `CFBundleShortVersionString` and `CFBundleVersion` must be updated. Usually, the `CFBundleVersion` is always `1`, while the `CFB    undleShortVersionString` is updated with the new version. App Store uses the pair `(CFBundleShortVersionString, CFBundleVersion)` to identify a release.
> 
> It's not correct anymore

@pedroferreira1 Updated. Can you check again, please?

PR #49, opened 2020-03-17
fix: xpub missing after update

obiyankenobi:


pedroferreira1:
Should we start our QA loading a wallet that was already started in the old app version?

obiyankenobi:
> Should we start our QA loading a wallet that was already started in the old app version?

Added an extra step on QA.

PR #55, opened 2020-04-13
refactor/essential-methods: move methods from screen/Dashboard.js to App.js

obiyankenobi:
Several essential methods for the wallet were being executed in `src/screens/Dashboard.js`, even though they were not related to this screen.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -8,18 +8,23 @@
 import '../shim';
 
 import React from 'react';
-import { StyleSheet, View } from 'react-native';
+import { AppState, StyleSheet, View } from 'react-native';
 import {
   createBottomTabNavigator, createStackNavigator, createSwitchNavigator, createAppContainer,
 } from 'react-navigation';
 import { Provider, connect } from 'react-redux';
 
+import * as Keychain from 'react-native-keychain';
+
 import hathorLib from '@hathor/wallet-lib';
 import IconTabBar from './icon-font';
-import { HATHOR_COLOR } from './constants';
-import { updateHeight } from './actions';
-
+import { HATHOR_COLOR, LOCK_TIMEOUT } from './constants';
+import { setSupportedBiometry } from './utils';
+import {
+  activateFetchHistory, newTx, resetData, setIsOnline, lockScreen, updateHeight, setTokens
+} from './actions';
 import { store } from './reducer';
+
 import DecideStackScreen from './screens/DecideStackScreen';
 import {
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -227,14 +279,77 @@ export class _AppStackWrapper extends React.Component {
...
+    if (nextAppState === 'active') {
+      if (this.appState === 'inactive') {
+        // inactive state means the app wasn't in background, so no need to lock
+        // the screen. This happens when user goes to app switch view or maybe is
+        // asked for fingerprint or face if
+        this.backgroundTime = null;
+      } else if (Date.now() - this.backgroundTime > LOCK_TIMEOUT) {
+        // this means app was in background for more than LOCK_TIMEOUT seconds,
+        // so display lock screen
+        this.props.lockScreen();
```
pedroferreira1, line 333: If the app gets inactive before starting the wallet, what this `lockScreen=true` will do? Will it redirect to the PinScreen? 

I believe one of the reasons this was in the `Dashboard` is because we should lock the screen only after the user has started the wallet. So we just need to test if this doesn't bring any problems.

We have a variable in the lib of `wallet:loaded`, that could help here.

obiyankenobi: This code is inside `AppStackWrapper`, which in turn is in a switch navigator:
```js
const SwitchNavigator = createSwitchNavigator({
  Decide: DecideStackScreen,
  App: AppStackWrapper,
  Init: InitStack,
}, {
  initialRouteName: 'Decide',
});
```

So it will only be executed after we're past the `InitStack`, when the user creates the wallet. If we reset the wallet, `AppStackWrapper` is removed and the unmount method called, so we remove the listeners.

I thought about placing `AppStackWrapper` on a different file, but it depends on other objects from `App.js` (like `AppStack`), so I couldn't find a good way.

obiyankenobi: I could create `AppStack` on a different file:
```js
const AppStack = createStackNavigator({
  Main: TabNavigator,
  About,
  Security,
  ChangePin,
  ResetWallet,
  PaymentRequestDetail,
  RegisterToken: RegisterTokenStack,
  ChangeToken,
  PinScreen: {
    screen: PinScreen,
    navigationOptions: disableSwipeDown,
  },
  CreateTokenStack,
  TokenDetail,
  UnregisterToken,
}, {
  mode: 'modal',
  headerMode: 'none',
});
```

I'm just not sure if that's best or we should keep all navigation code on `App.js`.

pedroferreira1: Ah, it makes sense. I was just confused with all stacks. But it's ok for me

PR #56, opened 2020-04-16
feat: support "whitelabel"

obiyankenobi:
These changes make it much easier for anyone to create an app for their own token. They just need to change some variables on `constants.js`, texts and images.

image
image


src/screens/SendAmountInput.js
```diff
@@ -118,6 +119,10 @@ class SendAmountInput extends React.Component {
       return ngettext(msgid`${amountAndToken} available`, `${amountAndToken} available`, available);
     };
 
+    const renderGhostElement = () => (
```
msbrogli, line 122: Shouldn't we create a component `GhostElement`? Or is it too specific for each screen?

src/screens/SendAmountInput.js
```diff
@@ -133,22 +138,17 @@ class SendAmountInput extends React.Component {
...
               >
-                <View style={{ width: 80, height: 40 }} />
+                {renderGhostElement()}
                 <AmountTextInput
                   ref={this.inputRef}
                   autoFocus
                   onAmountUpdate={this.onAmountChange}
                   value={this.state.amount}
                 />
-                <View style={{
```
msbrogli, line 143: What about this view?

src/screens/SendAmountInput.js
obiyankenobi: I just used it in two different screens. They're specific in the sense that each needs a particular size. In the case where they were used, both have the same size. I'm not sure to be honest if we need a component for such a simple element, since it's only used twice so far.

obiyankenobi:
> Can you attach some screenshots here when `IS_MULTI_TOKEN = false`, please?

Done

pedroferreira1:
Should we add a section in the QA to check all the screens (I would suggest only validate if the elements are ok, without executing any actions).

obiyankenobi:
> Should we add a section in the QA to check all the screens (I would suggest only validate if the elements are ok, without executing any actions).

Done

src/config.js
```diff
@@ -0,0 +1,40 @@
...
+ * LICENSE file in the root directory of this source tree.
+ */
+
+import hathorLib from '@hathor/wallet-lib';
+
+/**
+ * This file contains all variables a user has to modify if he wants to create
+ * an app for his own token. Files should never import them directly (they are
+ * prepended with '_'). Instead, constants.js imports and re-exports them.
+ */
```
obiyankenobi, line 14: @pedroferreira1 added the variables here. I kept `INITIAL_TOKENS` and `LOCK_TIMEOUT` on the other file. What do you think?

pedroferreira1: Good!

src/screens/SendAmountInput.js
msbrogli: I agree. They seem very specific to each screen.

src/screens/SendAmountInput.js
msbrogli: Really? Because it had many styles in it. But, if it was not necessary, you can mark it as resolved.

obiyankenobi: I imagine it was something we had before we refactored the element into `TokenBox`.

OttavodevN:
how can i change the image in  src/assets/image/logo.png
                                                   src/assets/image/logo.png@2x.png
                                                   src/assets/image/logo.png@3x.png ?

I cant find a place where i can download my own new logo for my wallet...

PR #57, opened 2020-04-20
fix(history): Loading history was too slow

### LINKED ISSUE #54: Spike: Wallet is taking too long to open after downloading all transactions
As described on https://github.com/HathorNetwork/hathor-wallet-mobile/issues/52 in the ios with 20k transactions, it takes the app 3 minutes to download all the transactions and 50 more minutes to open the wallet just to process the data.

msbrogli:
Closes #54: Wallet is taking too long to open after downloading all transactions

The problem was an `O(n^2)` algorithm, where `n` is the number of transactions in the history.

### DIFF
--- a/src/reducer.js
+++ b/src/reducer.js
@@ -319,7 +319,6 @@ const onFetchHistorySuccess = (state, action) => {
   const { history } = action.payload;
   const { addresses } = action.payload;
   const tokensHistory = {};
-  const tokensBalance = {};
   // iterate through all txs received and map all tokens this wallet has, with
   // its history and balance
   for (const tx of Object.values(history)) {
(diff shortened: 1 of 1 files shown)


src/reducer.js
```diff
@@ -334,12 +334,17 @@ const onFetchHistorySuccess = (state, action) => {
...
       tokenHistory.push(getTxHistoryFromTx(tx, tokenUid, tokenTxBalance));
-      const totalBalance = getBalance(tokenUid);
-      // update token total balance
-      tokensBalance[tokenUid] = totalBalance;
+      tokenUidSeen[tokenUid] = 1;
     }
   }
 
+  const tokensBalance = {};
+  for (const tokenUid in tokenUidSeen) {
```
obiyankenobi, line 342: Do we need `tokenUidSeen`? Isn't it the same as the keys of `tokenHistory`?

If we do need, maybe a `set` is better than a `dict`.

msbrogli: You're right; we can use `tokensHistory`. I haven't used it before because we don't have to update the balance of all tokens, but it always start empty, so we can safely use it. Can you review again, please?

pedroferreira1:
I did the same test as described in #54 with 20k txs and the wallet took less than 5 seconds to open after downloading all transactions (instead of 50 minutes).

PR #58, opened 2020-04-28
fix: increasing sqlite db cursor and size limit

pedroferreira1:


msbrogli:
@jansegre tested this solution in his Android, and it worked fine.

android/app/src/main/java/com/hathormobile/MainApplication.java
```diff
@@ -65,5 +67,15 @@ public ReactNativeHost getReactNativeHost() {
   public void onCreate() {
     super.onCreate();
     SoLoader.init(this, /* native exopackage */ false);
+
+    try {
```
msbrogli, line 71: I think we should include the link to the issue in react-native, and also leave a comment about what it is fixing.

pedroferreira1: Done

PR #61, opened 2020-05-28
feat: using tx mining API to send transactions

pedroferreira1:
It's still a draft because there are some things to be done:

- [x] Add comments and docstring;
- [x] Handle error case in create token (still need work on the lib);
- [x] Fix linter


src/screens/SendConfirmScreen.js
```diff
@@ -59,21 +58,35 @@ class SendConfirmScreen extends React.Component {
...
-      this.onSuccess, this.onError
-    );
+    const ret = this.props.sendTx(this.amount, this.address, this.token, pinCode);
+    if (ret.success) {
+      // show loading modal
+      this.setState({
+        modal:
+          // eslint-disable-next-line react/jsx-indent
+          <SendTransactionFeedbackModal
+            text={t`Your transfer is being processed`}
```
msbrogli, line 75: I think we should say: "Your transaction is being resolved...".
We may also add something like: "This is an anti-spam measure to prevent abuse."

What do you think?

pedroferreira1: I think the app is for more common use and some technical information is not always good. Besides that, I think adding the sentence could become too much text. I could change 'processed' to 'resolved' but I still don't know which one is the best.

obiyankenobi: Agree with @pedroferreira1. I actually think we could change to "Your transaction is being processed", but I also don't mind the way it is.

msbrogli: I agree with you, but I'd like to differentiate a "mining slowness" from a "propagation slowness". How can we do it without complicating the UX?

Maybe we can just say "Preparing your transaction..." or "Mining your transaction..." or "Resolving your transaction", and add a link for further information.

pedroferreira1: Where would we add this link for further information? I think we don't have where to put it. RIght now we are already adding a text "Propagating transaction to the network", when it moves from mining to propagating, so it already differentiate both action, however the propagation is so fast that is not even noticed.

msbrogli: Ok. I agree that we have a small box, and we still don't have any explanations in our website. I think we can create another issue to do it later and move on.

pedroferreira1: Done

PR #62, opened 2020-06-09
fix: removing xpub migration from walletData and getting it from xpriv

pedroferreira1:
…priv

### DIFF
--- a/src/screens/PinScreen.js
+++ b/src/screens/PinScreen.js
@@ -98,10 +98,37 @@ class PinScreen extends React.Component {
     });
   }
 
+  /**
+   * Handle data migration on unlock screen
+   * This method is executed when the wallet is unlocked, so we can check
+   * if we need to change anything on the app data after an update
+   *
+   * @param {String} pin Unlock PIN written by the user
+   */
+  handleDataMigration = (pin) => {
+    const accessData = hathorLib.wallet.getWalletAccessData();
+    const walletData = hathorLib.wallet.getWalletData();
+
+    if (accessData !== null &&
+        accessData.xpubkey === undefined &&
+        accessData.mainKey !== undefined) {
+      // Two situations are handled here:
+      // 1. From v0.12.0 to v0.13.0 of the lib, xpubkey has changed from wallet:data to wallet:accessData.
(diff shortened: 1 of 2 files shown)


src/screens/PinScreen.js
```diff
@@ -98,10 +98,37 @@ class PinScreen extends React.Component {
...
   }
 
+  /**
+   * Handle data migration on unlock screen
+   * This method is executed when the wallet is unlocked, so we can check
+   * if we need to change anything on the app data after an update
+   *
+   * @param {String} pin Unlock PIN written by the user
+   */
+  handleDataMigration = (pin) => {
```
msbrogli, line 108: I'm not this is the best place to run the migrations, but it is better than `LoadHistoryScreen`.

src/screens/PinScreen.js
```diff
@@ -98,10 +98,35 @@ class PinScreen extends React.Component {
...
+      // Two situations are handled here:
+      // 1. From v0.12.0 to v0.13.0 of the lib, xpubkey has changed from wallet:data
+      // to wallet:accessData. So if the user is still in an app version before v0.13.0,
+      // we can't use xpubkey directly from accessData
+      //
+      // 2. When the user updated the app to the newest version directly from a version before we've
+      // executed the xpubkey migration. In that case we have deleted the user walletData before
+      // migrating the xpubkey to the accessData
+      const xpubkey = hathorLib.wallet.getXPubFromPin(pin);
+      accessData.xpubkey = xpubkey;
```
obiyankenobi, line 121: Don't we also need to persist this new info (`this.setWalletAccessData`)?

pedroferreira1: Yeah, it's better. The way it's done now I would need to generate xpubkey every time the user unlocks the wallet.

src/screens/PinScreen.js
pedroferreira1: I think we should have a screen for data migration with a loading and an explanation for the user (and asking for the PIN, if needed). But this is just the fix for the current version, I don't think we should do this now, only when doing the data migration task

msbrogli: I agree. We can move forward.

PR #73, opened 2020-06-30
fix: prevent choosing pin with more than 6 digits

pedroferreira1:


src/screens/PinScreen.js
```diff
@@ -142,7 +143,7 @@ class PinScreen extends React.Component {
   }
 
   onChangeText = (text) => {
-    if (text.length === 6) {
+    if (text.length === PIN_SIZE) {
       setTimeout(() => this.validatePin(text), 300);
     }
```
obiyankenobi, line 148: I'm wondering if we should also have this code here.
```
    if (text.length > PIN_SIZE) {
      return;
    }
```

It seems it might also be possible to add an extra char to the pin here, although we haven't been able to do it during tests.

pedroferreira1: This screen is only used for validation, so we don't change anything in storage. In the code we use 

```
if (text.length === PIN_SIZE) {
      setTimeout(() => this.validatePin(text), 300);
    }
```

We don't use a state variable in the validation method, we pass `text` as a parameter (and `text` is the input when it hits 6 digits). So it's impossible to add the extra pin, we will always validate with 6 digits only.

It's not that we weren't able to do with the manual tests, I believe with this code it's impossible to do.

obiyankenobi: You're right, won't cause any errors.

It might end up adding an extra char to the pin state, as the code after this block you sent is executed. It will be cleared immediately, so it's not a problem right now, but might be if we change something in the future and use `state.pin` in some other situation. Anyway, we could just add a `return` after the `setTimeout`.

pedroferreira1: I added the `if` block code in the beginning of the validation method.

```
    if (text.length > PIN_SIZE) {
      return;
    }
```

PR #75, opened 2020-07-09
feat: add js error handling

pedroferreira1:


src/errorHandler.js
```diff
@@ -0,0 +1,62 @@
...
+  Clipboard.setString(`${error.name} - ${error.message}\n${accessData}`);
+};
+
+/**
+ * Capture all JS exceptions that might happen and show an alert to the user
+ */
+export const errorHandler = (error, isFatal) => {
+  if (isFatal) {
+    Alert.alert(
+      'Unexpected error occurred',
```
obiyankenobi, line 43: Missing localization markers for all strings.

README.md
```diff
@@ -6,6 +6,14 @@
 
 `npm run node-hack`
 
+### Podfile install for iOS
+
```
msbrogli, line 10: What are podfiles?

android/sentry.properties
```diff
@@ -0,0 +1,4 @@
+defaults.url=https://sentry.io/
+defaults.org=hathor
+defaults.project=wallet-mobile
+auth.token=1d9fa34e90034945a983323f9f7a1b9039b00d6a507d4246b02658c79d0c0cb8
```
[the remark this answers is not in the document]
msbrogli: I agree that it should be kept in a config file. Can we use envvars?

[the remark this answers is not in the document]
msbrogli: Even though it will be packed with the final AAB. I'm not sure if anyone can download and inspect an AAB. Do you know, @obiyankenobi and @jansegre ?

index.js
```diff
@@ -7,7 +7,15 @@
 
 import './i18nInit';
 import { AppRegistry } from 'react-native';
+import { setJSExceptionHandler } from 'react-native-exception-handler';
 import App from './src/App';
 import { name as appName } from './app.json';
+import { errorHandler } from './src/errorHandler';
+
+// - errorHandler is the callback
+// - 'true' is 'allowInDevMode' is an optional parameter
```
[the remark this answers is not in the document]
msbrogli: I agree with @obiyankenobi.

src/errorHandler.js
```diff
@@ -0,0 +1,62 @@
...
+    dsn: 'https://c1ebae9159f741e8937abdbfbeba8e8a@o239606.ingest.sentry.io/5304101',
+  });
+  Sentry.withScope((scope) => {
+    scope.setExtra('App version', JSON.stringify(VersionNumber));
+    Sentry.captureException(error);
+  });
+};
+
+/**
+ * Copy access data from storage to the clipboard.
```
msbrogli, line 27: It is really dangerous to do it. I agree it may really help identify some issues, but we cannot simply copy to the clipboard because many users may share this information, and without a strong password, they seed may be compromised.

Maybe we can copy the keys, content type and length, but never the content itself.

src/errorHandler.js
```diff
@@ -0,0 +1,62 @@
...
+};
+
+/**
+ * Capture all JS exceptions that might happen and show an alert to the user
+ */
+export const errorHandler = (error, isFatal) => {
+  if (isFatal) {
+    Alert.alert(
+      'Unexpected error occurred',
+      '\nUnfortunately a fatal error happened and you need to restart your app.\n\nBefore doing that, we kindly ask you to report this error to the Hathor team clicking on the button below.',
```
msbrogli, line 44: It would be good to say that no sensitive information will be shared.

src/errorHandler.js
```diff
@@ -0,0 +1,62 @@
...
+      [{
+        text: 'Report error',
+        onPress: () => {
+          sentryReportError(error);
+        }
+      },
+      {
+        text: 'Copy data',
```
msbrogli, line 52: Maybe we should remove this button. Simply saying "Copy data" may mislead many users.

src/screens/About.js
```diff
@@ -28,6 +29,16 @@ export class About extends React.Component {
     },
   }));
 
+  touchCount = 0;
+
+  onPressLogo = () => {
+    // Method created to test error handling
```
msbrogli, line 35: Maybe you should set a timer to clear this counter after N seconds.

src/screens/About.js
```diff
@@ -28,6 +29,16 @@ export class About extends React.Component {
...
+    this.touchCount += 1
+    if (this.touchCount === 10) {
+      throw new Error('Hathor test error.')
```
msbrogli, line 38: It might open some "Dev tools" where we can include the "Copy data" you were talking about. Even though, I'm not sure we should have "copy data" in the first place.

android/sentry.properties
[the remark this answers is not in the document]
obiyankenobi: I have no idea. But we need to check exactly what they are. @pedroferreira1 was telling me that sentry recommends adding it to source control, so they might not need to be secret.

src/errorHandler.js
pedroferreira1: Yeah, I agree with you. What about if we send to Sentry (if the user reports the error) this information you listed (the keys, content type and length)? I think this would already help us debug many cases and we are not sharing any sensitive data. And we could remove the 'Copy data' button.

README.md
pedroferreira1: It's an iOS package manager. Sentry needs to add some native codes, so we must install an iOS native lib of Sentry.

Actually Apple already released a new package manager 'Swift Package Manager' but it's not supported for React native yet

android/sentry.properties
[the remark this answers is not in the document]
pedroferreira1: This is a file generated automatically by sentry when configuring it in react native. Sentry recommends adding it to the version control, that's why I commited it and didn't change it.

https://docs.sentry.io/platforms/react-native/

> The Sentry Wizard will guide you through the process of setting everything up correctly. This has to be done only once, and the files created can go into your version control system.

I am still studying more to understand exactly if there is any harm in having this exposed here.

src/errorHandler.js
pedroferreira1: As in the comment above, I agree we shouldn't allow users to copy sensitive data, so there is no reason to keep this button. Will remove it, okay?

src/errorHandler.js
pedroferreira1: Added this, thanks

src/screens/About.js
pedroferreira1: Do you really think it's necessary? Every time the user leaves the 'About' screen, the count goes to 0 again. It's almost impossible to click 10 times in a row on the logo by mistake

src/screens/About.js
pedroferreira1: Yeah, I agree with you. This copy data may not be good, I think we should go without it, at least for now.

index.js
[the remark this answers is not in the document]
pedroferreira1: Updated the comment. Can you see if it's good?

src/errorHandler.js
pedroferreira1: I've added the markers and the translation, however when I do `make compile_msgs` I get errors because the new messages were not translated to danish. 

```
mkdir -p src/locale/pt-br/
npx ttag po2json locale/pt-br/texts.po > src/locale/pt-br/texts.po.json
msgcmp locale/da/texts.po locale/texts.pot
locale/texts.pot:8: this message is used but not defined in locale/da/texts.po
locale/texts.pot:18: this message is used but not defined in locale/da/texts.po
locale/texts.pot:22: this message is used but not defined in locale/da/texts.po
msgcmp: found 3 fatal errors
make: *** [locale/da/texts.po] Error 1
```

@msbrogli Am I missing something here on the translating update process?

src/errorHandler.js
obiyankenobi: Should we replace it by a Cancel button? What's the user supposed to do if he doesn't want to send the info? Restart the app? Maybe a Close button then.

pedroferreira1: For what I know Apple has never allowed apps to close themselves, that's why there are no react native commands to do it, only libraries (https://github.com/wumke/react-native-exit-app) that execute native commands. This Apple restriction was in the past, I don't know today. I could do some tests with this library but still no sure if it's worth it to risk having the app rejected by Apple.

obiyankenobi: Got it, I agree to just remove the button then.

android/sentry.properties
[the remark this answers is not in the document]
pedroferreira1: So I've found a stackoverflow reply of one of the Sentry developers saying the opposite of the Sentry official docs. https://stackoverflow.com/a/48376918/2117670

> We recommend treating a sentry.properties like an .env file.

It is basically the same, so you should add it to your e.g. .gitignore.

The reason why it's called sentry.properties is because of android gradle, we needed it to be read natively.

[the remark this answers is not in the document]
pedroferreira1: Done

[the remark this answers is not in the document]
obiyankenobi: Maybe we should commit a template file without `auth.token`. I think the rest of the information is not sensitive. And explain in the README that we need this file to be created in the `ios` and `android` folders. Or just add the file template directly on the README.

src/errorHandler.js
msbrogli: @pedroferreira1 You need to update the PO files as well. We don't have a `make` command for it; you can just follow the instructions in the README.

README.md
msbrogli: Thanks! I think you can add this information at the beginning of this section.

src/errorHandler.js
msbrogli: That's a great idea!

src/errorHandler.js
msbrogli: Ok! Go ahead!

src/screens/About.js
msbrogli: I didn't know it would reset every time user left the screen. I think it is ok.

android/sentry.properties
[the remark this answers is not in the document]
pedroferreira1: Added to README. Can you see if it's good?

README.md
pedroferreira1: Done

src/errorHandler.js
pedroferreira1: Button deleted.

src/errorHandler.js
pedroferreira1: Added the keys/content type/length on sentry report.

android/sentry.properties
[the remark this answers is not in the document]
obiyankenobi: Good!

src/errorHandler.js
```diff
@@ -0,0 +1,59 @@
...
+import { Alert } from 'react-native';
+import hathorLib from '@hathor/wallet-lib';
+import * as Sentry from '@sentry/react-native';
+import VersionNumber from 'react-native-version-number';
+import { t } from 'ttag';
+
+/**
+ * Send error to Sentry
+ */
+export const sentryReportError = (error) => {
```
obiyankenobi, line 17: Do we need to export this?

pedroferreira1: No need to. I've removed the `export`

pedroferreira1:
I've changed a bit the alert message and flow. Given that we are not handling native errors, the app won't need to be restarted (it won't freeze). So the behaviour can be the same as in the wallet desktop, we ask to report the error and the user can continue using the app because we are handling the js error.

Wallet desktop:
desktop

Wallet mobile:
app

@msbrogli @obiyankenobi what do you think?

obiyankenobi:
My only concern is that this error might be caused by some inconsistent state. In that case, forcing the user to restart the app seems a more "reliable" solution. I'd prefer that we displayed a modal that can't be dismissed asking the user to restart the app. We could adopt a similar solution for the Desktop wallet.

msbrogli:
> My only concern is that this error might be caused by some inconsistent state. In that case, forcing the user to restart the app seems a more "reliable" solution. I'd prefer that we displayed a modal that can't be dismissed asking the user to restart the app. We could adopt a similar solution for the Desktop wallet.

I tend to agree with @obiyankenobi that it seems to be a more "reliable" solution, even though it can put users in an infinite restart loop. I have mixed feeling about forcing the user to restart, but it seems to be the best thing here.

pedroferreira1:
I don't like the static alert for two reasons, one is the infinite loop and the other is to stop the user in the middle of a long action (e.g. in the end of a long form). But we don't have 'long actions' in our app for now and it's good to have the static alert to prevent something weird from happening.

Here the two possible cases:

1. User did not send the error to Sentry

gif1

2. User sent error to Sentry

gif2

PR #93, opened 2021-03-19
Integrate app with library facade

pedroferreira1:


src/actions.js
```diff
@@ -121,65 +127,116 @@ export const clearInitWallet = () => ({ type: types.SET_INIT_WALLET, payload: nu
...
+        // ERROR
+        dispatch(fetchHistoryError());
+      } else if (state === 3) {
+        // READY
+        const historyTransactions = wallet.getTxHistory();
+        const addresses = wallet.getAllAddresses()
+        dispatch(fetchHistorySuccess(historyTransactions, addresses));
+      } else if (state === 0) {
+        // CLOSED
+        // XXX Should we remove the conn event listeners?
```
andreabadesso, line 177: What event listeners are you talking about?  `wallet` event listeners or  `connection` event listeners?

I believe that `wallet` event listeners will be closed automatically, but I believe we should manually close the `connection` event listeners

src/constants.js
```diff
@@ -20,10 +20,8 @@ import {
   _SENTRY_DSN as SENTRY_DSN,
 } from './config';
 
-hathorLib.storage.setStore(new AsyncStorageStore());
-hathorLib.storage.setItem('wallet:server', 'https://mobile.wallet.hathor.network/v1a/');
-hathorLib.network.setNetwork('mainnet');
-
+export const STORE = new AsyncStorageStore()
+hathorLib.storage.setStore(STORE);
```
andreabadesso, line 24: We don't need to `setNetwork(...)` anymore? Is it being set somewhere else?

src/actions.js
```diff
@@ -121,65 +127,116 @@ export const clearInitWallet = () => ({ type: types.SET_INIT_WALLET, payload: nu
...
-      const historyTransactions = walletData.historyTransactions || {};
-      const { keys } = hathorLib.wallet.getWalletData();
-      dispatch(fetchHistorySuccess(historyTransactions, keys));
-    }, () => {
-      dispatch(fetchHistoryError());
+
+  wallet.start().then((serverInfo) => {
+    dispatch(setServerInfo(serverInfo));
+    wallet.on('state', (state) => {
+      if (state === 4) {
```
msbrogli, line 167: Use `HathorWallet.ERROR`.

src/actions.js
```diff
@@ -121,65 +127,116 @@ export const clearInitWallet = () => ({ type: types.SET_INIT_WALLET, payload: nu
...
+        // ERROR
+        dispatch(fetchHistoryError());
+      } else if (state === 3) {
```
msbrogli, line 170: Use `HathorWallet.READY`.

src/actions.js
```diff
@@ -121,65 +127,116 @@ export const clearInitWallet = () => ({ type: types.SET_INIT_WALLET, payload: nu
...
+        // READY
+        const historyTransactions = wallet.getTxHistory();
+        const addresses = wallet.getAllAddresses()
+        dispatch(fetchHistorySuccess(historyTransactions, addresses));
+      } else if (state === 0) {
```
msbrogli, line 175: Use `HathorWallet.CLOSED`.

src/actions.js
```diff
@@ -121,65 +127,116 @@ export const clearInitWallet = () => ({ type: types.SET_INIT_WALLET, payload: nu
...
+    });
+
+    connection.on('height-updated', (height) => {
+      dispatch(updateHeight(height));
     });
-  }, () => {
-    dispatch(fetchHistoryError());
+
+    connection.on('is-online', (value) => {
+      dispatch(setIsOnline(value));
```
msbrogli, line 197: Can't we do it in the `state` event of the wallet?

src/components/ReceiveMyAddress.js
```diff
@@ -43,8 +47,9 @@ class ReceiveMyAddress extends React.Component {
   }
 
   getNextAddress = () => {
-    const nextAddress = hathorLib.wallet.nextAddress();
-    this.setState({ address: nextAddress });
+    // Mark current address as used, then get the next
+    this.props.wallet.getCurrentAddress({ markAsUsed: true });
+    this.setState({ address: this.props.wallet.getCurrentAddress() });
```
msbrogli, line 52: Should we create a `getNextAddress()` method?

src/constants.js
msbrogli: The wallet is supposed to call `setNetwork()`.

msbrogli, line 24: Shouldn't we pass the store to the wallet instead of directly calling `hathorLib.storage.setStore()`?

src/reducer.js
```diff
@@ -434,10 +444,44 @@ const onUpdateHeight = (state, action) => {
   return state;
 };
 
+const onSetWallet = (state, action) => {
+  if (state.wallet && state.wallet.state !== 0) {
```
msbrogli, line 448: Use `HathorWallet.CLOSED`.

src/reducer.js
```diff
@@ -434,10 +444,44 @@ const onUpdateHeight = (state, action) => {
...
+    // Wallet was not closed
+    state.wallet.stop();
```
msbrogli, line 450: Does it return a promise? Should we wait for it?

src/screens/PinScreen.js
```diff
@@ -129,6 +131,12 @@ class PinScreen extends React.Component {
       // in case it's the lock screen, we just have to execute the data migration
       // method an change redux state. No need to execute callback or go back on navigation
       this.handleDataMigration(pin);
+      if (!this.props.wallet) {
+        // We are saving HathorWallet object in redux, so if the app has lost redux information and is in locked screen
+        // we must start the HathorWallet object again
+        const words = hathorLib.wallet.getWalletWords(pin);
```
msbrogli, line 137: Shouldn't we create a `getWalletWords()` in the utils and call this method from there? I feel we should put all direct access to the walletLib in the same place for future refactor.

src/utils.js
```diff
@@ -33,7 +33,7 @@ export const getMyTxBalance = (tx, myKeys) => {
       continue;
     }
     if (txout.decoded && txout.decoded.address
```
msbrogli, line 35: Shouldn't this whole method (`getMyTxBalance()`) be replaced by the wallet object?

msbrogli (changes requested):

src/actions.js
pedroferreira1: Actually this is an old comment. When I stop the wallet (and the state changes to CLOSED), we remove the wallet event listeners and stop the connection, which removes the conn event listeners, so I think it's enough.

What do you think?

src/constants.js
pedroferreira1: Yeah, the `Connection` object sets the network when it's started.

src/actions.js
pedroferreira1: Yes, we can. Changed it

src/constants.js
pedroferreira1: This is one of the parts of the code that I left to a later refactor. The storage is needed in the whole wallet, not only when we have a wallet object. Because of that we need to have a global store in the lib (we have some keys that are used in the wallet, e.g. 'loaded'). I believe removing this store dependency should be in a different refactor.

src/reducer.js
pedroferreira1: It does not return a promise.

src/screens/PinScreen.js
pedroferreira1: Makes sense, done that.

src/components/ReceiveMyAddress.js
pedroferreira1: I agree with you. Even though I am not 100% sure about this, if we keep this feature in the wallet we should have this method

src/utils.js
pedroferreira1: We have another method `getBalance` in `utils.js` that should be full removed and I haven't noticed. I will look better on how we calculate the balance here and try to use only the wallet object.

src/components/ReceiveMyAddress.js
pedroferreira1: Done

src/utils.js
pedroferreira1: I've removed the method `getBalance` but I am currently keeping this `getMyTxBalance`. This method is used when receiving a new tx to get some information about the tokens and balance of the new tx for the wallet. We use this to update some internal redux variables, including the tokens list.

What do you think? It's not exactly getting the wallet balance, I feel we could keep like this.

src/constants.js
msbrogli: I got it. Let's just add a comment here explaining why it is needed and and it should be refactored later on.

src/components/ReceiveMyAddress.js
msbrogli: Great! Thanks!

src/utils.js
msbrogli: I feel the HathorWallet instance should handle all of this. When a new tx is received, the Hathor Wallet instance will update its internals including the list of tokens. What do you think?

src/constants.js
pedroferreira1: Done

src/actions.js
andreabadesso: I agree. That should be enough

src/utils.js
pedroferreira1: I agree with you and the HathorWallet instance already has the list of tokens. The wallet instance has a method `getBalance`, which receives a token and returns the balance of it. I will explain better what this method (`getMyTxBalance`) is doing:

Because some calculus in the lib are expensive before we integration with wallet service, we pre save some information on redux. When the app receives a new tx in the ws we execute the following actions:

- Call `getMyTxBalance` to get know which tokens are spent/received for the wallet in this tx (we might have an output for HTR but it's not for this wallet and this wallet is only receiving PedroCoin).
- With this information we (i) update the PedroCoin history list, and (ii) call `wallet.getBalance` for PedroCoin token to update the token balance.

Then the method `getMyTxBalance` is used to know which tokens are spent/received in the new tx, in order to run the updates. We will always need to update redux variables and I think this is the best approach, I don't feel we are running any lib internal operation here.

This will change with the wallet service integration but I think we should keep like this for now.

msbrogli: I think that the wallet must be just a UI/UX over the `HathorWallet` instance. It doesn't have to do any processing except update the UI when needed. Maybe we should move `getMyTxBalance` to the wallet-lib, so the UI can efficiently update itself without recalculating all the balances. Does it make sense?

andreabadesso (approved):

src/utils.js
pedroferreira1: Yeah, it makes sense. I am now using a method `getTxBalance` from the wallet instance.

msbrogli (approved):

PR #104, opened 2021-05-27
feat: add check version script to CI and Makefile

pedroferreira1:


scripts/check_version
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+
+PACKAGE_VERSION=`grep '"version":' ./package.json | cut -d '"' -f4`
+ANDROID_GRADLE_VERSION=`grep "versionName " ./android/app/build.gradle | cut -d '"' -f2`
+TEMP_IOS_VAR=`grep "MARKETING_VERSION " ./ios/HathorMobile.xcodeproj/project.pbxproj | cut -d"=" -f2 | awk '{print substr($0, 2, length($0) - 2)}'`
```
msbrogli, line 5: You can use this perl command to capture the versions in two lines. I feel it is better than the awk solution at the end.

`perl -n -e '/MARKETING_VERSION = (.*);/ && print "$1\n"' ./ios/HathorMobile.xcodeproj/project.pbxproj`

msbrogli (approved):

PR #107, opened 2021-06-21
new issue on send screen

mbnunes:
I add paste funcion on address manual, and change Qrcode screen order. 

@pedroferreira1


mbnunes:
https://user-images.githubusercontent.com/931636/123686902-adbaf600-d826-11eb-8caf-4cad2c9c2aba.mp4

src/screens/SendAddressInput.js
```diff
@@ -36,6 +39,15 @@ class SendAddressInput extends React.Component {
     this.setState({ address: text, error: null });
   }
 
+  onPasteAddress = async () => {
+    let pasteText = await Clipboard.getString();
+    if (this.state.address.length === 0) {
+      this.state.address = '';
```
msbrogli, line 45: You can't change the state like this.

src/screens/SendAddressInput.js
```diff
@@ -36,6 +39,15 @@ class SendAddressInput extends React.Component {
...
+    }
+    pasteText = this.state.address + pasteText;
```
msbrogli, line 47: Should we trim and remove all non-address stuff from the text?

msbrogli (approved):
@mbnunes I guess the video added here is outdated because I can't see the paste icon. Would you please update it?

I liked the new flow and the code is good to me except for one point of improvement.

pedroferreira1:
I've found 3 problems testing this code:

1. The paste button does not work properly when I open the send screen. It only works if I type something in the input before pasting.

https://user-images.githubusercontent.com/3298774/124313731-a9a11800-db47-11eb-953f-2d3269620f6a.mov

2. We show an error on the screen if the address is invalid, then the Paste button loses the alignment
image
image

3. We must remove the Back button from the send screen (it's doing nothing)

mbnunes:
Hi, i get this issue again. How can open again?

Oi eu peguei nesta issue novamente, Como posso fazer para dar continudade?

mbnunes:
@pedroferreira1  corrigi  o que voce mencionou anteriormente, estou enviando o video. desculpe pela demora.

@pedroferreira1  correct what you mentioned earlier, I'm uploading the video. Sorry for the delay.

https://user-images.githubusercontent.com/931636/184691135-f7805580-28a4-4432-bae2-bc30c2e03ab4.mp4


https://user-images.githubusercontent.com/931636/184691144-58eded8d-9209-463f-84d1-52b879489a94.mp4

mbnunes:
@pedroferreira1  the branche is not master, is qr-code in my fork

mbnunes:
Done i merged with my master branch 

Pronto! eu fiz merg do branche qr-code para o meu master.

pedroferreira1:
@mbnunes thanks for the update. Just some things

1. In your last video here (https://github.com/HathorNetwork/hathor-wallet-mobile/pull/107#issuecomment-1215540581) you can see that your first 'Paste' adds a `null` and it's happening to me as well.
2. We should remove the back button from the send screen with manual address now that it's the first screen.
3. There are some comments @msbrogli did in your code that should be addressed.
4. The paste button is not aligned with the input.
image
5. After reading the QRCode with the address, the app goes back to the manual input screen without the address.
6. When we are typing the address, we must have a way to close the keyboard, so we can continue interacting with the app menu.
7. When we move from the app menu, we must clean what's in the address input.

PR #111, opened 2021-07-30
Integration of new facades signatures with app

pedroferreira1:
This code will work for both facades and it's ready to receive the rollout feature flag


src/actions.js
```diff
@@ -135,50 +139,123 @@ export const clearInitWallet = () => ({ type: types.SET_INIT_WALLET, payload: nu
...
 export const startWallet = (words, pin) => (dispatch) => {
   // If we've lost redux data, we could not properly stop the wallet object
   // then we don't know if we've cleaned up the wallet data in the storage
   walletUtil.cleanLoadedData();
 
-  const connection = new Connection({
-    network: 'mainnet', // app currently connects only to mainnet
-    servers: ['https://mobile.wallet.hathor.network/v1a/'],
-  });
+  const networkName = 'testnet';
```
andreabadesso, line 209: I don't understand, why is this hardcoded and why is it testnet?

src/actions.js
```diff
@@ -135,50 +139,123 @@ export const clearInitWallet = () => ({ type: types.SET_INIT_WALLET, payload: nu
...
 
   dispatch(setWallet(wallet));
 
   dispatch(fetchHistoryBegin());
 
+  wallet.on('state', (state) => {
+    if (state === HathorWallet.ERROR) {
+      // ERROR
+      dispatch(fetchHistoryError());
+    } else if (state === HathorWallet.READY || state === 'Ready') {
```
andreabadesso, line 244: Why do we need this `OR`?

src/screens/SendConfirmScreen.js
```diff
@@ -75,23 +75,35 @@ class SendConfirmScreen extends React.Component {
...
+    } catch (err) {
+      if (err instanceof hathorLib.errors.WalletError) {
+        this.onError(err);
+      } else {
+        throw err;
+      }
+    }
+    // show loading modal
+    this.setState({
+      modal:
```
andreabadesso, line 97: Add parenthesis around the component

andreabadesso (changes requested):

src/actions.js
```diff
@@ -135,50 +139,123 @@ export const clearInitWallet = () => ({ type: types.SET_INIT_WALLET, payload: nu
...
+    const history = await wallet.getTxHistory({ token_id: token });
+    tokensHistory[token] = history.map((element) => mapTokenHistory(element, token));
+    /* eslint-enable no-await-in-loop */
+  }
+
+  return { tokensHistory, tokensBalance };
+};
+
+/**
+ * Fetch pagination history for specific token
```
msbrogli, line 191: Typo: paginated history

src/actions.js
```diff
@@ -135,50 +139,123 @@ export const clearInitWallet = () => ({ type: types.SET_INIT_WALLET, payload: nu
...
 
-  const beforeReloadCallback = () => {
-    dispatch(activateFetchHistory());
-  };
+  let wallet;
+  // TODO substitute for feature flag when rollout code is implemented
+  if (false) {
+    const connection = new Connection({
+      network: networkName, // app currently connects only to mainnet
+      servers: ['https://node1.foxtrot.testnet.hathor.network/v1a/'],
```
msbrogli, line 216: Hardcoded URL to testnet?

src/actions.js
```diff
@@ -135,50 +139,123 @@ export const clearInitWallet = () => ({ type: types.SET_INIT_WALLET, payload: nu
...
-    servers: ['https://mobile.wallet.hathor.network/v1a/'],
-  });
+  const networkName = 'testnet';
 
-  const beforeReloadCallback = () => {
-    dispatch(activateFetchHistory());
-  };
+  let wallet;
+  // TODO substitute for feature flag when rollout code is implemented
+  if (false) {
```
msbrogli, line 213: Maybe create a variable `const useWalletService = false`, so we know what this condition is related to.

src/actions.js
msbrogli: I agree it feels really weird and should be removed.

src/actions.js
```diff
@@ -135,50 +139,123 @@ export const clearInitWallet = () => ({ type: types.SET_INIT_WALLET, payload: nu
...
-        dispatch(fetchHistorySuccess(historyTransactions));
-      }
-    });
+    walletUtil.storePasswordHash(pin);
+    walletUtil.storeEncryptedWords(words, pin);
+    dispatch(setServerInfo({ version: null, network: networkName }));
 
-    wallet.on('new-tx', (tx) => {
+    // XXX no real time update for now
+    /* wallet.on('new-tx', (tx) => {
```
msbrogli, line 258: This is removing the real time update for both modes (full node API and wallet service). Shouldn't we keep it enabled for the full node API?

src/screens/CreateTokenConfirm.js
```diff
@@ -68,28 +68,46 @@ class CreateTokenConfirm extends React.Component {
...
-    const address = this.props.wallet.getCurrentAddress({ markAsUsed: true });
-    const ret = this.props.wallet.createNewToken(
-      this.name, this.symbol, this.amount, address, { pinCode: pin }
-    );
+    const { address } = this.props.wallet.getCurrentAddress({ markAsUsed: true });
+    this.props.wallet.prepareCreateNewToken(
+      this.name, this.symbol, this.amount, { address, pinCode: pin }
+    ).then((tx) => {
+      let sendTransaction;
+      // TODO substitute for feature flag when rollout code is implemented
```
msbrogli, line 76: Maybe create a constant that will be replaced later (e.g. `isWalletService`)?

src/screens/MainScreen.js
```diff
@@ -176,13 +187,43 @@ class TxHistoryView extends React.Component {
...
+  };
+
+  loadMoreHistory = async () => {
+    if (!this.state.canLoadMore) {
+      // Already loaded all history
+      return;
+    }
+
+    this.setState({ loading: true });
+    const newHistory = await fetchMoreHistory(
```
msbrogli, line 206: I feel we could use an async iterator here.

src/screens/PinScreen.js
```diff
@@ -111,7 +111,7 @@ class PinScreen extends React.Component {
   handleDataMigration = (pin) => {
     const accessData = hathorLib.wallet.getWalletAccessData();
 
-    if (accessData !== null && accessData.xpubkey === undefined) {
+    if (accessData !== null && accessData.xpubkey === undefined && accessData.mainKey) {
```
msbrogli, line 114: Why do we need this new criteria?

src/screens/SendConfirmScreen.js
```diff
@@ -75,23 +75,35 @@ class SendConfirmScreen extends React.Component {
...
-            sendTransaction={ret.sendTransaction}
-            successText={<TextFmt>{t`Your transfer of **${this.amountAndToken}** has been confirmed`}</TextFmt>}
-            onDismissSuccess={this.exitScreen}
-            onDismissError={() => this.setState({ modal: null })}
-          />,
-      });
+    const outputs = [{ address: this.address, value: this.amount, token: this.token.uid }];
+    let sendTransaction;
+    // TODO substitute for feature flag when rollout code is implemented
+    if (false) {
```
msbrogli, line 81: This is the second time I see this code. I feel we should move it to an util method (or pass it as `this.props.sendTransaction` from mapStateToProps).

msbrogli (changes requested):

src/actions.js
pedroferreira1: No need to, we have a `wallet.isReady()` method and I should've used it. Done

src/actions.js
pedroferreira1: The method that was called when a new tx arrived changed a bit and I am not sure how the update will happen in the wallet service facade. That's why I preferred not to waste time making the real time work for the old facade because it might not work for the new one also and end up being refactored. 

I don't think we should handle this in this PR.

src/actions.js
pedroferreira1: I was testing the wallet service that was working only for testnet and forgot to change back. Fixed https://github.com/HathorNetwork/hathor-wallet-mobile/pull/111/commits/6f6c05de5d4714040ee97ffdd2dcddf37e9ac619

src/actions.js
pedroferreira1: I was testing the wallet service that was working only for testnet and forgot to change back. Fixed https://github.com/HathorNetwork/hathor-wallet-mobile/pull/111/commits/6f6c05de5d4714040ee97ffdd2dcddf37e9ac619

src/actions.js
pedroferreira1: Created a redux variable that is being used now as feature flag. https://github.com/HathorNetwork/hathor-wallet-mobile/pull/111/commits/6f6c05de5d4714040ee97ffdd2dcddf37e9ac619

src/screens/CreateTokenConfirm.js
pedroferreira1: Created a redux variable that is being used now as feature flag. https://github.com/HathorNetwork/hathor-wallet-mobile/pull/111/commits/6f6c05de5d4714040ee97ffdd2dcddf37e9ac619

src/screens/SendConfirmScreen.js
pedroferreira1: Created a redux variable that is being used now as feature flag. https://github.com/HathorNetwork/hathor-wallet-mobile/pull/111/commits/6f6c05de5d4714040ee97ffdd2dcddf37e9ac619

src/screens/PinScreen.js
pedroferreira1: This is an old code for a data migration that we did more than a year ago (I thought about removing this code to be honest) where we didn't have the xpubkey saved on storage but now we do.

The wallet service facade does not store the `mainKey` data on storage like the old one, that's why I had to add this new condition.

src/screens/MainScreen.js
pedroferreira1: Why do you think would be good? In the old facade we already have the history pre-calculated, then yielding one of returning a page (15) is not worse and in the wallet service facade we execute a request to get the history, should we request many and yield only one, or request one by one? Either way I feel we just add complexity to the code without any benefits.

src/screens/CreateTokenConfirm.js
msbrogli: Great!

andreabadesso (approved):

src/actions.js
msbrogli: Great! Thanks!

src/actions.js
msbrogli: Great! Thanks!

src/actions.js
msbrogli: I don't think we should merge it in `dev` unless the full node's Wallet API is working because it will block future rollouts of the wallet. I mean, we cannot release a new version without the real time update for the full node's Wallet API.

src/screens/MainScreen.js
msbrogli: It feels natural to let the screen decide how many items it would like to show without diving into the details of how it is loaded from the backend. But I feel we don't have to do it in this PR. Maybe we should just create an issue to discuss it further. What do you think?

src/screens/PinScreen.js
msbrogli: Ah! Ok. Thanks!

src/screens/SendConfirmScreen.js
msbrogli: Great! Thanks!

src/actions.js
msbrogli: Seems good to me.

msbrogli (changes requested):
I'd approve it, but I feel we are breaking the dev branch removing real time updates from the full node's Wallet API.

src/actions.js
pedroferreira1: I agree. I will try to make the real time work in the full node facade in a way that we lose the minimum work as possible when we develop the new real time handling.

pedroferreira1: Real time working now for old facade. https://github.com/HathorNetwork/hathor-wallet-mobile/pull/111/commits/4ea14e548db702a271e092250251e5062394b09f

PR #125, opened 2021-09-22
feat: using device unique identifier on unleash and screen to share it

andreabadesso:
# Acceptance criteria

The user must be able to share his device's unique identifier and Hathor team should be able to use it to fix his device on the new wallet facade

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -359,7 +359,7 @@ export const fetchNewTxTokenBalance = async (wallet, tx) => {
   const updatedBalanceMap = {};
   const balances = wallet.getTxBalance(tx);
   // we now loop through all tokens present in the new tx to get the new balance
-  for (const [tokenUid, _tokenTxBalance] of Object.entries(balances)) {
+  for (const [tokenUid] of Object.entries(balances)) {
     /* eslint-disable no-await-in-loop */
     updatedBalanceMap[tokenUid] = await fetchTokenBalance(wallet, tokenUid);
   }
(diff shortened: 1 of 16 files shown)


src/actions.js
```diff
@@ -237,8 +243,8 @@ export const startWallet = (words, pin) => async (dispatch) => {
   walletUtil.cleanLoadedData();
 
   const networkName = 'mainnet';
-  const firstAddress = getFirstAddressFromWords(words, networkName);
```
pedroferreira1, line 240: Can we remove `getFirstAddressFromWords` method?

pedroferreira1 (approved):

src/actions.js
andreabadesso: Yes we can, done in 0ed36ff319aa50ca4e96c0c564c5e7d3f27eca30

src/actions.js
```diff
@@ -353,7 +359,7 @@ export const fetchNewTxTokenBalance = async (wallet, tx) => {
   const updatedBalanceMap = {};
   const balances = wallet.getTxBalance(tx);
   // we now loop through all tokens present in the new tx to get the new balance
-  for (const [tokenUid, tokenTxBalance] of Object.entries(balances)) {
+  for (const [tokenUid, _tokenTxBalance] of Object.entries(balances)) {
```
r4mmer, line 362: if `_tokenTxBalance` is never used, wouldn't a `_` be better?

r4mmer (approved):

src/actions.js
andreabadesso: I think the effect is the same (making it clear that the variable is unused)

I prefer leaving the variable name there so we know what it is

PR #127, opened 2021-09-28
feat: show NFT amount as integer and allow only sending integers as well

pedroferreira1:
## Screens for NFTs

image
image
image
image
image
image
image
image
image

### DIFF
--- a/src/components/AmountTextInput.js
+++ b/src/components/AmountTextInput.js
@@ -27,12 +27,13 @@ class AmountTextInput extends React.Component {
       return;
     }
 
+    let parsedText = text;
     if (this.props.allowOnlyInteger) {
       // We allow only integers for NFT
-      text = text.replace(/[^0-9]/g, '');
+      parsedText = parsedText.replace(/[^0-9]/g, '');
     }
 
-    const parsedText = getAmountParsed(text);
+    parsedText = getAmountParsed(parsedText);
 
     if (!this.props.allowOnlyInteger) {
       const amount = getIntegerAmount(parsedText);
--- a/src/components/AmountTextInput.js
+++ b/src/components/AmountTextInput.js
@@ -27,16 +27,25 @@ class AmountTextInput extends React.Component {
       return;
     }
 
+    if (this.props.allowOnlyInteger) {
+      // We allow only integers for NFT
+      text = text.replace(/[^0-9]/g, '');
+    }
+
     const parsedText = getAmountParsed(text);
-    const amount = getIntegerAmount(parsedText);
-    if (Number.isNaN(amount) || amount < 0) {
-      return;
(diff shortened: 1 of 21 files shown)


src/components/AmountTextInput.js
```diff
@@ -27,16 +27,25 @@ class AmountTextInput extends React.Component {
       return;
     }
 
+    if (this.props.allowOnlyInteger) {
+      // We allow only integers for NFT
+      text = text.replace(/[^0-9]/g, '');
```
andreabadesso, line 32: I agree with the lint error, assigning to function parameter is bad

Use `let newText = (...)` to avoid side effects

src/components/TokenSelect.js
```diff
@@ -31,6 +31,9 @@ const TokenSelect = (props) => {
 
     const balance = item.uid in props.tokensBalance
       ? props.tokensBalance[item.uid].available : 0;
+
+    const isNFT = item.uid in props.tokenMetadata && props.tokenMetadata[item.uid].nft;
```
andreabadesso, line 35: Same comment from my `hathor-wallet` PR, add a helper `isNFT` method somewhere and re-use it everywhere

andreabadesso (changes requested):
I think there are also some important lint errors to check, at least on the changed files

src/components/AmountTextInput.js
pedroferreira1: Yeah, makes sense. Done

src/components/TokenSelect.js
pedroferreira1: I agree. Did this refactor, can you se again please?

andreabadesso (approved):

r4mmer (approved):

andreabadesso (approved):

src/screens/CreateTokenConfirm.js
```diff
@@ -79,7 +79,9 @@ class CreateTokenConfirm extends React.Component {
           this.props.wallet, { transaction: tx }
         );
       } else {
-        sendTransaction = new hathorLib.SendTransaction({ transaction: tx, pin });
+        sendTransaction = new hathorLib.SendTransaction(
+          { transaction: tx, pin, network: this.props.wallet.getNetworkObject() }
```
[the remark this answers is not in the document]
pedroferreira1: The `SendTransaction` uses the network to validate if the address has the valid first byte, i.e. we need to know the network to validate the addresses.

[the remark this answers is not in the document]
pedroferreira1: Actually you are correct. For this case we can only `runFromMining` and not `run`, then it doesn't need the network object.

PR #136, opened 2021-11-24
fix: change pin should update storage data that depends on the password

andreabadesso:
### Acceptance Criteria
- Should fix https://github.com/HathorNetwork/internal-issues/issues/62
- Depends on https://github.com/HathorNetwork/hathor-wallet-lib/pull/291

### Security Checklist

### DIFF
--- a/src/reducer.js
+++ b/src/reducer.js
@@ -69,7 +69,12 @@ const initialState = {
   tokenMetadata: {},
   metadataLoaded: false,
   uniqueDeviceId: null,
+  // If recoveringPin is set, the app will display
+  // the RecoverPin screen instead of the default navigator
   recoveringPin: false,
+  // tempPin is used to hold the user PIN when recovering the
+  // PIN that encrypted the data on accessData on the RecoverPin
+  // screen
   tempPin: null,
 };
 
--- a/src/reducer.js
+++ b/src/reducer.js
@@ -466,6 +466,8 @@ const onSetInitWallet = (state, action) => ({
 const onSetRecoveringPin = (state, action) => ({
   ...state,
   recoveringPin: action.payload,
+  // Reset temp pin if recovering pin is set to false
+  tempPin: !action.payload ? null : state.tempPin,
 });
 
 const onSetTempPin = (state, action) => ({
--- a/src/reducer.js
+++ b/src/reducer.js
@@ -69,6 +69,8 @@ const initialState = {
   tokenMetadata: {},
   metadataLoaded: false,
   uniqueDeviceId: null,
+  recoveringPin: false,
+  tempPin: null,
 };
 
 const reducer = (state = initialState, action) => {
@@ -133,6 +135,10 @@ const reducer = (state = initialState, action) => {
       return onTokenMetadataLoaded(state, action);
     case types.SET_UNIQUE_DEVICE_ID:
       return onSetUniqueDeviceId(state, action);
+    case types.SET_RECOVERING_PIN:
+      return onSetRecoveringPin(state, action);
+    case types.SET_TEMP_PIN:
+      return onSetTempPin(state, action);
     default:
       return state;
   }
@@ -451,6 +457,16 @@ const onSetInitWallet = (state, action) => ({
   initWallet: action.payload,
 });
 
+
+const onSetRecoveringPin = (state, action) => ({
+  ...state,
+  recoveringPin: action.payload,
+});
+
+const onSetTempPin = (state, action) => ({
+  ...state,
+  tempPin: action.payload,
+});
 
 /**
  * Update height value on redux
--- a/src/screens/ChangePin.js
+++ b/src/screens/ChangePin.js
@@ -159,8 +159,12 @@ class ChangePin extends React.Component {
   }
 
   executeChangePin = () => {
-    const success = hathorLib.wallet.changePin(this.state.pin1, this.state.pin2);
-    if (success) {
+    const successChangePin = hathorLib.wallet.changePin(this.state.pin1, this.state.pin2);
+    // We also need to change the password because on the mobile app we
+    // use the PIN as Password
+    const successChangePassword = hathorLib.wallet.changePassword(this.state.pin1, this.state.pin2);
+
+    if (successChangePin && successChangePassword) {
       this.setState({ done: true });
     } else {
       // Should never get here because we've done all the validations before
--- a/src/screens/ChangePin.js
+++ b/src/screens/ChangePin.js
@@ -22,6 +22,9 @@ import TextFmt from '../components/TextFmt';
 import baseStyle from '../styles/init';
 import checkIcon from '../assets/images/icCheckBig.png';
 import { PIN_SIZE } from '../constants';
+import {
+  changePin,
+} from '../utils';
 
 
 class ChangePin extends React.Component {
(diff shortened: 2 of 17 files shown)


pedroferreira1 (approved):

src/screens/ChangePin.js
```diff
@@ -159,8 +159,12 @@ class ChangePin extends React.Component {
   }
 
   executeChangePin = () => {
-    const success = hathorLib.wallet.changePin(this.state.pin1, this.state.pin2);
-    if (success) {
+    const successChangePin = hathorLib.wallet.changePin(this.state.pin1, this.state.pin2);
```
pedroferreira1, line 162: Should we create an util method `changePin` that call both methods?

src/screens/RecoverPin.js
```diff
@@ -0,0 +1,89 @@
...
+    this.props.setRecoveringPin(false);
+    this.props.lockScreen();
+  }
+
+  render() {
+    const renderLoading = () => (
+      <View style={{ alignItems: 'center' }}>
+        <Spinner size={48} animating />
+        <Text style={[styles.text, { marginTop: 32, color: 'rgba(0, 0, 0, 0.5)' }]}>
+          {t`Upgrading your wallet, please hang on.`}
```
msbrogli, line 67: We need to give some feedback that it is running indeed. I think we can add a progress wheel or a progress bar that grows with the advance in the `guessPin` method.

msbrogli (changes requested):

src/screens/RecoverPin.js
andreabadesso: I added the percentage below the progress wheel. What do you think?

src/screens/ChangePin.js
andreabadesso: I agree, done in d8390ee0fdec2e30e0b0afbc606ab4a481c42661

src/utils.js
```diff
@@ -258,3 +259,109 @@ export const renderValue = (amount, isInteger) => {
...
+  }
+
+  if (pinFound !== -1) {
+    return [true, pinFound];
+  }
+
+  return [false, ''];
+};
+
+function guessScheduler(data, begin, step, progressCb, successCb, errorCb) {
```
pedroferreira1, line 299: Missing docstring

Should we use `const` instead of `function`?

src/utils.js
```diff
@@ -258,3 +259,109 @@ export const renderValue = (amount, isInteger) => {
 export const isTokenNFT = (uid, metadatas) => (
   uid in metadatas && metadatas[uid].nft
 );
+
+const guessPartial = (data, begin, end) => {
```
pedroferreira1, line 263: Missing docstring

pedroferreira1 (approved):

src/screens/PinScreen.js
```diff
@@ -161,11 +170,26 @@ class PinScreen extends React.Component {
...
+          throw new Error('Words decrypted with pin are invalid');
+        }
+
+        this.dismiss(pin);
+      } else {
+        this.removeOneChar();
+      }
+    } catch (e) {
+      this.props.unlockScreen();
+      this.props.setRecoveringPin(true);
```
msbrogli, line 191: Can't we send the temporary pin to this method (and remove the next call)?

src/screens/RecoverPin.js
```diff
@@ -0,0 +1,149 @@
...
+        erroed: true,
+      });
+      return;
+    }
+
+    const currentPin = this.props.tempPin;
+    // update the encrypted words with the pin the user knows about
+    hathorLib.wallet.changePassword(pin, currentPin);
+
+    this.props.setRecoveringPin(false);
```
msbrogli, line 80: Shouldn't we clean up the temporary PIN here?

src/screens/RecoverPin.js
msbrogli: Seems good. Thanks!

src/utils.js
```diff
@@ -258,3 +259,123 @@ export const renderValue = (amount, isInteger) => {
...
+    newPin,
+  );
+  // We also need to change the password because on the mobile app we
+  // use the PIN as Password
+  const successChangePassword = hathorLib.wallet.changePassword(
+    oldPin,
+    newPin,
+  );
+
+  return successChangePassword && successChangePin;
```
msbrogli, line 380: What if one is successful and the other fails? I think we should handle this situation here.

msbrogli (changes requested):

src/screens/RecoverPin.js
andreabadesso: We should! Done in 291d59f80ff01a90d9d0c56ffe4caf4ce07b26f3

src/utils.js
andreabadesso: Hmm, you are right, this would be a huge problem because it would lead to the same problem we have just fixed

But what do you think we should do here?

Maybe rolling back to the `oldPin` and then throwing an error that we can catch on sentry

src/reducer.js
```diff
@@ -69,6 +69,8 @@ const initialState = {
   tokenMetadata: {},
   metadataLoaded: false,
   uniqueDeviceId: null,
+  recoveringPin: false,
```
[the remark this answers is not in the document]
andreabadesso: Done in 42a2159720487aa0a21ba45478e984cf9040503e

src/screens/PinScreen.js
andreabadesso: We could, but I think that having those two actions being dispatched make it more explicit and easier to read, without compromising performance, so I see no downside of keeping both

But I can refactor it if you prefer having a single action

src/utils.js
andreabadesso: This should be fixed now with https://github.com/HathorNetwork/hathor-wallet-lib/pull/296

Can you please check again?

src/screens/RecoverPin.js
msbrogli: Isn't it better to automatically do it when setting recovering pin to false?

src/utils.js
msbrogli: Great! Thanks!

src/screens/PinScreen.js
msbrogli: I'm mostly concerning on misusage of these actions. But a minor change can make it safe enough for me. I commented about it in another conversation.

src/reducer.js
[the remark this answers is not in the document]
msbrogli: Thanks!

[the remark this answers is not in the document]
msbrogli: Sorry! I couldn't find the comments.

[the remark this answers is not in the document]
andreabadesso: https://github.com/HathorNetwork/hathor-wallet-mobile/pull/136/commits/42a2159720487aa0a21ba45478e984cf9040503e

src/screens/RecoverPin.js
andreabadesso: Great idea, fixed in 2eb4cc0d0b7d1ddae8960fb4b6a1b2ce6e95a47e

src/reducer.js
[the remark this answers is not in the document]
msbrogli: Thanks!

src/screens/RecoverPin.js
msbrogli: Great! Thanks!

msbrogli (approved):

PR #139, opened 2021-12-12
feat: added websocket support for wallet service facade

andreabadesso:
### Acceptance Criteria

- The new wallet service facade should receive new transactions and transactions updates in real-time
- The new wallet service facade should properly update balances and transaction history when a new tx is received or a tx is updated

### Security Checklist

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -358,6 +358,27 @@ export const startWallet = (words, pin) => async (dispatch) => {
     }
   });
 
+  const handlePartialUpdate = async (updatedBalanceMap) => {
+    const tokens = Object.keys(updatedBalanceMap);
+    const tokensHistory = {};
+    const tokensBalance = {};
+
+    for (const token of tokens) {
+      /* eslint-disable no-await-in-loop */
+      const balance = await wallet.getBalance(token);
+      const tokenBalance = balance[0].balance;
+      tokensBalance[token] = {
+        available: tokenBalance.unlocked,
+        locked: tokenBalance.locked,
+      };
+      const history = await wallet.getTxHistory({ token_id: token });
+      tokensHistory[token] = history.map((element) => mapTokenHistory(element, token));
+      /* eslint-enable no-await-in-loop */
+    }
+
+    dispatch(partiallyUpdateHistoryAndBalance({ tokensHistory, tokensBalance }));
+  };
+
   wallet.start({ pinCode: pin, password: pin }).then((serverInfo) => {
     walletUtil.storePasswordHash(pin);
     walletUtil.storeEncryptedWords(words, pin);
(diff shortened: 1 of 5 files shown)


src/actions.js
```diff
@@ -418,7 +466,7 @@ export const fetchNewTxTokenBalance = async (wallet, tx) => {
     return null;
   }
   const updatedBalanceMap = {};
-  const balances = wallet.getTxBalance(tx);
+  const balances = await wallet.getTxBalance(tx);
```
pedroferreira1, line 469: The old facade `getTxBalance` method is not `async`, will this continue working for the old facade?

src/actions.js
```diff
@@ -350,52 +366,84 @@ export const startWallet = (words, pin) => async (dispatch) => {
...
 
-      wallet.on('update-tx', (tx) => {
-        fetchNewTxTokenBalance(wallet, tx).then((updatedBalanceMap) => {
-          if (updatedBalanceMap) {
-            dispatch(updateTx(tx, updatedBalanceMap));
+          // Since we are not keeping track of the tokens balance and history
+          // on the wallet service facade, we need to fetch it from the server
+          // on every new transaction received
+          if (useWalletService) {
+            handlePartialUpdate(updatedBalanceMap);
```
pedroferreira1, line 405: This methods calculates the balance of each token just like `fetchNewTxTokenBalance` and you update the redux on `dispatch(newTx)` above. Why do you need it?

src/actions.js
andreabadesso: Yes it will, just tested it

src/actions.js
andreabadesso: You are right, this is a bug actually...

The `onNewTx` reducer will try to update the balances using the `getTxBalance` method from the old facade that is synchronous, but the `getTxBalance` from the wallet service is asyncronous. I'm keeping the `onNewTx` call because it is needed for the request payment feature, but returning before the balance calculation in case we are using the wallet service facade, and using the new `handlePartialUpdate` method for it.

Done in 53e2975ac7c52a0c6195969184122ff8b5f823b

andreabadesso: Done in 53e2975ac7c52a0c6195969184122ff8b5f823b

src/actions.js
```diff
@@ -350,52 +366,81 @@ export const startWallet = (words, pin) => async (dispatch) => {
...
-      });
-    }
+    wallet.conn.on('state', (state) => {
+      let isOnline;
+      if (state === Connection.CONNECTED) {
+        isOnline = true;
+      } else {
+        isOnline = false;
+      }
+      dispatch(setIsOnline(isOnline));
```
r4mmer, line 436: `dispatch(setIsOnline(state == Connection.CONNECTED));` ? Would this work?

src/actions.js
```diff
@@ -350,52 +366,81 @@ export const startWallet = (words, pin) => async (dispatch) => {
...
-      wallet.conn.on('best-block-update', (height) => {
-        fetchNewHTRBalance(wallet).then((data) => {
-          if (data) {
-            dispatch(updateHeight(height, data));
+    wallet.on('update-tx', (tx) => {
+      fetchNewTxTokenBalance(wallet, tx).then((updatedBalanceMap) => {
+        if (updatedBalanceMap) {
+          dispatch(updateTx(tx, updatedBalanceMap));
+
+          // Read comment on new-tx event
```
r4mmer, line 413: What comment?

r4mmer (approved):

src/actions.js
pedroferreira1: Why don't you change `getTxBalance` method to be async so you don't need to do specific code for both facades? Both facades are supposed to implement the same interface, so we don't need to do things so differently.

andreabadesso: That won't work because reducers need to be pure functions, so they can't be async

`getTxBalance` currently needs to be async on the Wallet Service because we fetch the wallet's addresses from the Wallet Service every time we call it.

We could refactor it to keep the addresses in memory and keeping it updated somehow, this way `getTxBalance` could work without being async.

That would be the only way to maintain compatibility for both facades on the reducer

What do you think?

pedroferreira1 (approved):

src/actions.js
andreabadesso: It would! Thanks

Done in 4e4f7c7c57df45526b1e89847ca9befdce654b94

src/actions.js
andreabadesso: This code has been refactored, this comment no longer exists

PR #142, opened 2022-01-11
refactor: passing a method to show pin screen for result to the new facade

andreabadesso:
### Acceptance Criteria
- The wallet mobile should use the new wallet service facade signature, passing an async method to return the pin screen when requested from the wallet-lib

### Security Checklist


src/screens/SendConfirmScreen.js
```diff
@@ -95,16 +101,11 @@ class SendConfirmScreen extends React.Component {
...
-        // eslint-disable-next-line react/jsx-indent
-        (<SendTransactionFeedbackModal
-          text={t`Your transfer is being processed`}
-          sendTransaction={sendTransaction}
-          promise={promise}
-          successText={<TextFmt>{t`Your transfer of **${this.amountAndToken}** has been confirmed`}</TextFmt>}
-          onDismissSuccess={this.exitScreen}
-          onDismissError={() => this.setState({ modal: null })}
-        />),
+      modal: {
```
luislhl, line 104: We still have this method setting the state in the old format. Should we update it?

```js
  onError = (message) => {
    this.setState({
      modal:
        // eslint-disable-next-line react/jsx-indent
        <FeedbackModal
          icon={<Image source={errorIcon} style={{ height: 105, width: 105 }} resizeMode='contain' />}
          text={message}
          onDismiss={() => this.setState({ modal: null })}
        />,
    });
  }
```

src/screens/CreateTokenConfirm.js
```diff
@@ -163,7 +162,21 @@ class CreateTokenConfirm extends React.Component {
...
+          // eslint-disable-next-line react/jsx-indent
+          <SendTransactionFeedbackModal
+            text={this.state.modal.text}
+            sendTransaction={this.state.modal.sendTransaction}
+            promise={this.state.modal.promise}
+            successText={<TextFmt>{t`**${this.name}** created successfully`}</TextFmt>}
+            onTxSuccess={this.onTxSuccess}
+            onDismissSuccess={this.exitScreen}
+            onDismissError={() => this.setState({ modal: null })}
+            hide={this.props.isShowingPinScreen}
```
luislhl, line 176: What's the advantage of hiding it like this instead of just not instantiating the component here?
Is there one?

Something like changing the condition above:

```js
{ this.state.modal && !this.props.isShowingPinScreen && (
```

src/actions.js
```diff
@@ -301,6 +312,23 @@ export const startWallet = (words, pin) => async (dispatch) => {
   // Set useWalletService on the redux store
   dispatch(setUseWalletService(useWalletService));
 
+  const showPinScreenForResult = async () => new Promise((resolve) => {
+    const params = {
+      cb: (_pin) => {
+        dispatch(setIsShowingPinScreen(false));
+        resolve(_pin);
+      },
+      canCancel: true,
```
luislhl, line 321: What does this mean? Should we `setIsShowingPinScreen(false)` if this is cancelled as well?

src/actions.js
```diff
@@ -301,6 +312,23 @@ export const startWallet = (words, pin) => async (dispatch) => {
...
+      screenText: t`Enter your 6-digit pin to authorize operation`,
+      biometryText: t`Authorize operation`,
+    };
+
+    NavigationService.navigate('PinScreen', params);
+
+    // We should set the global isShowingPinScreen
+    dispatch(setIsShowingPinScreen(true));
```
luislhl, line 329: Why do we need to manually hide the other component when navigating to `PinScreen`?

Shoudn't the navigation lib handle the stacking of screens on top of each other for us, which would hide the previous screen? Then just unstack it after it's done?

I'm just curious, I don't really know this lib, but that's how I thought them worked.

src/screens/CreateTokenConfirm.js
```diff
@@ -68,7 +69,11 @@ class CreateTokenConfirm extends React.Component {
    *
    * @param {String} pinCode User PIN
    */
-  executeCreate = (pin) => {
+  executeCreate = async (pin) => {
+    if (this.props.useWalletService) {
+      await this.props.wallet.validateAndRenewAuthToken(pin);
```
obiyankenobi, line 74: We probably don't need to wait.

src/screens/SendConfirmScreen.js
```diff
@@ -79,12 +80,17 @@ class SendConfirmScreen extends React.Component {
...
    * @param {String} pin User PIN already validated
    */
-  executeSend = (pin) => {
+  executeSend = async (pin) => {
     const outputs = [{ address: this.address, value: this.amount, token: this.token.uid }];
     let sendTransaction;
 
     if (this.props.useWalletService) {
-      sendTransaction = new hathorLib.SendTransactionWalletService(this.props.wallet, { outputs });
+      await this.props.wallet.validateAndRenewAuthToken(pin);
```
obiyankenobi, line 88: We probably don't need to wait as we're just updating the token, which is still valid.

obiyankenobi (approved):

src/screens/SendConfirmScreen.js
andreabadesso: The problem here is that it might be expired as we don't schedule token renewals, we only check it when we are about to use it.

So if the user left the app open for 30 minutes and tries to create a transaction, it will ask for his pin twice: When signing inputs and on the renew token method

That's why I added the await there

obiyankenobi: Got it. Can we do a local check if the token is still valid before making this call?

src/screens/CreateTokenConfirm.js
obiyankenobi: Same comment as the other. I'll concentrate the discussion there.

src/NavigationService.js
```diff
@@ -0,0 +1,21 @@
+import { NavigationActions } from 'react-navigation';
```
pedroferreira1, line 1: I'm not sure if it's that relevant but we should keep the initial file standard we have in the repo.

src/screens/SendConfirmScreen.js
pedroferreira1: I don't like having that much logic in the app. Maybe we could create a method `renewIfNeeded` or something like that in the lib

src/components/SendTransactionFeedbackModal.js
```diff
@@ -183,6 +187,7 @@ SendTransactionFeedbackModal.propTypes = {
   onDismissSuccess: PropTypes.func,
   // optional method to be executed when the error modal is dismissed
   onDismissError: PropTypes.func,
+  hide: PropTypes.bool,
```
pedroferreira1, line 190: Missing parameter description

Should we add a default value of `false` to it?

src/actions.js
andreabadesso: I agree that the lib should handle the stacking of screens, but it is not doing that atm, maybe it is a bug on the version we are using as we are on an old version:

image

src/screens/SendConfirmScreen.js
andreabadesso: @pedroferreira1 

Currently `validateAndRenewAuthToken` is only issuing a new token when the current one is either expired or will expire in less than 60s, so it is already a "renew if needed"

I guess that what @obiyankenobi is suggesting is to renew the token every time the user inserts his PIN on the app to reduce the number of times we ask for user input

src/NavigationService.js
```diff
@@ -0,0 +1,21 @@
...
+
+let _navigator;
+
+function setTopLevelNavigator(navigatorRef) {
+  _navigator = navigatorRef;
+}
+
+function navigate(routeName, params) {
```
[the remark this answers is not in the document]
andreabadesso: Yes.

We are calling `setTopLevelNavigator` on app startup (App.js, L#347), so it won't be a problem for now

It might be a problem for instance when we implement push notifications that do something when they are clicked, like navigating to a specific screen (tx detail maybe). We could then implement some kind of queue that will be called in order after the `setTopLevelNavigator` method is called. But I think the current implementation is OK for now. What do you think?

src/NavigationService.js
```diff
@@ -0,0 +1,21 @@
+import { NavigationActions } from 'react-navigation';
+
+let _navigator;
+
+function setTopLevelNavigator(navigatorRef) {
```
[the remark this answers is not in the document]
andreabadesso: Done in daf86120ad973e5aa86b7418506ead0f5458a990

src/components/SendTransactionFeedbackModal.js
andreabadesso: We should.

Done in cde2e017b02cd1042e5dfa3165d44dd47ef76925

src/actions.js
andreabadesso: We could pass a callback to `showingPinScreen` to false, but I think that we should force the user to input his PIN so his token is refreshed

Done in 26155acad255c2a0b45acfc291c9531212b72f3f

What do you think?

src/screens/SendConfirmScreen.js
andreabadesso: `onError` is not being used on `SendConfirm` screen, errors are being handled on the `SendTransactionFeedbackModal` (by listening to `promise` rejections)

But it was being used on `CreateTokenConfirm`, I refactored it on 30bd4d94d73ce47ce41d4a6e0cf8e58b8a9754bf

What do you think?

src/screens/SendConfirmScreen.js
pedroferreira1: As far as I understood you are awaiting here because you don't know if you need to renew it or not and @obiyankenobi suggested to add a local check to await only in case it's needed. This local check that I don't like having in the app and I think should be in the lib. Not sure if you added it already but that's what I meant.

src/NavigationService.js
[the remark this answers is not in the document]
pedroferreira1: Yeah, I agree it's good

src/components/SendTransactionFeedbackModal.js
pedroferreira1: Still missing the parameter description

src/screens/SendConfirmScreen.js
obiyankenobi: Correct. Moreover, I suggest we renew the token (non blocking) every time the user inputs the pin.

src/components/SendTransactionFeedbackModal.js
andreabadesso: Done in 84ca74dacfadedc7450430cae26ca19bfe300a69

src/screens/SendConfirmScreen.js
andreabadesso: @pedroferreira1 @obiyankenobi 

`validateAndRenewAuthToken` will resolve without going to the wallet service if the token is still valid, so I don't see any reason to add this additional local check

In regards to renewing the token every time the user inputs the PIN, I agree and I will refactor the `validateAndRenewAuthToken` method to do this automatically (and non-blocking) everytime it receives a user-inputted PIN

This refactor will be done on the https://github.com/HathorNetwork/hathor-wallet-lib/pull/307 PR as it will be transparent for the wallet mobile

pedroferreira1 (approved):

src/screens/SendConfirmScreen.js
andreabadesso: @obiyankenobi @pedroferreira1 

Changes made on this line:
https://github.com/HathorNetwork/hathor-wallet-lib/pull/307/files#diff-6fa02af64d4fc88536c6977812dc0ee699b0d49377e7e85e7a5df5e274876bb9R634

PR #166, opened 2022-05-06
feat: ios changes to make app work with new react native version

pedroferreira1:
### Summary

I will list here every problem I found and the solutions:

#### Build in new M1 macOS

I was having many different errors that were fixed when I executed the `pod install` with the CPU prefix.

**`arch -x86_64 pod install`**

#### Icon not showing

The icons were not showing in the menubar in the app.

**Fix:** Add `tabbar.tff`  to `Build Phases > Copy Bundle Resources`.

#### Camera error when opening QRCode reader

I was getting a permission error when trying to access the camera.

**Fix:** Install new package: `npm install react-native-permissions --save`

Add the lines below to `ios/Podfile`

```
  # The react-native-qrcode-scanner requires installing react-native-permissions now
  # https://github.com/moaazsidat/react-native-qrcode-scanner#react-native-permissions
  # The two lines below are required by the installation of react-native-permissions
  # https://github.com/zoontek/react-native-permissions#ios
  permissions_path = '../node_modules/react-native-permissions/ios'
  pod 'Permission-Camera', :path => "#{permissions_path}/Camera"
```

#### TabView with a blank space in the bottom (ReceiveScreen):

The same issue described here: https://github.com/satya164/react-native-tab-view/issues/1342

**Fix:** Downgrade `react-native-pager-view` to `v5.4.11`. Before was using `v5.4.15`.

#### Splash screen

There was no splash screen, only the react-native default one.

**Fix:** copy the previous launch screen file we had in the project and substitute the current launch screen.

#### Token List

This list was with the wrong padding for some new iPhone screen models.

**Fix:** consider the correct padding for new iPhone models. See changes in this file `src/components/TokenSelect.js`.

#### Keyboard opened after creating a token

The keyboard from the token name screen opens after the token creation message, i.e. when we open the Token detail screen.

This was the only issue that I didn't fix, the reason is detailed below.

I was receiving a lot of deprecation errors when using the app: `"ERROR currentlyFocusedField is deprecated and will be removed in a future release. Use currentlyFocusedInput"`

I noticed that this error was being generated by `react-navigation` package. The second step of the upgrade dependencies project is to upgrade this package. I manually changed the files inside this package that had the deprecated code `node_modules/@react-navigation/native/dist/createKeyboardAwareNavigator.js`  and the bug stopped happening.

I feel we should leave this issue for the next upgrade step because naturally, we will need to upgrade this package. The only downside of this is that any user that creates a token will see the keyboard opening on a screen without input. This might be a good reason for us to focus in this next step of upgrade soon.

### Acceptance Criteria
- The app should work as before with the new react native version.

### Security Checklist


src/components/TokenSelect.js
```diff
@@ -30,16 +30,52 @@ const getInsets = () => {
...
   const model = DeviceInfo.getModel();
-  // The values are taken from the list at
-  // https://github.com/react-native-device-info/react-native-device-info/blob/master/ios/RNDeviceInfo/RNDeviceInfo.m#L181-L296
-  if (model.startsWith('iPhone X')) {
-    // This covers IPhone X, XS, XS Max, XR
+  const deviceId = DeviceInfo.getDeviceId();
+
+  // This method will return true for devices with new screen format (introduced in iPhone X)
+  // this is important to get the best padding top here
+  const isNewIphoneScreen = () => {
```
andreabadesso, line 37: Should we move this method to an utils file so we can use elsewhere?

andreabadesso (changes requested):

src/components/TokenSelect.js
r4mmer: I didn't check the implications of [this comment](https://github.com/HathorNetwork/hathor-wallet-mobile/blob/1189aee16e3f7873a0d3fa2ce72339dbd4bd1218/src/components/TokenSelect.js#L87) but i believe not using `SafeAreaView` is the reason we have to manually pad this component.

I couldn't find other instances of this anywhere else, that's why it is contained to this component.
Ideally we should upgrade to use the most recent `SafeAreaView` from react-navigation once we upgrade react-navigation itself, then we can remove this padding. Until then I think we should treat this as a temporary patch and not a feature.

r4mmer (approved):

src/components/TokenSelect.js
pedroferreira1: Done https://github.com/HathorNetwork/hathor-wallet-mobile/pull/166/commits/eaecc1b4c2104b1f670c862a1a1c63b961cc1710

andreabadesso (approved):

PR #172, opened 2022-05-18
fix: reset biometry on pin change

### LINKED ISSUE #170: Biometry issue with resetting pin
During the execution of the app if the pin is changed we cannot send transactions using biometry until the app is closed and opened again.

This behavior can be reproduced by:
1. Open the wallet
1. Change the pin
1. Enable biometry (tested with fingerprint on an Android)
1. Try to send a transaction, it will fail

The error itself seems to change if the biometry was enabled when the pin is changed but if will fail when trying to send the transaction to the mining service either way.

After reproducing the error you can close then open the wallet and if biometry was enabled it will ask for biometry and pin to start.

r4mmer:
### Acceptance Criteria
- Change the Keychain password when changing the pin

This closes #170 

### Security Checklist

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -362,7 +362,7 @@ export const guessPin = async (accessData, progressCb) => {
 
 /**
  * Set the pin on the keychain so the user can use biometry
- * @params {string} pinin
+ * @params {string} pin
  */
 export const setKeychainPin = (pin) => {
   Keychain.setGenericPassword(KEYCHAIN_USER, pin, {
--- a/src/utils.js
+++ b/src/utils.js
@@ -6,13 +6,14 @@
  */
 
 import hathorLib from '@hathor/wallet-lib';
+import * as Keychain from 'react-native-keychain';
 import CryptoJS from 'crypto-js';
 import React from 'react';
 import { t } from 'ttag';
 import { Linking, Platform, Text } from 'react-native';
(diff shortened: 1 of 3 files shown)


andreabadesso (approved):

src/utils.js
```diff
@@ -365,11 +366,18 @@ export const guessPin = async (accessData, progressCb) => {
...
+  const success = hathorLib.wallet.changePinAndPassword({
     oldPin,
     newPin,
     oldPassword: oldPin,
     newPassword: newPin,
-  })
-);
+  });
+  if (success) {
+    Keychain.setGenericPassword(KEYCHAIN_USER, newPin, {
```
pedroferreira1, line 377: I feel we should create an util method in this same file `setKeychainPassword(pin)` that does this code, so we don't repeat this here and in the `startWallet` method. What do you think?

src/utils.js
```diff
@@ -359,17 +360,32 @@ export const guessPin = async (accessData, progressCb) => {
   });
 };
 
+/**
+ * Set the pin on the keychain so the user can use biometry
+ * @params {string} pinin
```
pedroferreira1, line 365: Typo pin

pedroferreira1 (approved):

src/utils.js
r4mmer: Fixed

PR #192, opened 2022-09-29
fix: potential crash when receiving a transaction for a token that is currently loading

andreabadesso:
### Acceptance Criteria
- When the token history is loading, we won't have the `data` array, so we need to have a fallback to it on realtime updates

### Security Checklist

### DIFF
--- a/src/reducer.js
+++ b/src/reducer.js
@@ -247,7 +247,7 @@ const onUpdateTokenHistory = (state, action) => {
       [token]: {
         ...state.tokensHistory[token],
         data: [
-          ...state.tokensHistory[token].data,
+          ...get(state.tokensHistory, `${token}.data`, []),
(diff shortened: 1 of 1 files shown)


tuliomir (approved):

src/reducer.js
```diff
@@ -247,7 +247,7 @@ const onUpdateTokenHistory = (state, action) => {
       [token]: {
         ...state.tokensHistory[token],
         data: [
-          ...state.tokensHistory[token].data,
+          ...get(state.tokensHistory, `${token}.data`, []),
```
pedroferreira1, line 250: If you do this and we download the history later, we will end up having this tx duplicated, or I am misunderstanding the events flow?

pedroferreira1 (approved):

PR #203, opened 2022-11-01
fix: invalidating history on wallet reload

andreabadesso:
### Acceptance Criteria
- We should invalidate the token history after a connection reload so we make sure we didn't lose transactions

### Security Checklist


r4mmer (approved):

src/sagas/wallet.js
```diff
@@ -502,10 +508,42 @@ export function* onWalletReloadData() {
...
+  // Wait until the wallet is ready
+  yield take(types.WALLET_STATE_READY);
+
+  try {
+    const registeredTokens = yield call(loadTokens);
+
+    const customTokenUid = DEFAULT_TOKEN.uid;
+    const htrUid = hathorLibConstants.HATHOR_TOKEN_CONFIG.uid;
+    // We might have lost transactions during the reload, so we must invalidate the
+    // token histories:
```
pedroferreira1, line 527: I would just add a comment here explaining why you skip the invalidation for HTR and default token. As far as I understood, the other reload wallet method already handles balance/history for them, right?

Also the names seem a bit confusing to me:

`onWalletReloadData` and `walletReloading`

Maybe this new one should be called `beforeWalletReload`. What do you think?

andreabadesso: Actually, we should activate this saga when `reload` is emitted from the facade as well. 

Refactored this in f7f70a20861de3573f1141cd21201b8aa02fa115

@r4mmer can you please review again?

src/sagas/wallet.js
```diff
@@ -543,11 +530,11 @@ export function* saga() {
   yield all([
     takeLatest('START_WALLET_REQUESTED', startWallet),
     takeLatest('WALLET_CONN_STATE_UPDATE', onWalletConnStateUpdate),
-    takeLatest('WALLET_RELOADING', walletReloading),
+    takeLatest('WALLET_RELOADING', onWalletReloadData),
+    takeLatest('WALLET_RELOAD_DATA', onWalletReloadData),
```
r4mmer, line 534: Didn't we remove this action?

andreabadesso: Fixed in ca011434a77409fb5a8b22e73f8c3f66dd03e7e6

pedroferreira1 (approved):

pedroferreira1 (approved):

r4mmer (approved):

PR #204, opened 2022-11-02
fix: QRCode address processing

### LINKED ISSUE #141: Allow to read QRCode with only the address
The following line of the utils method requires the qrcode with address to be in the format `hathor:<ADDRESS>` https://github.com/HathorNetwork/hathor-wallet-mobile/blob/master/src/utils.js#L153

We should study the problems (if there are any) of allowing the qrcode reader to accept only `<ADDRESS>` directly

alexruzenhack:
- Closes #141

### Acceptance Criteria
- add possibility to process address without 'hathor:<address>' format

### Test Proof
https://www.loom.com/share/c10c9c3b64e84f0fa3bde385c9b066f4

### Security Checklist


tuliomir (commented):
I felt it a bit weird that the `.eslintrc` file was changed in a PR that has no relation to it, and an automatically generated file `gradlew.bat` also affected.

Maybe move this linebreak refactoring to another PR to bring more light to it? I think it's good to have it in a well defined pattern, even in the linter.

.eslintrc
```diff
@@ -41,6 +41,6 @@
     "no-use-before-define": 0,
     "no-unused-vars": [1, {"vars": "all", "args": "none"}],
     "no-underscore-dangle": 0,
-    "no-restricted-syntax": ["error", "ForInStatement", "LabeledStatement", "WithStatement"],
+    "no-restricted-syntax": ["error", "ForInStatement", "LabeledStatement", "WithStatement"]
```
tuliomir, line 44: since you're not changing this rule, I would suggest reverting this single character change too.

tuliomir (approved):
I think it would be better to separate the `gitattributes` and `gradlew.bat` changes into another PR, but they're also small enough to be made here.

I would just suggest to highlight this on the PR description / acceptance criteria

src/utils.js
```diff
@@ -200,6 +194,23 @@ export const parseQRCode = (data) => {
...
+ * 'hathor:<address>' -> '<address>'
+ *
+ * @example
+ * '<address>' -> '<address>'
+ *
+ * @param {string} plainText containing wallet address.
+ */
+function extractAddress(plainText) {
+  const [addressAlone, addressFromComposition] = plainText.split(':');
+  if (addressFromComposition) return addressFromComposition;
```
pedroferreira1, line 210: I think we should keep the same validation we previously had in this case:

```
if (addressParts[0] !== 'hathor' || addressParts.length !== 2) {
    return {
      isValid: false,
      error: 'This QR code does not contain a Hathor address or payment request.',
    };
  }
```

alexruzenhack: This piece of code removes the ability to accept a simple address, the one without the `hathor:` prefix. What I understood from the issue was that we want to allow this simple address to be recognized. Did I understand it wrong?

pedroferreira1: No, you understood correctly. However, when we have a string with `:`, we must validate that the first part is `hathor` and we have only two parts.

alexruzenhack: Got it!

pedroferreira1 (approved):

pedroferreira1 (approved):

src/utils.js
```diff
@@ -204,11 +208,37 @@ export const parseQRCode = (data) => {
  * '<address>' -> '<address>'
  *
  * @param {string} plainText containing wallet address.
+ *
+ * @typedef ExtractedAddress
```
tuliomir, line 212: From my experience, static code analyzers cannot read a `@typedef` within the same comment block as a function. No warning is shown, but the autocomplete features do not work.

Maybe it would be better to make it two comment blocs: one for the `@typedef` and another one for the function, with its `@example`s and `@return`. Or just describe the properties directly on the `@return`, without declaring a new type since it's not used elsewhere.

Also, both `error` and `address` are optional properties of this type, but are declared as mandatory.

PR #209, opened 2022-11-30
fix: use spawn instead of fork

andreabadesso:
### Acceptance Criteria
- Token metadata should download without blocking the `loadTokens` generator

### Security Checklist

### DIFF
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -28,6 +28,7 @@ import {
   race,
   take,
   fork,
+  spawn,
 } from 'redux-saga/effects';
 import { eventChannel } from 'redux-saga';
 import { getUniqueId } from 'react-native-device-info';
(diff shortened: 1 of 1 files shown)


src/sagas/wallet.js
```diff
@@ -249,7 +250,7 @@ export function* loadTokens() {
     }, []);
 
   // We don't need to wait for the metadatas response, so just fork it
```
luislhl, line 252: Should this comment be updated?

tuliomir (approved):

src/sagas/wallet.js
andreabadesso: Actually, the comment made more sense now than it did before the change, `spawn` is just a "detached" version of `fork`

But I updated it to make the difference clearer on 19fe352ea625c1ef35c58e62df1865497d7629ee

PR #212, opened 2022-12-11
fix: restart after wallet service error

andreabadesso:
### Acceptance Criteria
- We should handle errors in the `startWallet` saga, allowing the user to retry the request
- We should fix an error where the wallet would break (and reset) when initializing on the wallet-service
  - The `walletUtils.reloadElectron` method was broken, I removed it and added a mechanism for starting the wallet using the saga again

### Security Checklist


src/sagas/wallet.js
```diff
@@ -325,7 +342,7 @@ export function* listenForFeatureFlags(featureFlags) {
       const oldUseWalletService = yield select((state) => state.useWalletService);
 
       if (oldUseWalletService !== newUseWalletService) {
-        NativeModules.HTRReloadBundleModule.restart();
```
r4mmer, line 328: This removes the need for `HTRReloadBundleModule`? Because we have an implementation of this module in both java and Obj-C, maybe we can remove those files.

r4mmer (commented):
We have tests for nodejs v8, 10, 12 and 14, are those necessary?
From what i remember we are using v12 on the mobile (I couldn't find any files that say which version should be used when building/running)

src/sagas/wallet.js
andreabadesso: We can! Thanks

9172518d9e80c9ba05703fdf4b18285ca20ab958

r4mmer (approved):

src/sagas/wallet.js
```diff
@@ -86,7 +86,7 @@ export function* startWallet(action) {
 
   // If we've lost redux data, we could not properly stop the wallet object
   // then we don't know if we've cleaned up the wallet data in the storage
-  walletUtil.cleanLoadedData();
+  walletUtil.cleanLoadedData({ cleanAccessData: false });
```
pedroferreira1, line 89: Maybe explain why we can't clean access data here?

Mainly because this is now different from the desktop wallet, where we have:

```
    oldWalletUtil.cleanLoadedData({ cleanAccessData: !fromXpriv});
```

But we clean loaded data only if it's not wallet service. It's important to understand why they are different.

src/sagas/helpers.js
```diff
@@ -55,3 +55,21 @@ export function* dispatchAndWait(action, successAction, failureAction) {
     falure: take(failureAction),
   });
 }
+
+/**
+ * Handles errors thrown from the main saga (started with call) by yielding
+ * an action passed as a parameter
+ *
+ * @param action - The action to call
```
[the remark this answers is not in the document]
andreabadesso: Fixed in 7578021d7082b6a534c9c4da6a19a1f1abb2be11

src/sagas/wallet.js
andreabadesso: Improved the comment here, explaining why I am not cleaning the accessData anymore

ce09e21c0bbf54758cb93881d8f23c71c60a7205

pedroferreira1 (approved):

PR #213, opened 2022-12-19
feat: push notification settings

alexruzenhack:
- fix: QRCode address processing
- feat: add PushNotification page to settings
- feat: (wip) implement push notification switch
- feat: add PushNotification page for settings

### Acceptance Criteria

### Security Checklist


src/screens/PushNotification.js
```diff
@@ -0,0 +1,210 @@
...
+    this.props.pushFirstTimeRegistration(this.props.pushNotification.deviceId);
+  }
+
+  onShowAmountSwitchChange = (showAmountEnabled) => {
+    const settings = {
+      ...getPushNotificationSettings(this.props.pushNotification),
+      showAmountEnabled,
+      deviceId: this.props.pushNotification.deviceId,
+    };
+    console.log('settings', settings);
```
andreabadesso, line 142: Remove?

src/screens/PushNotification.js
```diff
@@ -0,0 +1,210 @@
...
+  },
+});
+
+const mapInvoiceDispatchToProps = (dispatch) => ({
+  pushApiReady: () => dispatch(pushApiReady()),
+  pushFirstTimeRegistration: (deviceId) => dispatch(pushFirstTimeRegistration({ deviceId })),
+  pushUpdateRequested: (settings) => dispatch(pushUpdateRequested(settings)),
+});
+
+class PushNotification extends React.Component {
```
andreabadesso, line 49: Should we use function components instead of class components so we can slowly transition to more modern React?

src/screens/Settings.js
```diff
@@ -106,13 +106,17 @@ export class Settings extends React.Component {
...
-)}
+              )}
               isFirst
             />
             <ListMenu
               title={t`Security`}
               onPress={() => this.props.navigation.navigate('Security')}
             />
+            <ListMenu
+              title='Push Notification'
```
andreabadesso, line 117: Missing translation on this string?

src/App.js
```diff
@@ -208,6 +211,7 @@ const mapDispatchToProps = (dispatch) => ({
   setTokens: (tokens) => dispatch(setTokens(tokens)),
   lockScreen: () => dispatch(lockScreen()),
   resetData: () => dispatch(resetData()),
+  pushInit: (payload) => dispatch(pushInit(payload)),
```
andreabadesso, line 214: Remove (explained in next comments)

src/actions.js
```diff
@@ -501,3 +516,93 @@ export const walletStateReady = () => ({
 export const onWalletReload = () => ({
   type: types.WALLET_RELOADING,
 });
+
+// Push notification actions
+
+/**
+ * @param {{deviceId: string, settings: { enabled, showAmountEnabled }, hasBeenEnabled: boolean, enabledAt: number}} payload
+ */
+export const pushInit = (payload) => ({
```
andreabadesso, line 525: Remove (Explained in next comments)

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,412 @@
...
+    if (key === NEW_TRANSACTION_RECEIVED_DESCRIPTION_WITH_TOKENS) {
+      if (!args) {
+        console.log('args is null for key', key);
+        return '';
+      }
+      const countArgs = args.length;
+      if (countArgs === 3) {
+        const [firstToken, secondToken, other] = args;
+        const otherCount = parseInt(other, 10);
+        message = t`you have received ${firstToken}, ${secondToken} and ${otherCount} other token on a new transaction.`;
```
andreabadesso, line 108: Should `You` be captalized?

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,412 @@
...
+ *   bodyLocKey: 'new_transaction_received_description_with_tokens',
+ *   titleLocKey: 'new_transaction_received_title',
+ *   txId: 'txId1',
+ * }
+ * @inner
+ */
+const messageHandler = async (message) => {
+  const { data } = message;
+  if (!localization.hasKey(data.titleLocKey)) {
+    console.log('unknown message titleLocKey', data.titleLocKey);
```
andreabadesso, line 148: Remove comment?

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,412 @@
...
+  notifee.displayNotification({
+    title,
+    body,
+    android: {
+      channelId: TRANSACTION_CHANNEL_ID,
+    },
+  });
+};
+
+const showPinScreenForResult = async (dispatch) => new Promise((resolve) => {
```
andreabadesso, line 170: Maybe export this as an util somewhere and re-use on `sagas/wallet.ts` and here instead of duplicating code

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,412 @@
...
+      hasBeenEnabled,
+      enabledAt
+    }
+  } = action;
+  const deviceId = yield call(getDeviceId);
+
+  // If the deviceId is different from the persisted one, we should update it
+  if (persistedDeviceId && persistedDeviceId !== deviceId.toString()) {
+    STORE.setItem(pushNotificationKey.deviceId, deviceId);
+    put(pushUpdateDeviceId({ deviceId }));
```
andreabadesso, line 208: Is this working? Shouldn't it be:

```javascript
yield put(pushUpdateDeviceId({ deviceId });
```

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,412 @@
...
+  // Add listeners for push notifications on foreground and background
+  messaging().onMessage(onForegroundMessage);
+  messaging().setBackgroundMessageHandler(onBackgroundMessage);
+
+  // Check if the last registration call was made more then a week ago
+  if (hasBeenEnabled) {
+    const timeSinceLastRegistration = moment().diff(enabledAt, 'weeks');
+    if (timeSinceLastRegistration > 1) {
+      // Update the registration, as per Firebase's recommendation
+      put(pushRegistrationRequested({
```
andreabadesso, line 231: Same here

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,412 @@
...
+  yield put(pushLoadWalletRequested());
+
+  // wait for the wallet to be loaded
+  const [loadWalletSuccess, loadWalletFail] = yield race([
+    take(types.PUSH_WALLET_LOAD_SUCCESS),
+    take(types.PUSH_WALLET_LOAD_FAILED)
+  ]);
+
+  if (loadWalletFail) {
+    yield put(pushRegisterFailed());
```
andreabadesso, line 342: Should we return here?

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,412 @@
...
```
andreabadesso, line 382: Same here

andreabadesso (changes requested):

src/screens/PushNotification.js
alexruzenhack: I'm not aware of it. Can you give some materials to read?

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,412 @@
...
+    }
+  } catch (error) {
+    console.log('Error updating device: ', error.cause);
+    yield put(pushUpdateFailed());
+  }
+}
+
+export function* saga() {
+  yield all([
+    takeEvery(types.PUSH_INIT, appInitialization),
```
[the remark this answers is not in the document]
alexruzenhack: The challenge of working with Effects is that we can lose some processes from sight. A future maintainer will have no clue that the push notification was initialized when the wallet initializes as well because it will be hidden in the saga middleware.

The way I did the process is explicit in the app initialization, signaling that the push notification exists and the app initialization is the entry point.

src/reducer.js
```diff
@@ -179,6 +215,20 @@ const reducer = (state = initialState, action) => {
       return onStartWalletFailed(state);
     case types.WALLET_BEST_BLOCK_UPDATE:
       return onWalletBestBlockUpdate(state, action);
+    case types.PUSH_INIT:
+      return onPushInit(state, action);
```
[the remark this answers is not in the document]
alexruzenhack: I need a reducer to populate the state and I believe this is appropriate. What do you suggest to replace it? The saw the initialization starting from the effect `fork` suggestion, but I still need to hold the state in the reducer storage.

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,412 @@
...
+const localization = {
+  keys: new Set([
+    NEW_TRANSACTION_RECEIVED_DESCRIPTION_WITH_TOKENS,
+    NEW_TRANSACTION_RECEIVED_DESCRIPTION_WITHOUT_TOKENS,
+    NEW_TRANSACTION_RECEIVED_TITLE
+  ]),
+  hasKey: (key) => localization.keys.has(key),
+  getMessage: (key, args) => {
+    if (!localization.hasKey(key)) {
+      console.log('unknown key', key);
```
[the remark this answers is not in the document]
alexruzenhack: No. But I have improved it to:
```
console.log('Unknown localization key for push notification message.', key);
```

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,412 @@
...
+      return '';
+    }
+
+    let message = '';
+    if (key === NEW_TRANSACTION_RECEIVED_DESCRIPTION_WITH_TOKENS) {
+      if (!args) {
+        console.log('args is null for key', key);
```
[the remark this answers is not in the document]
alexruzenhack: No. But I have improved it to:
```
console.log(`The args for push notification message key ${NEW_TRANSACTION_RECEIVED_DESCRIPTION_WITH_TOKENS} cannot be null or undefined.`, key);
```

src/actions.js
alexruzenhack: I keep this initialization because the push init on saga use this action to update the redux store.

src/reducer.js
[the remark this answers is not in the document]
alexruzenhack: I kept to let the saga use it to update the redux store.

src/sagas/pushNotification.js
alexruzenhack: Show we avoid to log?

src/sagas/pushNotification.js
alexruzenhack: Done.

src/sagas/pushNotification.js
alexruzenhack: Done.

src/sagas/pushNotification.js
alexruzenhack: I spent some time on the PushNotification refactoring. Can this refactor wait until I finish the main parts yet letting? I can do this refactoring in the next PRs.

src/sagas/pushNotification.js
andreabadesso: I guess it's OK to log, the problem is that it's going nowhere, so it's kind of pointless I guess

src/sagas/pushNotification.js
andreabadesso: I just checked and this is exactly the same code as the one in `src/sagas/wallet.js`, wouldn't it be simple as exporting and using on both places?

I think it's bad to add duplicated code like this, we might forget to refactor it later

andreabadesso (approved):

src/components/ActionModal.js
```diff
@@ -0,0 +1,98 @@
...
+ *  message={this.state.modal.message}
+ *  button={this.state.modal.button}
+ *  onAction={this.state.modal.onAction}
+ *  onDismiss={this.state.modal.onDismiss}
+ * />
+ *
+ * @param {{ title, message, button, onAction, onDismiss }} props The props of the component
+ * @returns The action modal component to display
+ */
+const ActionModal = (props) => (
```
pedroferreira1, line 25: We have a component called `FeedbackModal`, which is used in the whole app. I understand that there is a small difference (the button "I agree") but I don't know if it's needed. To be honest, after the latest wallet update with the terms of service, I feel this modal won't be needed anymore at all, I'm confirming this.

Anyway, there are two clear benefits if we can use the `FeedbackModal` instead of creating this new component, (i) it's easier to maintain the code, (ii) we keep consistency with UI. The latter is the most critical for me because this new modal is a bit different in design compared to the other ones (I think, I didn't check the style but it looked like that in the video).

What do you think?

src/reducer.js
```diff
@@ -93,6 +95,35 @@ const initialState = {
...
+     * showAmountEnabled {boolean} if user has enabled the option to show amount
+     * of token balance on the push notification
+     */
+    showAmountEnabled: false,
+    /**
+     * hasBeenEnabled {boolean} if user has enabled push notification before
+     * this is used to show the terms and conditions modal only the first time
+     * user enables push notification
+     */
+    hasBeenEnabled: false,
```
pedroferreira1, line 121: If we decide to remove the modal, then we can remove this state here

src/reducer.js
```diff
@@ -639,6 +684,89 @@ export const onWalletBestBlockUpdate = (state, action) => {
...
+  ...state,
+  pushNotification: {
+    ...state.pushNotification,
+    apiStatus: PUSH_API_STATUS.READY,
+    enabled,
+    showAmountEnabled,
+  },
+});
+
+export const onPushUpdateFailed = (state) => ({
```
pedroferreira1, line 762: Why do you need `onPushUpdateFailed` and `onPushRegisterFailed` with the same code? Can't we have a `onPushAPIFailed`?

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,432 @@
...
+  READY: 'ready',
+  FAILED: 'failed',
+  LOADING: 'loading',
+};
+
+/**
+ * this is the network name that will be used to load the wallet on the wallet-service,
+ * it is first hardcoded in the `startWallet` saga function, @see src\sagas\wallet.js.
+ */
+const NETWORK = 'mainnet';
```
pedroferreira1, line 56: Maybe we could have this in the constants file and import from there. What do you think? So when we need to change to testnet, then we change only in one place.

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,432 @@
...
+/**
+ * this is the message key for localization of new transaction when show amount is enabled
+ */
+const NEW_TRANSACTION_RECEIVED_DESCRIPTION_WITH_TOKENS = 'new_transaction_received_description_with_tokens';
```
pedroferreira1, line 60: I think I prefer the name `WITH_AMOUNT` and `WITHOUT_AMOUNT`. The current name is misleading and anyone reading the code could interpret is wrong.

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,432 @@
...
+    if (key === NEW_TRANSACTION_RECEIVED_DESCRIPTION_WITH_TOKENS) {
+      if (!args) {
+        console.log(`The args for push notification message key ${NEW_TRANSACTION_RECEIVED_DESCRIPTION_WITH_TOKENS} cannot be null or undefined.`, key);
+        return '';
+      }
+      const countArgs = args.length;
+      if (countArgs === 3) {
+        const [firstToken, secondToken, other] = args;
+        const otherCount = parseInt(other, 10);
+        message = t`You have received ${firstToken}, ${secondToken} and ${otherCount} other token on a new transaction.`;
```
pedroferreira1, line 111: I didn't understand this message. Isn't it suppose to show the amount received? Maybe an example of the expected message would help.

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,432 @@
...
+  getMessage: (key, args) => {
+    if (!localization.hasKey(key)) {
+      console.log('Unknown localization key for push notification message.', key);
+      return '';
+    }
+
+    let message = '';
+    if (key === NEW_TRANSACTION_RECEIVED_DESCRIPTION_WITH_TOKENS) {
+      if (!args) {
+        console.log(`The args for push notification message key ${NEW_TRANSACTION_RECEIVED_DESCRIPTION_WITH_TOKENS} cannot be null or undefined.`, key);
```
pedroferreira1, line 104: Change all `console.log` to `console.debug` maybe?

src/sagas/pushNotification.js
pedroferreira1: I agree this should be reused, and it's better to be in this PR.

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,432 @@
...
+  dispatch(setIsShowingPinScreen(true));
+});
+
+/**
+ * This function is called when the wallet is initialized with success.
+ */
+export function* onAppInitialization() {
+  yield take(types.START_WALLET_SUCCESS);
+
+  const { enabled, showAmountEnabled } = STORE.getItem(pushNotificationKey.settings);
```
pedroferreira1, line 196: Will you have this in the STORE the first time you start the wallet with this new feature? Or `STORE.getItem(pushNotificationKey.settings)` will return `null`?

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,432 @@
...
+  const hasBeenEnabled = STORE.getItem(pushNotificationKey.hasBeenEnabled);
+  const enabledAt = STORE.getItem(pushNotificationKey.enabledAt);
+
+  const persistedDeviceId = STORE.getItem(pushNotificationKey.deviceId);
+  const deviceId = yield call(getDeviceId);
+  // If the deviceId is different from the persisted one, we should update it
+  if (persistedDeviceId && persistedDeviceId !== deviceId.toString()) {
```
pedroferreira1, line 203: Where do you set `deviceId` in the STORE for the first time?

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,432 @@
...
+      id: TRANSACTION_CHANNEL_ID,
+      name: TRANSACTION_CHANNEL_NAME,
+    });
+  }
+
+  // Make sure deviceId is registered on the FCM
+  messaging().registerDeviceForRemoteMessages();
+  // Add listeners for push notifications on foreground and background
+  messaging().onMessage(onForegroundMessage);
+  messaging().setBackgroundMessageHandler(onBackgroundMessage);
```
pedroferreira1, line 232: Can any of those methods fail and break the wallet?

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,432 @@
...
+
+  if (loadWalletFail) {
+    yield put(pushRegisterFailed());
+    return;
+  }
+
+  const { walletService } = loadWalletSuccess.payload;
+  try {
+    const { success } = yield call(pushLib.PushNotification.registerDevice, walletService, {
+      pushProvider: pushLib.PushNotificationProvider.ANDROID,
```
pedroferreira1, line 358: Why is this hardcoded with `ANDROID`?

pedroferreira1:
I'm not sure why the CI didn't fail but it's missing the i18n file updates and the pt-br translation. If you need any help just let me know.

package.json
```diff
@@ -32,7 +35,7 @@
     "process": "^0.11.10",
     "prop-types": "^15.8.1",
     "react": "17.0.2",
-    "react-native": "0.67.4",
+    "react-native": "0.67.5",
```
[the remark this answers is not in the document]
alexruzenhack: Yes, we need. This patch solves a bug: https://github.com/facebook/react-native/issues/35210

src/utils.js
```diff
@@ -445,3 +445,23 @@ export const mapTokenHistory = (element, token) => {
   };
   return new TxHistory(data);
 };
+
+/**
+ * Verifies if user is enabling a feature
+ * @param {boolean} value should be true or false and comes from Switch component
+ * @returns {boolean} true if user is enabling a feature, false otherwise
+ */
+export const isEnablingFeature = (value) => (value && value === true);
```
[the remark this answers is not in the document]
alexruzenhack: This is a value that comes from the interaction with the user, it comes from the switch input, I believe it is wise to check the boolean value.

src/sagas/pushNotification.js
alexruzenhack: Good point. I can add a fallback. But it is working anyway.

src/sagas/pushNotification.js
alexruzenhack: Yes, it can fail. I will wrap then in a try/catch.

src/sagas/pushNotification.js
alexruzenhack: Because it is only available to android, right? But before allowing the settings for push notifications we need to verify the platform. I believe it should be the approach.

alexruzenhack:
> I'm not sure why the CI didn't fail but it's missing the i18n file updates and the pt-br translation. If you need any help just let me know.

Yes, I will need help in this matter.

src/reducer.js
alexruzenhack: Good point. I kept them separated hopping they could evolve differently, but this is not the case indeed. I will join them.

src/reducer.js
alexruzenhack: Actually, this is used to trigger the registration when the user enables the push notification for the first time. When the user enables it for the second time if any, the update function is triggered.

src/sagas/pushNotification.js
alexruzenhack: Good!

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,440 @@
...
+      */
+      const countArgs = args.length;
+      if (countArgs === 3) {
+        const [firstToken, secondToken, other] = args;
+        const otherCount = parseInt(other, 10);
+        /**
+         * @example
+         * You have received 10 T2, 5 T1 and 2 other token on a new transaction.
+         */
+        message = t`You have received ${firstToken}, ${secondToken} and ${otherCount} other token on a new transaction.`;
```
pedroferreira1, line 109:
```suggestion
        message = t`You have received ${firstToken}, ${secondToken} and ${otherCount} other ${otherCount === 1 ? 'token' : 'tokens'} on a new transaction.`;
```

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,440 @@
...
+  const { enabled, showAmountEnabled } = STORE.getItem(pushNotificationKey.settings) || { enabled: false, showAmountEnabled: false };
+  const hasBeenEnabled = STORE.getItem(pushNotificationKey.hasBeenEnabled);
+  const enabledAt = STORE.getItem(pushNotificationKey.enabledAt);
+
+  const persistedDeviceId = STORE.getItem(pushNotificationKey.deviceId);
+  const deviceId = yield call(getDeviceId);
+  // If the deviceId is different from the persisted one, we should update it.
+  // The first time the perisistedDeviceId will be null, and the deviceId will be
+  // the one returned by getDeviceId, which gets the deviceId from FCM.
+  if (deviceId && persistedDeviceId !== deviceId) {
```
pedroferreira1, line 222: If `deviceId` is `null`, it means the request to fetch it failed. How is the app handling this? What happens in that case?

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,440 @@
...
+    deviceId,
+    settings: {
+      enabled,
+      showAmountEnabled,
+    },
+    hasBeenEnabled,
+    enabledAt,
+  }));
+
+  yield call(createChannelIfNotExists);
```
pedroferreira1, line 238: If any of these requests below fail, the push notification feature won't work, right? Right now you have a `try...catch` but you are not doing anything to handle it. How will the app handle when those requests fail?

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,440 @@
...
+  let walletService;
+  if (!useWalletService) {
+    // Set urls for wallet service
+    config.setWalletServiceBaseUrl(WALLET_SERVICE_MAINNET_BASE_URL);
+    config.setWalletServiceBaseWsUrl(WALLET_SERVICE_MAINNET_BASE_WS_URL);
+    const network = new Network(NETWORK);
+
+    const pin = yield call(showPinScreenForResult, dispatch);
+    walletService = new HathorWalletServiceWallet({
+      requestPassword: pin,
```
pedroferreira1, line 281: `requestPassword` expects a callback, not the pin. I think this should be `showPinScreenForResult`

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,440 @@
...
+    console.error('Error registering device for the first time: ', error.cause);
+    yield put(pushRegisterFailed());
+  }
+}
+
+/**
+ * This function is responsible for registering the device on the wallet-service in the event
+ * of renewing the registration.
+ */
+export function* registration({ payload: { enabled, showAmountEnabled, deviceId } }) {
```
pedroferreira1, line 349: Except for one line (`STORE.setItem(pushNotificationKey.hasBeenEnabled, true);`), which is ok to be in this method too, isn't this method the same as `firstTimeRegistration`? 

I feel we should have only one of them and remove the duplicated code, what do you think?

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,440 @@
...
+    yield put(pushRegisterFailed());
+  }
+}
+
+/**
+ * This function is responsible for updating the registration of the device on the wallet-service
+ * in the event of changing the settings of the push notifications.
+ * @param {{ payload: { enabled: boolean, showAmountEnabled: boolean, deviceId: string } }} action
+ */
+export function* updateRegistration({ payload: { enabled, showAmountEnabled, deviceId } }) {
```
pedroferreira1, line 390: Another method that is exactly the same as the others (if I'm not mistaken). The three methods should be only one using the correct parameters.

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,440 @@
...
+  }
+
+  const { walletService } = loadWalletSuccess.payload;
+  try {
+    const { success } = yield call(pushLib.PushNotification.updateDevice, walletService, {
+      deviceId,
+      enablePush: enabled,
+      enableShowAmounts: showAmountEnabled,
+    });
+    if (success) {
```
pedroferreira1, line 411: Don't you need to update `enabledAt` in the `STORE` to `0`, in case `enabled` is `false`?

andreabadesso:
Just saw the video, should we display a `Loading` icon until we receive a response from the wallet-service?

locale/da/texts.po
```diff
@@ -8,11 +8,11 @@ msgstr ""
 "X-Generator: POEditor.com\n"
 "Plural-Forms: nplurals=2; plural=(n != 1);\n"
 
-#: src/errorHandler.js:51
+#: src\errorHandler.js:51
```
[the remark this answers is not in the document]
alexruzenhack: Yes. I will generate the translations through the WSL to avoid such things.

src/utils.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-len */
```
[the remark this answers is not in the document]
alexruzenhack: Because of the jsdoc @param in the `getPushNotificationSettings` method. But I refactored it to remove this rule.

src/sagas/pushNotification.js
alexruzenhack: Hmmm. I will halt the push initialization, and in the feature toggle PR I will complement by disabling the push notification.

src/sagas/pushNotification.js
alexruzenhack: Sure! I refactored it to remove first registration. Now we only have registration.

src/sagas/pushNotification.js
alexruzenhack: You right. I did the refactoring.

alexruzenhack:
> Just saw the video, should we display a `Loading` icon until we receive a response from the wallet-service?

Yes. I did a refactor to show loading feedback.

src/screens/PushNotification.js
```diff
@@ -0,0 +1,143 @@
...
+  const dispatch = useDispatch();
+
+  const onPushNotificationSwitchChange = (value) => {
+    const isFirstTime = isEnablingFeature(value) && !hasBeenEnabled;
+    if (isFirstTime) {
+      executeFirstRegistrationOnPushNotification();
+      return;
+    }
+
+    executeUpdateOnPushNotification(value);
```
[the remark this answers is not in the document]
alexruzenhack: I refactored it. This code is no more necessary.

src/screens/PushNotification.js
```diff
@@ -0,0 +1,143 @@
...
+          icon={<Spinner />}
+          text={apiRequestLoadingText}
+        />
+      )}
+
+      {hasPushApiFailed && (
+        <FeedbackModal
+          icon={(<Image source={errorIcon} style={styles.feedbackModalIcon} resizeMode='contain' />)}
+          text={apiRequestFailedText}
+          onDismiss={() => dispatch(pushApiReady())}
```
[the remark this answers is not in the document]
alexruzenhack: Yes! I have only 3 states:
* Ready
* Loading
* Failed

src/reducer.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-len */
```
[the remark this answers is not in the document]
alexruzenhack: Removed it.

src/reducer.js
```diff
@@ -774,6 +763,25 @@ export const onPushApiFailed = (state) => ({
   },
 });
 
+/**
+ * @param {Boolean} action.payload.error The captured Error object
+ * @param {Boolean} action.payload.isFatal Indicates if the error is fatal
+ */
+export const onExceptionCaptured = (state, { payload }) => {
```
andreabadesso, line 770: I guess this does not belong to this PR, it is already in `dev`, can you rebase?

andreabadesso (changes requested):

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,406 @@
...
+ * This function is called when the wallet is initialized with success.
+ */
+export function* onAppInitialization() {
+  yield take(types.START_WALLET_SUCCESS);
+
+  // If the channel is not created, we should not continue.
+  const isChannelCreated = yield call(createChannelIfNotExists);
+  if (!isChannelCreated) {
+    console.debug('Halting push notification initialization because the channel was not created.');
+    return;
```
pedroferreira1, line 236: Should we show an alert error here saying that the push notification won't work until the app is restarted?

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,406 @@
...
+  }
+
+  // If the device is not registered on FCM, we should not continue.
+  const isDeviceRegistered = yield call(confirmDeviceRegistrationOnFirebase);
+  if (!isDeviceRegistered) {
+    console.debug('Halting push notification initialization because the device is not registered on FCM.');
+    return;
+  }
+
+  yield call(installForegroundListener);
```
pedroferreira1, line 246: Should we stop the push notification process here as well if it fails?

src/sagas/pushNotification.js
```diff
@@ -0,0 +1,406 @@
...
+  try {
+    const { success } = yield call(pushLib.PushNotification.registerDevice, walletService, {
+      pushProvider: Platform.OS,
+      deviceId,
+      enablePush: !!enabled,
+      enableShowAmounts: !!showAmountEnabled,
+    });
+
+    if (success) {
+      const enabledAt = Date.now();
```
pedroferreira1, line 369: If `enabled` is `false` we shouldn't update this variable in the local storage, right? Or maybe update to `0`.

src/sagas/pushNotification.js
alexruzenhack: I have implemented the error capture here in the PR for the tx details.
https://github.com/HathorNetwork/hathor-wallet-mobile/pull/233/files#diff-b2f41e02468f99f5c9c39db9c11e7c79fbde4091cf522468d84af47845be7adaR77

src/sagas/pushNotification.js
alexruzenhack: I have implemented the error capture here in the PR for the tx details:
https://github.com/HathorNetwork/hathor-wallet-mobile/pull/233/files#diff-b2f41e02468f99f5c9c39db9c11e7c79fbde4091cf522468d84af47845be7adaR155

src/sagas/pushNotification.js
alexruzenhack: Good point!

pedroferreira1 (approved):

andreabadesso (approved):

PR #214, opened 2022-12-28
fix: handle app state change

andreabadesso:
### Acceptance Criteria
- We should handle properly the active to active state transition on the method `handleAppStateChange`, as described in https://github.com/HathorNetwork/internal-issues/issues/144

### Security Checklist

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -288,16 +288,16 @@ class _AppStackWrapper extends React.Component {
    * More info: https://reactnative.dev/docs/appstate
    */
   _handleAppStateChange = (nextAppState) => {
-    if (nextAppState === 'active') {
-      if (this.appState === 'active') {
-        // As per the Apple lifecycle documentation,  an active → active state
-        // transition should never happen, but we observed this happening in production
-        // builds on a very specific situation, described in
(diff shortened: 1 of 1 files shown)


pedroferreira1 (approved):

msbrogli (approved):

src/App.js
```diff
@@ -288,13 +288,25 @@ class _AppStackWrapper extends React.Component {
    * More info: https://reactnative.dev/docs/appstate
    */
   _handleAppStateChange = (nextAppState) => {
+    if (this.appState === nextAppState) {
+      // As per the Apple lifecycle documentation, an active → active state
+      // transition should never happen, but we observed this happening in production
+      // builds on a very specific situation, described in
+      // https://github.com/HathorNetwork/internal-issues/issues/144 so we should ignore
+      // all transitions that are not changing
+      return;
```
msbrogli, line 297: Should we log a debug message with something like "App transition from {this.appState} to {nextAppState}. This should never happen."?

andreabadesso: We won't see logs from production devices, but I think that it is indeed valid to log at least so we can see it during development!

Done in cdfc255e9f3e47038b1cca9b816a955c069d5782

PR #216, opened 2022-12-30
feat: add button to copy transaction id

### LINKED ISSUE #80: Easily copy transaction id
When we see the tx details, it'd be nice to be able to copy the id easily. We could add a copy icon or do like the address, just clicking on it.

image

obiyankenobi:
### Acceptance Criteria
- add a way to easily copy the transaction id

### Security Checklist

Closes #80

### DIFF
--- a/src/components/TxDetailsModal.js
+++ b/src/components/TxDetailsModal.js
@@ -14,6 +14,7 @@ import { getShortHash, getTokenLabel, renderValue } from '../utils';
 import { ListButton, ListItem } from './HathorList';
 import SlideIndicatorBar from './SlideIndicatorBar';
 import icShareActive from '../assets/icons/icShareActive.png';
+import CopyClipboard from './CopyClipboard';
 
 class TxDetailsModal extends Component {
   style = StyleSheet.create({
@@ -34,6 +35,7 @@ class TxDetailsModal extends Component {
     const idStr = getShortHash(tx.txId, 12);
     const explorerIcon = <Image source={icShareActive} width={24} height={24} />;
     const explorerLink = `https://explorer.hathor.network/transaction/${tx.txId}`;
+    const txIdComponent = <CopyClipboard text={idStr} copyText={tx.txId} textStyle={{ fontSize: 16 }}/>;
     return (
       <Modal
         isVisible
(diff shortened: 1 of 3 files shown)


obiyankenobi:
Opening as draft as there are some decisions to take.

### Option 1

This was the solution I coded, but ran into a setback. I added a button on the right side of the tx id so it could be easily copied. Do not mind the button icon now. It should be updated later if we go with this solution.

Simulator Screen Shot - iPhone SE (3rd generation) - 2022-12-30 at 15 24 50

The problem is there's no proper feedback to the user this was done. There's a subtle feedback on the button, which quickly changes color.

https://user-images.githubusercontent.com/3236756/210110145-92a749bf-7cd7-4a89-8f81-8ad224a6f762.mp4

Ideally, we should have something like a [Toast](https://developer.android.com/guide/topics/ui/notifiers/toasts), but that's only available on Android. It's not worth adding a library just for that.

### Option 2

Another option is doing something similar to the copy feature for addresses.

https://user-images.githubusercontent.com/3236756/210110194-170c7d30-fd13-4406-8c29-3d82389455db.mp4

## Recommendation

I personally prefer the first option, even though the feedback is very subtle.

pedroferreira1:
I prefer keeping consistency with the other copy that we have in the app following the standard in the screen that we copy the address.

If we decide to go with the first option, I feel we should change in the copy address as well, to be consistent in the whole app.

trondbjoroy:
Although the feedback in the first option is really subtle I think it could work with a proper icon that users are used to for copying details in apps and websites. I feel it's a more elegant solution than we have for copying addresses today. Either way, for consistency we should use the same option everywhere.

src/components/TxDetailsModal.js
```diff
@@ -34,6 +35,14 @@ class TxDetailsModal extends Component {
     const idStr = getShortHash(tx.txId, 12);
     const explorerIcon = <Image source={icShareActive} width={24} height={24} />;
     const explorerLink = `https://explorer.hathor.network/transaction/${tx.txId}`;
+    const txIdComponent = (
+      <CopyClipboard
+        text={idStr}
+        copyText={tx.txId}
+        copiedTimeout={800}
```
pedroferreira1, line 42: Why did you change the default? It's a bit fast from the video you shared. The default is used in the wallet copy component, and I think we could keep it the same.

obiyankenobi: Seemed to slow for me, especially when the modal might be closed. In that case, we get a warning of changing the state of an unmounted component.

pedroferreira1: The warning is not related to the slowness, the event listener must be removed in the unmount of the component.

I still feel 1500ms is better but I'm ok to move with 800ms. 

However we need to remove the listeners anyway.

obiyankenobi: In this case, I imagine it comes from the `setTimeout` on `src/components/CopyClipboard.js`. There's no listener to remove on `TxDetailsModal.js`.

pedroferreira1: Yeah, you need to get the `setTimeout` reference and clear it in the `componentUnmount`

obiyankenobi: Fixed and went back to default timeout.

pedroferreira1 (approved):

andreabadesso (approved):

PR #217, opened 2023-01-02
feat: only download wallet tokens

andreabadesso:
### Acceptance Criteria
- When a transaction is received in real-time, we should only download tokens that are going to (or from) an address that belongs to the current loaded wallet

### Security Checklist

### DIFF
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -400,7 +400,7 @@ export function* handleTx(action) {
   // txAddresses should be a Set with all addresses involved in the tx
   const [tokenAddressesMap, txAddresses] = data.reduce(
     (acc, io) => {
-      if (!io.decoded) {
+      if (!io.decoded || !io.decoded.address) {
         return acc;
       }
 
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -33,6 +33,7 @@ import {
 import { eventChannel } from 'redux-saga';
 import { getUniqueId } from 'react-native-device-info';
 import { t } from 'ttag';
+import { get } from 'lodash';
 import {
   STORE,
   DEFAULT_TOKEN,
@@ -68,7 +69,6 @@ import { fetchTokenData } from './tokens';
 import { specificTypeAndPayload } from './helpers';
 import NavigationService from '../NavigationService';
 import { setKeychainPin } from '../utils';
-import { get } from 'lodash';
 
 export const WALLET_STATUS = {
   READY: 'ready',
@@ -411,7 +411,7 @@ export function* handleTx(action) {
         return acc;
       }
 
-      if (!acc[token]) {
+      if (!acc[0][token]) {
         acc[0][token] = new Set([]);
       }
 
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -425,8 +425,8 @@ export function* handleTx(action) {
   const txWalletAddresses = yield call(wallet.checkAddressesMine.bind(wallet), [...txAddresses]);
   const tokensToDownload = [];
 
-  for (const [tokenUid, addresses] of tokenAddressesMap) {
-    for (const address of addresses) {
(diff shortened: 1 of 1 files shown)


src/sagas/wallet.js
```diff
@@ -386,24 +387,71 @@ export function* handleTx(action) {
...
+      }
+
+      acc[0][token].add(address);
+      acc[1].add(address);
+
+      return acc;
+    }, [{}, new Set([])],
+  );
+
+  const txWalletAddresses = yield call(wallet.checkAddressesMine.bind(wallet), [...txAddresses]);
```
pedroferreira1, line 425: Have you checked if this API replies fast enough with 500 addresses in the list?

pedroferreira1 (approved):

src/sagas/wallet.js
```diff
@@ -386,24 +387,71 @@ export function* handleTx(action) {
...
+      }
+
+      const { token, decoded: { address } } = io;
+
+      // We are only interested in registered tokens
+      if (registeredTokens.indexOf(token) === -1) {
+        return acc;
+      }
+
+      if (!acc[token]) {
```
r4mmer, line 414:
```suggestion
      if (!acc[0][token]) {
```

src/sagas/wallet.js
```diff
@@ -386,24 +387,71 @@ export function* handleTx(action) {
...
+
+  // tokenAddressesMap should be an object { [tokenUid]: Set(addresses) }
+  // txAddresses should be a Set with all addresses involved in the tx
+  const [tokenAddressesMap, txAddresses] = data.reduce(
+    (acc, io) => {
+      if (!io.decoded) {
+        return acc;
+      }
+
+      const { token, decoded: { address } } = io;
```
r4mmer, line 407:
```suggestion
      const { token, decoded: { address } } = io;
      
      if (!address) {
        return acc;
      }
```

If the address is not defined, should we ignore the `input/output`?
I think this can happen with data outputs.

src/sagas/wallet.js
```diff
@@ -386,24 +387,71 @@ export function* handleTx(action) {
...
+  );
+
+  const txWalletAddresses = yield call(wallet.checkAddressesMine.bind(wallet), [...txAddresses]);
+  const tokensToDownload = [];
+
+  for (const [tokenUid, addresses] of Object.entries(tokenAddressesMap)) {
+    for (const [address] of addresses.entries()) {
+      // txWalletAddresses should always have the address we requested, but we should double check
+      // here using lodash just in case
+      const inWallet = get(txWalletAddresses, address, false);
```
r4mmer, line 432: If we use `wallet.isAddressMine` as a filter when creating the `tokenAddressesMap` we can ensure that the keys of `tokenAddressesMap` only have tokens with at least one address in the wallet.

r4mmer (changes requested):

src/sagas/wallet.js
andreabadesso: Fixed! Thanks

src/sagas/wallet.js
andreabadesso: I agree, but that would only work on the fullnode facade, the wallet-service facade makes a http request to the wallet-service lambda to detect if the address belongs to the wallet

src/sagas/wallet.js
andreabadesso: Fixed! Thanks

src/sagas/wallet.js
andreabadesso: I think we should. Fixed in c36989f17dcee6be4c254f94a8e6ec5844d97fe1

r4mmer (approved):

src/sagas/wallet.js
andreabadesso: Just did this test, the lambda averages on ~7ms which translates into a average ~300ms total request duration (including latency) for me, in Brazil

PR #227, opened 2023-01-10
feat: opt-in question

alexruzenhack:
### Acceptance Criteria
- Ask the user to opt-in to the push notification the first time it opens the wallet after the wallet update.
- https://www.loom.com/share/184aa165d9b24d149c9770bfb78a5aca

### Security Checklist


src/components/AskForPushNotification.js
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <ActionModal
+      title={t`Do you want to enable push notifications for this wallet?`}
+      message={t`You can always change this later in the settings menu.`}
+      button={t`Yes, enable`}
+      onAction={() => onEnablePushNotifications(props)}
+      onDismiss={() => props.dismissOptInQuestion()}
+    />
+  );
+}
```
andreabadesso, line 31: Since we are already using function components, we can use [react hooks](https://reactjs.org/docs/hooks-intro.html) here:

```javascript
import React from 'react';
import { useDispatch, useSelector } from 'react-redux';
import { pushDismissOptInQuestion } from '../actions';

(...)

function AskForPushNotification ({ navigation }) {
  const dispatch = useDispatch();
  const optInDismissed = useSelector(({ pushNotification }) => pushNotification.optInQuestionDismissed);

  if (!optInDismissed) {
    return null;
  }

  const onAction = () => {
    navigation.navigate('PushNotification');
    dispatch(pushDismissOptInQuestion());
  };

  const onDismiss = () => dispatch(pushDismissOptInQuestion());

  return (
    <ActionModal
      title={t`Do you want to enable push notifications for this wallet?`}
      message={t`You can always change this later in the settings menu.`}
      button={t`Yes, enable`}
      onAction={() => onAction()}
      onDismiss={() => onDismiss()}
    />
  );
}
```

src/reducer.js
```diff
@@ -96,6 +96,15 @@ const initialState = {
...
   pushNotification: {
+    /**
+     * optInQuestionDismissed {boolean} if user has dismissed the push notification modal
+     * this is used to show the modal only the first time user opens the app
+     * after the modal is dismissed, it will not be shown again
+     * until the user clears the app data
+     * (this is done to avoid showing the modal to users that have already
+     * enabled push notification)
+     */
+    optInQuestionDismissed: true,
```
andreabadesso, line 107: Shouldn't this be initialized with `false`?

src/components/AskForPushNotification.js
andreabadesso: We could also use `useNavigation` to avoid needing to pass the navigation prop down the component tree, but it's not available in the current version of react-navigation

andreabadesso (changes requested):

src/components/AskForPushNotification.js
alexruzenhack: I didn't know about the `useNavigation`, and I did this PR before the Settings refactoring. I can refactor this one as well.

src/reducer.js
alexruzenhack: No. It is initialized as dismissed, and in the push initialization verifies if there is a dismiss stored, if not, it forces it to appear.

alexruzenhack: Sounds good. I will work on it.

src/components/AskForPushNotification.js
alexruzenhack: Yep. Moving without `useNavigation`.

src/reducer.js
```diff
@@ -210,6 +219,10 @@ const reducer = (state = initialState, action) => {
       return onStartWalletFailed(state);
     case types.WALLET_BEST_BLOCK_UPDATE:
       return onWalletBestBlockUpdate(state, action);
+    case types.PUSH_ASK_OPT_IN_QUESTION:
+      return onPushAskOptInQuestiong(state);
```
pedroferreira1, line 223: Typo `onPushAskOptInQuestiong`

src/components/ActionModal.js
```diff
@@ -0,0 +1,74 @@
...
+
+import React from 'react';
+import {
+  StyleSheet, Text, View,
+} from 'react-native';
+import PropTypes from 'prop-types';
+import Modal from 'react-native-modal';
+import NewHathorButton from './NewHathorButton';
+
+const FeedbackModal = (props) => (
```
pedroferreira1, line 17: Why did you call this function `FeedbackModal` if the file name is another one?

src/components/ActionModal.js
```diff
@@ -0,0 +1,74 @@
...
+  <Modal
```
pedroferreira1, line 18: If the idea is to create another component similar to the `FeedbackModal`, I prefer if we reuse the previous one to create the new, so we don't need to change more than one place if we decide to improve the UI.

For instance, in this case, we would need to add a `prop` for the `FeedbackModal` that is optional and receives an element to add below the text. Is that correct?

src/components/ActionModal.js
alexruzenhack: Did it wrong.

src/components/ActionModal.js
alexruzenhack: It is not so simple. They have different semantics and different structure. They are similar, but only it.

alexruzenhack: Pedro, as we talked I refactored the modals in following commits:
- https://github.com/HathorNetwork/hathor-wallet-mobile/pull/227/commits/7c7d855c587770b9aced78e8d963700d75799746
- https://github.com/HathorNetwork/hathor-wallet-mobile/pull/227/commits/6a448d5d4ed24ffe86bfd5302886b895e5a5930b

src/components/AskForPushNotification.js
```diff
@@ -0,0 +1,26 @@
...
+export default function AskForPushNotification(props) {
+  const showOptIn = useSelector((state) => state.pushNotification.showOptInQuestion);
+  const dispatch = useDispatch();
+
+  const onEnablePushNotifications = () => {
+    props.navigation.navigate('PushNotification');
+    dispatch(pushDismissOptInQuestion());
+  };
+
+  if (!showOptIn) return null;
```
andreabadesso, line 16:
```suggestion
  if (!showOptIn) {
    return null;
  }
```

andreabadesso (approved):

src/components/HathorModal.js
```diff
@@ -0,0 +1,51 @@
...
+    paddingHorizontal: 16,
+    paddingBottom: 56,
+    paddingTop: 48,
+    height: 290,
+  },
+});
+
+HathorModal.propTypes = {
+  // Function to execute on dismissing the modal
+  onDismiss: PropTypes.func,
```
pedroferreira1, line 48: Missing `children` here

src/components/ActionModal.js
```diff
@@ -0,0 +1,50 @@
...
+      </Text>
+    </View>
+    <View style={{ width: '100%', justifyContent: 'flex-end', marginTop: 60 }}>
+      <NewHathorButton onPress={() => props.onAction()} title={props.button} />
+    </View>
+  </HathorModal>
+);
+
+ActionModal.propTypes = {
+  // Text displayed on the modal
```
pedroferreira1, line 34: Title*

src/components/ActionModal.js
```diff
@@ -0,0 +1,50 @@
...
+
+  // Function to execute on dismissing the modal
+  onDismiss: PropTypes.func,
+
+  // Text displayed on the button
+  button: PropTypes.string,
+
+  // Function to execute on pressing the button
+  onAction: PropTypes.func,
+};
```
pedroferreira1, line 48: Missing `textProps`

src/components/ActionModal.js
```diff
@@ -0,0 +1,50 @@
...
+
+ActionModal.propTypes = {
+  // Text displayed on the modal
+  title: PropTypes.string,
+
+  // Text displayed on the modal
+  text: PropTypes.oneOfType([PropTypes.string, PropTypes.element]).isRequired,
+
+  // Function to execute on dismissing the modal
+  onDismiss: PropTypes.func,
```
pedroferreira1, line 41: This is required, isn't it?

pedroferreira1 (approved):

PR #229, opened 2023-01-10
fix: reset wallet from lockscreen was sending the user to the loading…

andreabadesso:
### Acceptance Criteria
- We should fix the navigation from `PinScreen` -> `ResetWallet`, it was being stuck on the loading screen
- The wallet was not cleaning the `localStorage` when the reset button is pressed from the `PinScreen`

### Security Checklist

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -176,7 +176,13 @@ const AppStack = createStackNavigator({
   About,
   Security,
   ChangePin,
-  ResetWallet,
+  ResetWallet: {
+    screen: ResetWallet,
+    // Dismissing the screen on ResetWallet would lead to a
+    // MainScreen with broken state if started from the PinScreen,
+    // so just disable the swipeDown animation
+    navigationOptions: disableSwipeDown,
+  },
   PaymentRequestDetail,
   RegisterToken: RegisterTokenStack,
   ChangeToken,
--- a/src/App.js
+++ b/src/App.js
@@ -344,7 +344,14 @@ class _AppStackWrapper extends React.Component {
         screen = <RecoverPin navigation={this.props.navigation} />;
       }
 
-      if (this.props.walletStartState === WALLET_STATUS.LOADING || this.props.isScreenLocked) {
+      const currentRouteIdx = this.props.navigation.state.index;
+      const currentRoute = this.props.navigation.state.routes[currentRouteIdx];
+
+      // We need to hide the loading screen if the wallet is being reset, otherwise
+      // the loading screen will be on top of the reset screen.
+      if (currentRoute.routeName !== 'ResetWallet'
+          && (this.props.walletStartState === WALLET_STATUS.LOADING
+              || this.props.isScreenLocked)) {
         return (
           <View style={this.style.auxView}>
             { screen }
--- a/src/App.js
+++ b/src/App.js
@@ -344,14 +344,9 @@ class _AppStackWrapper extends React.Component {
         screen = <RecoverPin navigation={this.props.navigation} />;
       }
 
-      const currentRouteIdx = this.props.navigation.state.index;
-      const currentRoute = this.props.navigation.state.routes[currentRouteIdx];
-
       // We need to hide the loading screen if the wallet is being reset, otherwise
       // the loading screen will be on top of the reset screen.
-      if (currentRoute.routeName !== 'ResetWallet'
-          && (this.props.walletStartState === WALLET_STATUS.LOADING
-              || this.props.isScreenLocked)) {
+      if (this.props.walletStartState === WALLET_STATUS.LOADING || this.props.isScreenLocked) {
         return (
           <View style={this.style.auxView}>
             { screen }
--- a/src/App.js
+++ b/src/App.js
@@ -344,8 +344,6 @@ class _AppStackWrapper extends React.Component {
         screen = <RecoverPin navigation={this.props.navigation} />;
       }
 
-      // We need to hide the loading screen if the wallet is being reset, otherwise
-      // the loading screen will be on top of the reset screen.
       if (this.props.walletStartState === WALLET_STATUS.LOADING || this.props.isScreenLocked) {
         return (
           <View style={this.style.auxView}>
--- a/src/actions.js
+++ b/src/actions.js
@@ -478,11 +478,6 @@ export const reloadWalletRequested = () => ({
   type: types.RELOAD_WALLET_REQUESTED,
 });
 
-export const startWalletLock = (payload) => ({
-  type: types.ON_START_WALLET_LOCK,
-  payload,
-});
-
 export const startWalletRequested = (payload) => ({
   type: types.START_WALLET_REQUESTED,
   payload,
--- a/src/actions.js
+++ b/src/actions.js
@@ -177,15 +177,6 @@ export const updateHeight = (height, htrBalance) => (
   { type: types.UPDATE_HEIGHT, payload: { height, htrBalance } }
 );
 
-/**
- * words {String} wallet words
- * pin {String} Pin chosen by user
- */
-export const setInitWallet = (words, pin) => (
-  { type: types.SET_INIT_WALLET, payload: { words, pin } }
-);
-
-export const clearInitWallet = () => ({ type: types.SET_INIT_WALLET, payload: null });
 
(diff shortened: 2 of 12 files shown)


src/actions.js
```diff
@@ -478,6 +478,11 @@ export const reloadWalletRequested = () => ({
   type: types.RELOAD_WALLET_REQUESTED,
 });
 
+export const startWalletLock = (payload) => ({
```
pedroferreira1, line 481: We already have a `onStartWalletLock` without payload, isn't it confusing to create this method here? Maybe changing the name is better. What's the difference between them?

And the reducer doesn't do anything with this payload.

src/screens/PinScreen.js
```diff
@@ -195,9 +199,17 @@ class PinScreen extends React.Component {
...
+      onBackPress: () => {
+        this.props.startWalletLock();
+        this.backFromReset();
+      },
+    });
     // make sure we won't show loadHistory screen
     this.props.setLoadHistoryStatus(false, false);
+    // Start wallet might be set to LOADING, we also need to
+    // set it to READY so the loading screen will hide
+    this.props.startWalletSuccess();
```
pedroferreira1, line 212: You set the wallet as ready even before the unlock only to show the reset wallet?

src/sagas/wallet.js
```diff
@@ -596,11 +596,29 @@ export function* onWalletReloadData() {
...
+    // hathorLib.wallet.cleanWallet
+    wallet.stop({ cleanStorage: true });
+
+    return;
+  }
+
+  // Wallet was not initialized yet, this might happen if resetWallet
+  // is called from the PinScreen. There is no event listeners to cleanup
+  // so we can call the cleanWallet method directly.
+  walletUtil.cleanWallet();
```
pedroferreira1, line 613: This method unsubscribe addresses from the ws and other actions from the connection. Doesn't it throw errors when you call it without a `wallet` object? We shouldn't have a connection yet.

andreabadesso: That's a valid point, currently, it is not throwing an error because it's defaulting to the old `WebSocketHandler`, which is an `EventEmitter`. I will refactor this to only clean the storage

src/actions.js
andreabadesso: Oops.... I shouldn't have created this action, removed it in c1c790e1867ea76506df7d8e837d9d17a7fcbdf8
 and using the `onStartWalletLock`

src/sagas/wallet.js
andreabadesso: Done in 52289657b65b452f4b1f999bc11a7057b048b928

src/screens/PinScreen.js
andreabadesso: 😳

Refactored with a better solution:
3beb1d5ebe85f74cdbb28a05907623991cceeaa2

src/screens/PinScreen.js
```diff
@@ -44,6 +46,8 @@ const mapDispatchToProps = (dispatch) => ({
   setLoadHistoryStatus: (active, error) => dispatch(setLoadHistoryStatus(active, error)),
   setInitWallet: (words, pin) => dispatch(setInitWallet(words, pin)),
   setTempPin: (pin) => dispatch(setTempPin(pin)),
+  startWalletSuccess: () => dispatch(startWalletSuccess()),
```
pedroferreira1, line 49: We don't need this anymore

src/App.js
```diff
@@ -338,7 +344,14 @@ class _AppStackWrapper extends React.Component {
         screen = <RecoverPin navigation={this.props.navigation} />;
       }
 
-      if (this.props.walletStartState === WALLET_STATUS.LOADING || this.props.isScreenLocked) {
+      const currentRouteIdx = this.props.navigation.state.index;
+      const currentRoute = this.props.navigation.state.routes[currentRouteIdx];
+
+      // We need to hide the loading screen if the wallet is being reset, otherwise
+      // the loading screen will be on top of the reset screen.
+      if (currentRoute.routeName !== 'ResetWallet'
```
pedroferreira1, line 352: I don't understand the need for this. The wallet is never loading because we move from the locked screen to the reset screen, so why it enters this `if (this.props.walletStartState === WALLET_STATUS.LOADING ..` in the first place?

src/screens/PinScreen.js
andreabadesso: You are right, thanks

Fixed in e9c92f356c017d00d4923fef3b442439d2505433

src/App.js
andreabadesso: We need this because, currently, the wallet defaults to `LOADING`, so when the `PinScreen` hides, it automagically navigates to `LoadHistoryScreen` which in turn will dispatch `startWalletRequested`

pedroferreira1 (approved):

r4mmer (approved):

src/reducer.js
```diff
@@ -565,6 +543,11 @@ export const onTokenFetchHistoryRequested = (state, action) => {
   };
 };
 
+export const onStartWalletNotStarted = (state) => ({
```
pedroferreira1, line 546: Why not `setWalletNotStarted`?

src/screens/ChoosePinScreen.js
```diff
@@ -18,15 +18,18 @@ import { t } from 'ttag';
...
+import { startWalletRequested, unlockScreen } from '../actions';
 import { PIN_SIZE } from '../constants';
 
 import baseStyle from '../styles/init';
 
 
 const mapDispatchToProps = (dispatch) => ({
   unlockScreen: () => dispatch(unlockScreen()),
-  setInitWallet: (words, pin) => dispatch(setInitWallet(words, pin)),
+  startWalletRequested: (words, pin) => dispatch(startWalletRequested({
```
pedroferreira1, line 29: This reducer is not doing anything with the payload, why do you send it?

src/reducer.js
```diff
@@ -565,6 +543,11 @@ export const onTokenFetchHistoryRequested = (state, action) => {
...
+  ...state,
+  walletStartState: WALLET_STATUS.NOT_STARTED,
+});
+
```
pedroferreira1, line 550: Maybe we should change the `onStartWalletLock` method to also change state to `NOT_STARTED`, what do you think?

src/screens/ChoosePinScreen.js
andreabadesso: The reducer is not doing anything with it, but the `startWallet` saga is

src/screens/PinScreen.js
```diff
@@ -210,6 +223,8 @@ class PinScreen extends React.Component {
     this.props.setLoadHistoryStatus(this.props.loadHistoryActive, false);
     // show lock screen again
     this.props.lockScreen();
+    // set the wallet state as NOT_STARTED so the LoadingScreen will hide.
+    this.props.startWalletNotStarted();
```
pedroferreira1, line 227: Why is it needed? If you don't unlock the wallet, it shouldn't change from state `NOT_STARTED`

src/App.js
```diff
@@ -338,6 +344,8 @@ class _AppStackWrapper extends React.Component {
         screen = <RecoverPin navigation={this.props.navigation} />;
       }
 
+      // We need to hide the loading screen if the wallet is being reset, otherwise
```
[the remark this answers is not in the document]
andreabadesso: Yes it is, removed in c8591ad0952eb6eb505486edaa4900f48b82c123

src/reducer.js
andreabadesso: I agree, refactored in 9fcb649c73fc0e5d62bfe787082f0dd8e98818cf

src/screens/PinScreen.js
andreabadesso: You are right, we don't need to change it to `NOT_STARTED` anymore as the flow has been refactored

Justed it and the state does not change

Refactored in d89a22dba5b879f471b052db49fe9c6509cfb0ea

src/reducer.js
andreabadesso: Actually, we don't need this action and reducer anymore

pedroferreira1 (approved):

PR #236, opened 2023-01-17
feat: reset push notification

alexruzenhack:
### Acceptance Criteria
- Reset push notification store and state after wallet reset

### Security Checklist


src/sagas/pushNotification.js
```diff
@@ -444,6 +445,28 @@ export function* dismissOptInQuestion() {
...
+  yield call(cleanToken);
+  // Clean the store
+  yield STORE.removeItem(pushNotificationKey.enabledAt);
+  yield STORE.removeItem(pushNotificationKey.settings);
+  yield STORE.removeItem(pushNotificationKey.hasBeenEnabled);
+  yield STORE.removeItem(pushNotificationKey.optInDismissed);
+  yield STORE.removeItem(pushNotificationKey.deviceId);
+  // Reset the state
+  yield put(pushReset());
+  yield fork(onAppInitialization);
```
pedroferreira1, line 467: Why do you need this in the end?

src/sagas/pushNotification.js
```diff
@@ -444,6 +445,28 @@ export function* dismissOptInQuestion() {
...
+}
```
pedroferreira1, line 468: Don't you need to also remove the push notifications settings from the wallet service?

src/reducer.js
```diff
@@ -802,6 +804,20 @@ export const onPushApiFailed = (state) => ({
   },
 });
 
+export const onPushReset = (state) => ({
+  ...state,
+  pushNotification: {
+    ...state.pushNotification,
```
pedroferreira1, line 810: Why do you need this? You are resetting all the keys, aren't you?

src/sagas/pushNotification.js
```diff
@@ -444,6 +445,28 @@ export function* dismissOptInQuestion() {
...
+ * This function is responsible for reset the push notification.
+ */
+export function* resetPushNotification() {
+  // Unregister the device from FCM
+  yield call(cleanToken);
+  // Clean the store
+  yield STORE.removeItem(pushNotificationKey.enabledAt);
+  yield STORE.removeItem(pushNotificationKey.settings);
+  yield STORE.removeItem(pushNotificationKey.hasBeenEnabled);
+  yield STORE.removeItem(pushNotificationKey.optInDismissed);
```
pedroferreira1, line 463: Why are you removing this? This should continue to be saved because you need to show the modal only the first time the user opens his wallet with the new feature, not every new initialized wallet.

src/reducer.js
andreabadesso: Maybe we should do something like 

```
pushNotification: initialState.pushNotification,
```

src/sagas/pushNotification.js
andreabadesso: Also confused about what was the idea here

`onAppInitialization` calls `take` on `onStartWalletSuccess`, and I guess that at this point, it will not get called again, right?

Also, if you want this to run as a process in parallel, you should use `spawn` instead of `fork` as `fork` will spawn an attached process

src/sagas/pushNotification.js
alexruzenhack: No, it is not necessary because we have a process to clean stale registers. And if the user re-enables the push notification this register will be updated.

And I did this way to avoid bothering the user with another pin prompt in the reset process.

src/sagas/pushNotification.js
alexruzenhack: I have changed this process in the feature toggle PR [a94016f](https://github.com/HathorNetwork/hathor-wallet-mobile/pull/240/commits/a94016f451a41be028b079b572fc5ce6e5c5e411#diff-b2f41e02468f99f5c9c39db9c11e7c79fbde4091cf522468d84af47845be7adaR247)

src/sagas/pushNotification.js
alexruzenhack: We should not reset the push notification feature after a wallet reset?

alexruzenhack: This is required especially because we reset the device ID as well.

andreabadesso (changes requested):

src/sagas/pushNotification.js
```diff
@@ -412,11 +413,34 @@ export function* dismissOptInQuestion() {
...
+  await messaging().unregisterDeviceForRemoteMessages();
+  await messaging().deleteToken();
+};
+
+/**
+ * This function is responsible for reset the push notification.
+ */
+export function* resetPushNotification() {
+  // Unregister the device from FCM
+  yield call(cleanToken);
```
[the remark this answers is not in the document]
alexruzenhack: Can I solve it on the next PR? It is where I implemented all the error-catching with the new `onExceptionCaptured`.

src/sagas/pushNotification.js
pedroferreira1: Isn't the device ID the same if the user resets the wallet? The idea for the opt in modal is like a message to the user that we have a new feature available. It shouldn't appear every time the user loads a wallet, otherwise it's not good in my opinion.

alexruzenhack: Solved it by removing the optInDismissed from reset.

src/sagas/pushNotification.js
pedroferreira1: So this is wrong? Why no change here? If it does not work as expected we shouldn't merge.

alexruzenhack: I realized it only after I was working on the feature toggle. As the PRs are chained, when I merge the feature toggle PR here, it will be resolved. That's why I didn't solve it here.

alexruzenhack: Removed.

src/sagas/pushNotification.js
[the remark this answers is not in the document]
alexruzenhack: It is already solved in the next PR of the chain.

andreabadesso (approved):

pedroferreira1 (approved):

PR #241, opened 2023-01-23
fix: wallet not reloading data after a reconnect and wallet reset after an error

andreabadesso:
## Summary

After the refactor done in https://github.com/HathorNetwork/hathor-wallet-mobile/pull/232, the wallet stopped reloading after a websocket reconnection.

The initial problem was that we were not setting the `walletStartedState` to `LOADING`, which made it reload the wallet but never display the loading screen.

But not only that, since we refactored the loading flow, the reload was still not happening. Upon investigation, I found out that the old mechanism for reloading the wallet was only working because the `LoadHistoryScreen` was dispatching the `startWalletRequested` action on the `componentDidUpdate`, as we can see [here](https://github.com/HathorNetwork/hathor-wallet-mobile/pull/232/files#diff-6b628a954cc3611565a6b19e839717bb5431913b6d02c38206de52df4b355fbbL55).

### Acceptance Criteria
- We should reload the wallet data after a reconnection
- We should refresh the wallet addresses after a reconnection
- We should stop resetting the wallet if an error occurs
- We should display the correct wallet-service connected server while on the wallet-service facade

### Security Checklist

### DIFF
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -609,6 +609,9 @@ export function* onWalletReloadData() {
     // We also need to refresh the addresses since we are not starting the wallet again
     yield call(wallet.getNewAddresses.bind(wallet));
 
+    // Then dispatch the refreshSharedAddress so our redux store is updated
+    yield put(walletRefreshSharedAddress());
+
     // Finally, set the wallet to READY by dispatching startWalletSuccess
     yield put(startWalletSuccess());
   } catch (e) {
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -486,6 +486,9 @@ export function* handleTx(action) {
   if (action.type === 'WALLET_NEW_TX') {
     yield put(newTx(tx));
   }
+
+  // We should sync the last shared address on our redux store with the facade's internal state
+  yield put(walletRefreshSharedAddress());
 }
 
 export function* setupWalletListeners(wallet) {
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -659,10 +659,7 @@ export function* refreshSharedAddress() {
 
   const { address, index } = wallet.getCurrentAddress();
 
-  yield put(sharedAddressUpdate({
-    lastSharedAddress: address,
-    lastSharedIndex: index,
-  }));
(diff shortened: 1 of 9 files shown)


r4mmer (approved):

src/sagas/wallet.js
```diff
@@ -568,16 +567,19 @@ export function* onWalletConnStateUpdate({ payload }) {
...
 
-  // Since we close the channel after a walletReady event is received,
-  // we must fork this saga again so we setup listeners again.
-  yield fork(listenForWalletReady, wallet);
+  // If we are using the wallet-service, we don't need to wait until the addresses
+  // are reloaded since they are stored on the wallet-service itself.
+  if (!useWalletService) {
+    // Since we close the channel after a walletReady event is received,
+    // we must fork this saga again so we setup listeners again.
+    yield fork(listenForWalletReady, wallet);
```
pedroferreira1, line 578: I understand we don't need to wait for the wallet ready to continue if it's wallet service but don't we need this listener anyway for other things, e.g., error handling when the load fails, ws connection lost (this last one I feel it's not affected)?

andreabadesso:
@pedroferreira1 and @r4mmer 

The addresses were not being refreshed after a reload on the wallet-service because we were only refreshing the token balances and history, this was fixed

I had also to change the mechanism for displaying the current address on the `ReceiveMyAddress` component as it was not being updated if the user reconnected while the screen was opened

I also fixed a bug where the wallet would reset if some error was thrown and the user didn't click to try again

src/sagas/wallet.js
```diff
@@ -596,6 +605,14 @@ export function* onWalletReloadData() {
 
       yield put(tokenInvalidateHistory(tokenUid));
     }
+
+    // We also need to refresh the addresses since we are not starting the wallet again
+    yield call(wallet.getNewAddresses.bind(wallet));
```
pedroferreira1, line 610: We don't have this method in the old facade.

andreabadesso: Fixed in 4250c651855426d27d8ba3e8c155a2462a4a95b2

Thanks

pedroferreira1 (approved):

PR #249, opened 2023-02-02
feat: using fake facade instead of wallet-service facade

andreabadesso:
### Acceptance Criteria

### Security Checklist


.detoxrc.js
```diff
@@ -0,0 +1,83 @@
...
+    },
+    jest: {
+      setupTimeout: 120000
+    }
+  },
+  apps: {
+    'ios.debug': {
+      type: 'ios.app',
+      binaryPath: 'ios/DerivedData/HathorMobile/Build/Products/Debug-iphonesimulator/HathorMobile.app',
+      build: 'xcodebuild -workspace ios/HathorMobile.xcworkspace -configuration Debug -scheme HathorMobile -destination id=0203B78D-AFD4-4964-8E33-16A3D078D9DC'
```
andreabadesso, line 16: This is the simulator UUID, we should find a better way to build for a generic ios device

PR #252, opened 2023-02-10
refactor: load from xpriv

andreabadesso:
## Summary

This improves the wallet load time by starting the wallet from the stored xpriv instead of generating them again during the loading step

During the investigation of why the wallet was taking up to 11s to load, I ran a benchmark on the methods the lib calls during initialization and found two bottlenecks:

| Method | Time (seconds) |
| --- | --- |
| generateCreateWalletAuthData | 5.401s |
| executeGenerateWallet | 2.453s |

Something interesting to notice is that when the debugger is activated, those methods are up to 40× faster and that's why it was so hard to figure out:

| Method | Time (seconds) |
| --- | --- |
| generateCreateWalletAuthData | 0.357s |
| executeGenerateWallet | 0.145s |

Apparently, when the debugger is activated, it changes the default javascript runtime to v8 (the one from the computer that is running the debugger). `bitcore-lib` is much faster on v8

I tried activating [Hermes](https://reactnative.dev/docs/hermes) on the build, but it didn't improve by a lot.

### Acceptance Criteria
- We should load the from xpriv instead of using the seed
  - The idea here is to avoid calling `executeGenerateWallet` when opening an already started wallet since it's taking a long time to run on mobile devices (because of bitcorelib)
- We should make sure that users migrating from older versions of the wallet won't have problems loading their wallets

### Security Checklist

### DIFF
--- a/src/screens/LoadWalletErrorScreen.js
+++ b/src/screens/LoadWalletErrorScreen.js
@@ -10,7 +10,7 @@ import {
   Text,
   View,
 } from 'react-native';
-import { useDispatch, useSelector } from 'react-redux';
+import { useDispatch } from 'react-redux';
 import { onStartWalletLock, lockScreen } from '../actions';
 import SimpleButton from '../components/SimpleButton';
 
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -79,7 +79,11 @@ export const WALLET_STATUS = {
 };
 
 export function* startWallet(action) {
-  const { words, pin } = action.payload;
+  const {
+    words,
+    pin,
+    fromXpriv,
+  } = action.payload;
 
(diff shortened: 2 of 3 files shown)


src/screens/LoadWalletErrorScreen.js
```diff
@@ -10,7 +10,7 @@ import {
   Text,
   View,
 } from 'react-native';
-import { useDispatch, useSelector } from 'react-redux';
+import { useDispatch } from 'react-redux';
```
andreabadesso, line 13: `useSelector` is not being used on this file so I removed it to avoid a warning on the linter, not related to this PR

r4mmer (approved):

msbrogli (approved):

PR #260, opened 2023-03-07
fix: push notification initialization for iOS

alexruzenhack:
### Acceptance Criteria
- add lacking iOS pod dependencies
- disable firebase auto initialization
- add guard in the firebase background listener initialization

### Security Checklist

### DIFF
--- a/ios/Podfile
+++ b/ios/Podfile
@@ -34,4 +34,9 @@ target 'HathorMobile' do
     react_native_post_install(installer)
     __apply_Xcode_12_5_M1_post_install_workaround(installer)
   end
+
+  pod 'Firebase', :modular_headers => true
+  pod 'FirebaseCore', :modular_headers => true
+  pod 'GoogleUtilities', :modular_headers => true
+  $RNFirebaseAsStaticFramework = true
 end
--- a/ios/Podfile
+++ b/ios/Podfile
@@ -35,7 +35,6 @@ target 'HathorMobile' do
     __apply_Xcode_12_5_M1_post_install_workaround(installer)
   end
 
-  pod 'Firebase', :modular_headers => true
   pod 'FirebaseCore', :modular_headers => true
   pod 'GoogleUtilities', :modular_headers => true
   $RNFirebaseAsStaticFramework = true
--- a/src/workers/backgroundListeners.js
+++ b/src/workers/backgroundListeners.js
@@ -28,7 +28,7 @@ const isRegisteredOnFirebase = () => {
     }
   } catch (error) {
     console.error('Error confirming the device is registered on firebase while'
-      + ' loading the background message listener. May auto initialization of firebase'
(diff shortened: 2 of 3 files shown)


ios/Podfile
```diff
@@ -34,4 +34,9 @@ target 'HathorMobile' do
     react_native_post_install(installer)
     __apply_Xcode_12_5_M1_post_install_workaround(installer)
   end
+
+  pod 'Firebase', :modular_headers => true
```
pedroferreira1, line 38: Here https://firebase.google.com/docs/ios/setup#available-pods we have this text:

> For apps that use CocoaPods, the Firebase pod is deprecated in v9.0 and higher. Instead, you need to reference product pods directly in your Podfile (for example, FirebaseCore instead of Firebase/Core and FirebaseFirestore instead of Firebase/Firestore).

Do you need both `Firebase` and `FirebaseCore`?

src/workers/backgroundListeners.js
```diff
@@ -15,10 +15,36 @@ import {
...
+const isRegisteredOnFirebase = () => {
+  try {
+    // Make sure deviceId is registered on the FCM
+    if (!messaging().isDeviceRegisteredForRemoteMessages) {
+      console.debug('The device is not registered on the firebase yet.');
+      return false;
+    }
+  } catch (error) {
+    console.error('Error confirming the device is registered on firebase while'
+      + ' loading the background message listener. May auto initialization of firebase'
```
pedroferreira1, line 31: Typo 'May' -> 'Maybe'?

ios/Podfile
andreabadesso: I was having issues installing the project on arm:

```bash
LoadError - dlopen(/Library/Ruby/Gems/2.6.0/gems/ffi-1.15.5/lib/ffi_c.bundle, 0x0009): tried: '/Library/Ruby/Gems/2.6.0/gems/ffi-1.15.5/lib/ffi_c.bundle' (mach-o file, but is an incompatible architecture (have 'x86_64', need 'arm64'))
```

To make it work, I had to install using `x86_64`:

```bash
arch -x86_64 pod install
```

But just checked, and removing this dependency (`Firebase`) solves the issue

alexruzenhack: Done here [e8970](https://github.com/HathorNetwork/hathor-wallet-mobile/pull/260/commits/e8970832089160e3eb1263c7d3aa35d1b0181e6d)

src/workers/backgroundListeners.js
alexruzenhack: Done here [e8970](https://github.com/HathorNetwork/hathor-wallet-mobile/pull/260/commits/e8970832089160e3eb1263c7d3aa35d1b0181e6d)

andreabadesso (approved):

pedroferreira1 (approved):

PR #261, opened 2023-03-08
fix: tx details modal over lock screen

alexruzenhack:
### Acceptance Criteria
- Avoid transaction details modal (balance modal) to appear over lock screen. With this fix the modal will only appear after unlock the screen.

Regarding the previous approach to render the reset page from the pin screen:

> The previous approach unlocks the wallet and navigates to ResetWallet
> screen with a back button pointing to StartWalletLockScreen,
> this interfering with the stack navigation.

### Security Checklist

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -237,6 +237,7 @@ const AppStack = createStackNavigator({
 const mapStateToProps = (state) => ({
   loadHistory: state.loadHistoryStatus.active,
   isScreenLocked: state.lockScreen,
+  isResetOnScreenLocked: state.resetOnLockScreen,
   isRecoveringPin: state.recoveringPin,
   walletStartState: state.walletStartState,
 });
@@ -400,7 +401,24 @@ class _AppStackWrapper extends React.Component {
 
       if (!this.props.isRecoveringPin) {
         if (this.props.isScreenLocked) {
-          screen = <PinScreen isLockScreen navigation={this.props.navigation} />;
+          /**
+           * NOTE:
+           * This approach shows the ResetWallet screen as an auxiliar view,
+           * replacing the PinScreen and setting the back button to drop the view,
+           * letting the PinScreen be re-rendered.
+           *
+           * This approach also keeps the navigation stack unchanged,
+           * therefore increasing the convinience for the user.
+           *
+           * The previous approach unlocks the wallet and navigates to ResetWallet
+           * screen with a back button pointing to StartWalletLockScreen,
+           * this interfering in the stack navigation.
+           */
+          if (this.props.isResetOnScreenLocked) {
+            screen = <ResetWallet navigation={this.props.navigation} />;
+          } else {
+            screen = <PinScreen isLockScreen navigation={this.props.navigation} />;
+          }
         } else {
           screen = <LoadHistoryScreen />;
         }
--- a/src/App.js
+++ b/src/App.js
@@ -409,10 +409,6 @@ class _AppStackWrapper extends React.Component {
            *
            * This approach also keeps the navigation stack unchanged,
            * therefore increasing the convinience for the user.
-           *
-           * The previous approach unlocks the wallet and navigates to ResetWallet
-           * screen with a back button pointing to StartWalletLockScreen,
-           * this interfering in the stack navigation.
            */
           if (this.props.isResetOnScreenLocked) {
             screen = <ResetWallet navigation={this.props.navigation} />;
--- a/src/actions.js
+++ b/src/actions.js
@@ -190,8 +190,23 @@ export const setUniqueDeviceId = (uniqueId) => ({
   payload: uniqueId,
 });
 
+/**
+ * Action to set the screen unlocked.
+ * @returns action
+ */
 export const unlockScreen = () => ({ type: types.SET_LOCK_SCREEN, payload: false });
 
+/**
+ * Check if the action is about to set screen to unlocked state.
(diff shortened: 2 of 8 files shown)


src/sagas/pushNotification.js
```diff
@@ -511,6 +512,11 @@ export const getTxDetails = async (wallet, txId) => {
  */
 export function* loadTxDetails(action) {
   const { txId } = action.payload;
+  const isLocked = select((state) => state.lockScreen);
+  if (isLocked) {
+    yield take(isUnlockScreen);
```
andreabadesso, line 517: This will be stuck here until `isUnlockScreen` action is dispatched, right?

I guess that if the user resets the screen from the `LockScreen`, starts a new wallet, and then unlocks it, it will display the tx details screen, right?

We can handle that by using race, something like:

```javascript
const { resetWallet, unlockWallet } = yield race({
  unlockWallet: isUnlockScreen,
  resetWallet: <reset_wallet_action_here>,
});

if (resetWallet) {
  return;
}
```

alexruzenhack: Resolvido em [591e6e3](https://github.com/HathorNetwork/hathor-wallet-mobile/pull/261/commits/591e6e31dea4b71c68501a173e4430b3f1cea2ae) [bcf38ab](https://github.com/HathorNetwork/hathor-wallet-mobile/pull/261/commits/bcf38abad7e148d55f126ff97571a3cbcc4ccebd)

src/App.js
```diff
@@ -400,7 +401,24 @@ class _AppStackWrapper extends React.Component {
...
+           * This approach shows the ResetWallet screen as an auxiliar view,
+           * replacing the PinScreen and setting the back button to drop the view,
+           * letting the PinScreen be re-rendered.
+           *
+           * This approach also keeps the navigation stack unchanged,
+           * therefore increasing the convinience for the user.
+           *
+           * The previous approach unlocks the wallet and navigates to ResetWallet
+           * screen with a back button pointing to StartWalletLockScreen,
+           * this interfering in the stack navigation.
```
andreabadesso, line 415: I don't think we should have this part of the comment here, just add it as a comment on the PR, future developers shouldn't care about the "previous approach"

src/actions.js
```diff
@@ -190,10 +191,37 @@ export const setUniqueDeviceId = (uniqueId) => ({
...
+ * @returns action
+ */
 export const unlockScreen = () => ({ type: types.SET_LOCK_SCREEN, payload: false });
 
+/**
+ * Check if the action is about to set screen to unlocked state.
+ * @param {{ type: string, payload: boolean }} action
+ * @returns {boolean} true if unlocked and false otherwise.
+ */
+export const isUnlockScreen = (action) => action.type === types.SET_LOCK_SCREEN && action.payload === false;
```
andreabadesso, line 205: This is not an action, it's just a helper method, I think we should move it to `sagas/pushNotification.js` or to `sagas/helpers.js`

andreabadesso: Also, just a suggestion, maybe rename it to `willUnlock`, what do you think?

andreabadesso (changes requested):

src/actions.js
alexruzenhack: I agree with the helper, but not the rename. `willUnlock` denotes some action in the future, but the action was already triggered, therefore it is near the present time, or strictly already in the past.

src/screens/ResetWallet.js
```diff
@@ -53,8 +61,16 @@ class ResetWallet extends React.Component {
...
+    if (this.props.isScreenLocked) {
+      this.onBackPress = this.props.dropResetOnLockScreen;
+    } else {
+      this.onBackPress = this.props.navigation.getParam('onBackPress', this.props.navigation.goBack);
+    }
+
+    this.hideBackButton = false;
+    if (this.props.navigation) {
+      this.hideBackButton = this.props.navigation.getParam('hideBackButton', false);
+    }
```
glevco, line 73: Another suggestion, using optional chaining:

```suggestion
    this.hideBackButton = this.props.navigation?.getParam('hideBackButton', false) ?? false
```

glevco (approved):

src/App.js
alexruzenhack: Resolvido em [716ecaf](https://github.com/HathorNetwork/hathor-wallet-mobile/pull/261/commits/716ecaf1c32fe99dc9375120061ee873019e20e1)

src/screens/ResetWallet.js
```diff
@@ -53,8 +61,16 @@ class ResetWallet extends React.Component {
...
-    this.onBackPress = this.props.navigation.getParam('onBackPress', this.props.navigation.goBack);
-    this.hideBackButton = this.props.navigation.getParam('hideBackButton', false);
+    if (this.props.isScreenLocked) {
+      this.onBackPress = this.props.dropResetOnLockScreen;
+    } else {
+      this.onBackPress = this.props.navigation.getParam('onBackPress', this.props.navigation.goBack);
+    }
+
+    this.hideBackButton = false;
+    if (this.props.navigation) {
```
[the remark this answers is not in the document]
alexruzenhack: Yes, it can, but there is no other way than let the navigation break if it is not passed to the scope. We can't enforce navigation there, nor use the `useNavigation` hook.

andreabadesso (approved):

PR #274, opened 2023-04-27
feat: new storage

r4mmer:
### Acceptance Criteria
- Refactor to adjust for changes on the new version of the lib, i.e. new storage scheme.

### Security Checklist

### DIFF
--- a/src/sagas/errorHandler.js
+++ b/src/sagas/errorHandler.js
@@ -16,7 +16,7 @@ import {
   race,
   take,
 } from 'redux-saga/effects';
-import { SENTRY_DSN, STORE } from '../constants';
+import { SENTRY_DSN } from '../constants';
 import { showErrorModal } from '../actions';
 
 /**
--- a/src/screens/PinScreen.js
+++ b/src/screens/PinScreen.js
@@ -157,7 +157,6 @@ class PinScreen extends React.Component {
       // This will only delete keys with the wallet prefix, so we don't delete
       // the biometry keys and new data.
       await STORE.clearItems(true);
-
     }
     // We have finished the migration so we can set the storage version to the most recent one.
     STORE.updateStorageVersion();
--- a/package-lock.json
+++ b/package-lock.json
@@ -1351,18 +1351,19 @@
       }
     },
     "@hathor/wallet-lib": {
-      "version": "0.46.2",
-      "resolved": "https://registry.npmjs.org/@hathor/wallet-lib/-/wallet-lib-0.46.2.tgz",
-      "integrity": "sha512-DZsbh3Oq7548HuKcjXA0qvjz/Cp784RWYKEu0Bqc3W+EXHtSECECIOB8McY7t2ZCDEXXr3+5UM3CtyYBoVtSXw==",
+      "version": "1.0.0-rc3",
+      "resolved": "https://registry.npmjs.org/@hathor/wallet-lib/-/wallet-lib-1.0.0-rc3.tgz",
+      "integrity": "sha512-OzYpvZgsUvZcTV4s6qCxjQ/nV3ETML5crxdUc95r6Q+AptchHZzWD1/AjJBJZ2vzGuTKbqQJRy8Skbh6NU53yQ==",
       "requires": {
         "axios": "^0.21.4",
         "bitcore-lib": "^8.25.10",
         "bitcore-mnemonic": "^8.25.10",
         "crypto-js": "^3.1.9-1",
         "isomorphic-ws": "^4.0.1",
-        "lodash": "^4.17.11",
+        "level": "^8.0.0",
+        "lodash": "^4.17.21",
         "long": "^4.0.0",
-        "ws": "^7.2.1"
+        "ws": "^7.5.9"
       },
       "dependencies": {
         "crypto-js": {


src/utils.js
```diff
@@ -310,22 +299,47 @@ export const setKeychainPin = (pin) => {
...
+    let accessData = STORE.getAccessData();
+    accessData = hathorLib.walletUtils.changeEncryptionPin(accessData, oldPin, newPin);
+    accessData = hathorLib.walletUtils.changeEncryptionPassword(accessData, oldPin, newPin);
+    STORE.saveAccessData(accessData);
+  } catch (err) {
+    return false;
+  }
+
+  try {
+    // Will throw if the access data does not have the seed.
```
alexruzenhack, line 332: Should we add `@throws` property in the docstring describing the exception and its context?

alexruzenhack (commented):
I still need to run a QA.

src/utils.js
r4mmer: This error is catched

src/store.js
```diff
@@ -29,10 +36,145 @@ class AsyncStorageStore {
...
+      // decrypt words with the old method so we can load the wallet
+      // when first starting after a version update
+      // This should only happen when migrating from wallet-lib v0.* to v1.*
+      const decryptedWords = CryptoJS.AES.decrypt(accessData.words, pin);
+      return decryptedWords.toString(CryptoJS.enc.Utf8);
+    }
+
+    // The access data was already loaded with lib version 1.0.0
+    // To decrypt the words we can use the decryptData util
+    // This may throw an exception but this is expected
```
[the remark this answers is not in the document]
r4mmer: Done

src/store.js
```diff
@@ -29,10 +36,145 @@ class AsyncStorageStore {
 
   setItem(key, value) {
     this.hathorMemoryStorage[key] = value;
-    if (key !== 'wallet:data') {
-      // No need to save wallet data (transactions history) on the async storage
-      AsyncStorage.setItem(key, JSON.stringify(value));
+    AsyncStorage.setItem(key, JSON.stringify(value));
+  }
+
+  async clearItems(onlyWalletKeys = false) {
```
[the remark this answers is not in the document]
r4mmer: Added docstring

alexruzenhack (approved):

src/sagas/wallet.js
```diff
@@ -120,9 +126,12 @@ export function* startWallet(action) {
...
 
-  // We don't want to clean access data since as if something goes
-  // wrong here, the stored words would be lost forever.
-  walletUtil.cleanLoadedData({ cleanAccessData: false });
+  // clean storage and metadata before starting the wallet
+  // this should be cleaned when stopping the wallet,
+  // but the wallet may be closed unexpectedly
+  const storage = STORE.getStorage();
+  yield storage.store.cleanMetadata();
+  yield storage.cleanStorage(true);
```
pedroferreira1, line 134: Are we keeping the access data as before?

src/sagas/wallet.js
```diff
@@ -205,11 +218,6 @@ export function* startWallet(action) {
     }
   }
 
-  walletUtil.storePasswordHash(pin);
```
pedroferreira1, line 208: You are storing the password somewhere else now, right?

src/sagas/wallet.js
```diff
@@ -537,6 +547,8 @@ export function* setupWalletListeners(wallet) {
     }));
 
     return () => {
+      // XXX: Is this cleanup working?
```
pedroferreira1, line 550: It seems like it's not working. Should we create an issue/KTLO for that?

src/store.js
```diff
@@ -29,18 +40,213 @@ class AsyncStorageStore {
...
+   * This will keep biometry keys and notification keys for example.
+   *
+   * @param {boolean} [onlyWalletKeys=false] If true, will delete only keys starting with 'wallet'
+   * @returns {Promise<void>}
+   */
+  async clearItems(onlyWalletKeys = false) {
+    const keys = await AsyncStorage.getAllKeys() || [];
+    this.hathorMemoryStorage = {};
+    if (onlyWalletKeys) {
+      await AsyncStorage.multiRemove(keys.filter((key) => key.startsWith('wallet')));
```
pedroferreira1, line 58: Maybe remove the keys that start with `wallet:`? I don't remember all the keys we have and if we might add keys that start with "wallet" but not as a prefix.

src/screens/PinScreen.js
```diff
@@ -132,50 +138,47 @@ class PinScreen extends React.Component {
...
+      const storage = STORE.getStorage();
+      await storage.saveAccessData(newAccessData);
+
+      // The access data is saved on the new storage, we can delete the old data.
+      // This will only delete keys with the wallet prefix, so we don't delete
+      // the biometry keys and new data.
+      await STORE.clearItems(true);
+
+      // We have finished the migration so we can set the storage version to the most recent one.
+      STORE.updateStorageVersion();
```
pedroferreira1, line 162: Should we set the storage version outside the `if`? You would be setting it only when you need the data migration but I feel it might be an important information, so maybe we could add it when starting the wallet as well, what do you think?

src/sagas/errorHandler.js
```diff
@@ -42,8 +37,16 @@ export const sentryReportError = (error) => {
...
+  };
 
-    scope.setExtra('Access information', data);
+  Sentry.init({
+    dsn: SENTRY_DSN,
+  });
+  Sentry.withScope((scope) => {
+    scope.setExtra('App version', JSON.stringify(VersionNumber));
+    scope.setExtra('Access information', removePersonalData(STORE.getAccessData()));
+    scope.setExtra('Loaded access information', removePersonalData(loadedAccessData));
```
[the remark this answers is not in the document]
r4mmer: We did, but it was removed from the `STORE` so we will remove it from here as well

src/sagas/wallet.js
r4mmer: Yes, `cleanMetadata` and `cleanStorage` only affect tokens, history, utxos and possibly loaded addresses (which would require `storage.cleanStorage(true, true)`).

This means we are persisting access data and addresses.
The addresses should not be an issue unless we want to change the network, which is not currently possible.

src/sagas/wallet.js
r4mmer: Yes, it is saved in the accessData, on the storage.
So it will be created where we create the accessData, which is the start method of the facade.

src/sagas/wallet.js
r4mmer: It is a memory leak, if the handlers are being kept.
Although I don't think it is a "serious" memory leak since it only happens when we are starting the wallet, which should not happen many times during the app lifecycle.

I'll create an issue for this and link to this line, what do you think?

src/store.js
r4mmer: Done

src/screens/PinScreen.js
r4mmer: Done

src/screens/UnregisterToken.js
```diff
@@ -73,7 +74,15 @@ class UnregisterToken extends React.Component {
       return;
     }
 
-    const promise = hathorLib.tokens.unregisterToken(tokenUnregister);
+    // XXX: maybe we should create a new action `removeToken`
+    // so we dont need to get all registered tokens to call setTokens
+    const promise = this.props.storage.unregisterToken(tokenUnregister).then(async () => {
+      const newTokens = [];
+      for await (const token of this.props.storage.getRegisteredTokens()) {
+        newTokens.push(token);
```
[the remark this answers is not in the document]
r4mmer: Changed to only include the token data fields

pedroferreira1 (approved):

PR #281, opened 2023-05-29
feat: custom async storage store

r4mmer:
### Acceptance Criteria
- Do not use `for await` syntax in any saga.
- Create a `HybridStore` that extends `MemoryStore` but with a persistent access data.
- Remove features specififc to the `LevelDBStore`
- Remove `fromXpriv` argument from start wallet saga
- Fix issues found when debugging the app.
  - TxHistory contructor expects `isVoided`
  - fix storage initialization for new wallets
  - use new operator on address class initialization
  - fix typo on `numberUtils` helpers

### Security Checklist

### DIFF
--- a/src/sagas/helpers.js
+++ b/src/sagas/helpers.js
@@ -5,6 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
+import hathorLib from '@hathor/wallet-lib';
 import { get } from 'lodash';
 import {
   put,
@@ -137,4 +138,31 @@ export const showPinScreenForResult = async (dispatch) => new Promise((resolve) 
 export function isUnlockScreen(action) {
   return action.type === types.SET_LOCK_SCREEN
     && action.payload === false;
+}
+
+/**
+ * Get registered tokens from the wallet instance.
+ * @param {HathorWallet} wallet
+ * @param {boolean} excludeHTR If we should exclude the HTR token.
+ * @returns {string[]}
+ */
+export async function getRegisteredTokens(wallet, excludeHTR = false) {
+  const htrUid = hathorLib.constants.HATHOR_TOKEN_CONFIG.uid;
+  const tokens = [];
+
+  // redux-saga generator magic does not work well with the "for await..of" syntax
+  // The asyncGenerator is not recognized as an iterable and it throws an exception
+  // So we must iterate manually, awaiting each "next" call
+  const iterator = wallet.storage.getRegisteredTokens();
+  let next = await iterator.next();
+  while (!next.done) {
+    const token = next.value;
+    if ((!excludeHTR) || token.uid !== htrUid) {
+      tokens.push(token.uid);
+    }
+    // eslint-disable-next-line no-await-in-loop
+    next = await iterator.next();
+  }
+
+  return tokens;
 }
--- a/src/sagas/tokens.js
+++ b/src/sagas/tokens.js
@@ -127,7 +127,7 @@ function* fetchTokenHistory(action) {
     }
 
     const response = yield call(wallet.getTxHistory.bind(wallet), { token_id: tokenId });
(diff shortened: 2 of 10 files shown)


src/sagas/helpers.js
```diff
@@ -138,3 +139,30 @@ export function isUnlockScreen(action) {
...
     && action.payload === false;
 }
+
+/**
+ * Get registered tokens from the wallet instance.
+ * @param {HathorWallet} wallet
+ * @param {boolean} excludeHTR If we should exclude the HTR token.
+ * @returns {string[]}
+ */
+export async function getRegisteredTokens(wallet, excludeHTR = false) {
```
r4mmer, line 149: Redux-saga does not support iterating on async generators (see [ref](https://github.com/redux-saga/redux-saga/blob/8f60bcb27b088b4785bbd106eaf6cded13d99562/packages/core/src/internal/proc.js#L13)) so we cannot use the `for await` syntax to get the registered tokens from the storage.

The solution is to manually iterate on the asnc generator, this was encased in this helper.
If we need to create more helpers like this we can create a more generic `asyncGeneratorToArray` with the while logic.

src/sagas/wallet.js
```diff
@@ -115,7 +116,6 @@ export function* startWallet(action) {
   const {
     words,
     pin,
-    fromXpriv,
```
r4mmer, line 118: `fromXpriv` is not needed anymore since starting from the peristed access data is faster

src/sagas/wallet.js
```diff
@@ -291,16 +279,7 @@ export function* loadTokens() {
 
   const wallet = yield select((state) => state.wallet);
 
-  const registeredTokens = call(async () => {
-    const tokens = [];
-    for await (const token of wallet.storage.getRegisteredTokens()) {
```
r4mmer, line 296: `wallet.storage.getRegisteredTokens` exposes an `asyncIterator` method so the runtime can iterate properly, but redux-saga only allows for iterations on objects with an `iterator` method. The async generator is not identified as an iterable and the for syntax will fail.

This was solved by the `getRegisteredTokens` helper which iterates manually so it does not rely on the runtime lib.

src/utils.js
```diff
@@ -336,7 +337,7 @@ export const mapTokenHistory = (element, token) => {
     timestamp: element.timestamp,
     balance: element.balance,
     // in wallet service this comes as 0/1 and in the full node comes with true/false
-    voided: Boolean(element.voided),
+    isVoided: Boolean(element.voided),
```
r4mmer, line 340: The `TxHistory` constructor expects `isVoided` and not `voided`.

src/sagas/tokens.js
```diff
@@ -127,7 +127,7 @@ function* fetchTokenHistory(action) {
     }
 
     const response = yield call(wallet.getTxHistory.bind(wallet), { token_id: tokenId });
-    const data = response.map((txHistory) => mapTokenHistory(txHistory, tokenId));
+    const data = response.map((item) => mapTokenHistory(item, tokenId));
```
andreabadesso, line 130: why?

r4mmer: I did some tests and refactors in this method, the necessary change went to the lib and when i went to undo the changes I forgot this one.

src/store.js
```diff
@@ -12,6 +12,7 @@ import { NETWORK } from './constants';
 
 
 const ACCESS_DATA_KEY = 'asyncstorage:access';
+const REGISTERED_TOKENS_KEY = 'asyncstorage:registeredTokens';
```
andreabadesso, line 15: We'll need to migrate users when they're updating their apps

r4mmer: Done

pedroferreira1 (approved):

PR #284, opened 2023-06-08
fix: wallet reset cleanup

r4mmer:
### Acceptance Criteria
- When restarting the wallet we should fully clean any persisted data
- The feature toggle cleanup was interfering with the wallet reset
- Loading the tokens into redux should not use "for await" 
- Remove unnecessary copy of the storage instance.

### Security Checklist

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -313,7 +313,7 @@ class _AppStackWrapper extends React.Component {
     const version = DeviceInfo.getVersion();
     // We use this string to parse the version from user agent
     // in some of our services, so changing this might break another service
-    if (this.props.wallet && this.props.wallet.storage) {
+    if (this.props.wallet?.storage) {
       this.props.wallet.storage.config.setUserAgent(`Hathor Wallet Mobile / ${version}`);
     }
     // set notification foreground listener
--- a/src/App.js
+++ b/src/App.js
@@ -21,7 +21,7 @@ import hathorLib from '@hathor/wallet-lib';
 
 import IconTabBar from './icon-font';
 import NavigationService from './NavigationService';
-import { IS_MULTI_TOKEN, PRIMARY_COLOR, LOCK_TIMEOUT, PUSH_ACTION } from './constants';
+import { IS_MULTI_TOKEN, PRIMARY_COLOR, LOCK_TIMEOUT, PUSH_ACTION, INITIAL_TOKENS } from './constants';
 import { setSupportedBiometry } from './utils';
 import {
   resetData,
@@ -313,7 +313,9 @@ class _AppStackWrapper extends React.Component {
     const version = DeviceInfo.getVersion();
     // We use this string to parse the version from user agent
     // in some of our services, so changing this might break another service
-    this.props.wallet.storage.config.setUserAgent(`Hathor Wallet Mobile / ${version}`);
+    if (this.props.wallet && this.props.wallet.storage) {
+      this.props.wallet.storage.config.setUserAgent(`Hathor Wallet Mobile / ${version}`);
+    }
     // set notification foreground listener
     this.setNotifeeForegroundListener();
   }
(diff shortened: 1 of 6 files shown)


src/App.js
```diff
@@ -313,7 +313,9 @@ class _AppStackWrapper extends React.Component {
     const version = DeviceInfo.getVersion();
     // We use this string to parse the version from user agent
     // in some of our services, so changing this might break another service
-    this.props.wallet.storage.config.setUserAgent(`Hathor Wallet Mobile / ${version}`);
+    if (this.props.wallet && this.props.wallet.storage) {
```
tuliomir, line 316: :bulb:  This and the `updateReduxTokens` method can be simplified with an [optional chaining](https://javascript.info/optional-chaining#optional-chaining)
```suggestion
    if (this.props.wallet?.storage) {
```

src/store.js
```diff
@@ -170,7 +183,6 @@ class AsyncStorageStore {
    */
   async initStorage(seed, pin) {
     // Remove old storage if it exists
```
tuliomir, line 185: Should we remove this comment too?

src/store.js
```diff
@@ -183,17 +195,16 @@ class AsyncStorageStore {
     await storage.saveAccessData(accessData);
   }
 
+  /* eslint-disable class-methods-use-this */
```
tuliomir, line 198: Why not change `getStorage` to a `static` method?

Maybe rename it to `generateNewStorage` or something like that, also, to indicate it is not fetching anything that already exists, but rather creating something new and empty.

src/sagas/featureToggle.js
```diff
@@ -144,25 +144,19 @@ export function* monitorFeatureFlags(currentRetry = 0) {
 
 export function* setupUnleashListeners(unleashClient) {
   const channel = eventChannel((emitter) => {
-    const listener = (state) => emitter(state);
+    const l1 = () => emitter({ type: 'FEATURE_TOGGLE_UPDATE' });
+    const l2 = () => emitter({ type: 'FEATURE_TOGGLE_READY' });
+    const l3 = (err) => emitter({ type: 'FEATURE_TOGGLE_ERROR', data: err });
```
tuliomir, line 149: Is there any specific reason we changed the event listeners from anonymous functions to `const` variables?

r4mmer: No specific reason, I had to remove the cleanup below and figured I would create listeners as it seemed like the intention with the old `listener` but I does not have any deep reasoning behind it

src/store.js
r4mmer: Created a new `generateStorage` helper to remove this from the class methods

src/store.js
r4mmer: Removed

src/App.js
r4mmer: Done

src/store.js
r4mmer: Actually, other methods are dependent on this method as well, and it is better to have all of these as method on the `STORE` instead of separate helpers so I changed to have it as a method with the eslint skip comment again

src/App.js
```diff
@@ -353,14 +355,24 @@ class _AppStackWrapper extends React.Component {
    * These tokens are known as 'registered' tokens
    */
   updateReduxTokens = async () => {
-    const tokens = [];
-    for await (const tokenData of this.props.wallet.storage.getRegisteredTokens()) {
+    if ((!this.props.wallet) || (!this.props.wallet.storage)) {
```
tuliomir, line 358: This could also benefit from optional chaining
```suggestion
    if (!this.props.wallet?.storage) {
```

tuliomir (approved):

andreabadesso (approved):

PR #287, opened 2023-06-15
chore: remove rc from ios

r4mmer:
### Motivation

Apple does not allow the `-rcX` suffix for the release version.
This [link](https://developer.apple.com/library/archive/technotes/tn2420/_index.html#//apple_ref/doc/uid/DTS40016603-CH1-NUMBERING_CONVENTIONS) explain the format required by apple, they allow `X`, `X.X` and `X.X.X` where each X is a number.

### Explanation

A release is identifies by 2 versions, the version number (internally called `CFBundleShortVersionString`) and the build number (internally called `CFBundleVersion`). Our build scripts use the `MARKETING_VERSION` for the version number and we have a fixed 1 for the build number.

### Proposal

Since both these versions serve as identity the proposed way of working would be:
`(0.26.0, 0.1.0)` for rc1 -> `(0.26.0, 0.2.0)` for rc2 and so on until we have a release version, which would be `(0.26.0, 1.0.0)`.
For the next version number we can reset the build number and start from 0.1.0, i.e. `(0.26.1, 0.1.0)` for rc1 -> `(0.26.1, 0.2.0)` for rc2 ...

This means that the express version number in ios would not match the same release on the android store while we have a release candidate for the internal testers, but it would be the same when we make the release public.

To make the app easier to identify as a release candidate, we will check the _major_ of the build number, if we find a 0 we will add the _minor_ as the release candidate version to the end of the version number, for instance `(0.26.0, 0.2.0)` would be `0.26.0-rc2` and this would be displayed as the version on the `About` screen.

On the `About` screen we will also use the _minor_ of the build version to show as the build number, which will most likely always show as build 0 unless we need to rebuild a version.

### Acceptance Criteria
- Remove rc from ios version number
- Update check_version script to check for the correct version in ios
- Show the version and release candidate version on the About screen as described above.

### Security Checklist


scripts/check_version
```diff
@@ -6,26 +6,34 @@ TEMP_IOS_VAR=`perl -n -e '/MARKETING_VERSION = (.*);/ && print "$1\n"' ./ios/Hat
 IOS_PROJECT_VERSION_1=`echo $TEMP_IOS_VAR | awk '{print $1}'`
 IOS_PROJECT_VERSION_2=`echo $TEMP_IOS_VAR | awk '{print $2}'`
 
+# This will read everything of the string up to the first dash.
+IFS='-'
```
msbrogli, line 10: Why not `STRIPPED_PACKAGE_VERSION=$(echo $PACKAGE_VERSION| cut -d- -f1)`?

msbrogli (dismissed):

scripts/check_version
r4mmer: I took this from a split script I had, but i'll update to using `cut` since its more concise and clearer

scripts/check_version
```diff
@@ -3,29 +3,34 @@
...
+  echo Version different in package.json \($PACKAGE_VERSION\) and first MARKETING_VERSION \($IOS_MARKETING_VERSION_1\) on ios/HathorMobile.xcodeproj/project.pbxproj
 	EXITCODE=-1
 fi
 
-if [[ x${PACKAGE_VERSION}x != x${IOS_PROJECT_VERSION_2}x ]]; then
-	echo Version different in package.json and second MARKETING_VERSION on ios/HathorMobile.xcodeproj/project.pbxproj
+if [[ x${STRIPPED_PACKAGE_VERSION}x != x${IOS_MARKETING_VERSION_2}x ]]; then
+	echo Version different in package.json \($PACKAGE_VERSION\) and second MARKETING_VERSION \($IOS_MARKETING_VERSION_2\) on ios/HathorMobile.xcodeproj/project.pbxproj
 	EXITCODE=-1
 fi
```
pedroferreira1, line 35: Can we add a new check here if the `CURRENT_PROJECT_VERSION` is in the correct format `X.Y.Z`?

If we are going to throw error in the About page, we must guarantee that this version is correct here.

src/screens/About.js
```diff
@@ -40,6 +41,27 @@ export class About extends React.Component {
     }
   }
 
+  renderBuildVersion() {
+    if (Platform.OS === 'android') {
+      // Android
+      return (<Text style={this.style.text}>{`v${VersionNumber.appVersion} (build ${VersionNumber.buildVersion})`}</Text>);
```
pedroferreira1, line 47: Maybe create a method that receive version and build and return this html? It would help us if we decide to change this in the future, so we don't need to change in 3 different places.

r4mmer: This is why I created the `renderBuildVersion` method, to isolate this logic, do you think this should be a helper?

scripts/check_version
r4mmer: Done

RELEASING.md
```diff
@@ -26,6 +26,12 @@ In the `ios/HathorMobile.xcodeproj/project.pbxproj`:
 
   > App Store uses the pair `(CFBundleShortVersionString, CFBundleVersion)` to identify a release. They are both from `ios/HathorMobile/Info.plist`.
 
+- Set the `CURRENT_PROJECT_VERSION` to a valid build version
+
+  > If the version is the first release candidate, the `CURRENT_PROJECT_VERSION` should be `0.1.0` and we should increate the minor version for each new release candidate, i.e. v1.0.0-rc5 would give us `CURRENT_PROJECT_VERSION = 0.5.0`.
+  > The major version should be 0 for release candidates and 1 for releases.
+  > The patch (last digit) of the `CURRENT_PROJECT_VERSION` should be always `0` unless we are re-launching the version.
```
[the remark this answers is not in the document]
r4mmer: Changed

RELEASING.md
```diff
@@ -26,6 +26,12 @@ In the `ios/HathorMobile.xcodeproj/project.pbxproj`:
 
   > App Store uses the pair `(CFBundleShortVersionString, CFBundleVersion)` to identify a release. They are both from `ios/HathorMobile/Info.plist`.
 
+- Set the `CURRENT_PROJECT_VERSION` to a valid build version
+
+  > If the version is the first release candidate, the `CURRENT_PROJECT_VERSION` should be `0.1.0` and we should increate the minor version for each new release candidate, i.e. v1.0.0-rc5 would give us `CURRENT_PROJECT_VERSION = 0.5.0`.
```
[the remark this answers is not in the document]
r4mmer: Done

src/screens/About.js
pedroferreira1: Not a helper in the project, just the `Text` line could be in a different method, if we want to change the display in the future for any reason. I don't think it's critical

pedroferreira1 (approved):

andreabadesso (approved):

PR #291, opened 2023-06-27
Fix/migration from previous version

r4mmer:
### Acceptance Criteria
- Use correct data structure when checking pin before migration
- Do not clear all keys when migrating the access data
- Update registered tokens when loading a wallet

### Security Checklist

### DIFF
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -63,6 +63,7 @@ import {
   types,
   setAvailablePushNotification,
   resetWalletSuccess,
+  setTokens,
 } from '../actions';
 import { fetchTokenData } from './tokens';
 import {
@@ -279,22 +280,26 @@ export function* loadTokens() {
 
   const wallet = yield select((state) => state.wallet);
 
-  const registeredTokens = yield getRegisteredTokens(wallet, true);
+  const registeredTokens = yield getRegisteredTokens(wallet);
+
+  yield put(setTokens(registeredTokens));
+
+  const registeredUids = registeredTokens.map(t => t.uid);
 
   // We don't need to wait for the metadatas response, so we can just
   // spawn a new "thread" to handle it.
   //
   // `spawn` is similar to `fork`, but it creates a `detached` fork
-  yield spawn(fetchTokensMetadata, registeredTokens);
+  yield spawn(fetchTokensMetadata, registeredUids);
 
   // Since we already know here what tokens are registered, we can dispatch actions
   // to asynchronously load the balances of each one. The `put` effect will just dispatch
   // and continue, loading the tokens asynchronously
-  for (const token of registeredTokens) {
+  for (const token of registeredUids) {
     yield put(tokenFetchBalanceRequested(token));
   }
 
-  return registeredTokens;
+  return registeredUids;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/sagas/wallet.js
```diff
@@ -279,22 +280,26 @@ export function* loadTokens() {
 
   const wallet = yield select((state) => state.wallet);
 
-  const registeredTokens = yield getRegisteredTokens(wallet, true);
+  const registeredTokens = yield getRegisteredTokens(wallet);
+
+  yield put(setTokens(registeredTokens));
```
r4mmer, line 285: When loading a new wallet we need to update the registered tokens on the state so they will show on the dashboard

src/screens/PinScreen.js
```diff
@@ -192,13 +192,26 @@ class PinScreen extends React.Component {
         return;
       }
 
-      const pinCorrect = cryptoUtils.checkPassword(accessData.words, pin);
+      let wordsEncryptedData = accessData.words
+      if (!accessData.words.data) {
```
r4mmer, line 196: In the old format, `words` is a string, but in the new one it is an object, so we need to create the encrypted data object from the data present on the access data.

src/screens/UnregisterToken.js
```diff
@@ -78,12 +78,21 @@ class UnregisterToken extends React.Component {
     // so we dont need to get all registered tokens to call setTokens
     const promise = this.props.storage.unregisterToken(tokenUnregister).then(async () => {
       const newTokens = [];
-      for await (const token of this.props.storage.getRegisteredTokens()) {
```
r4mmer, line 81: We need to remove all uses of `for await` syntax from the codebase

src/store.js
```diff
@@ -147,11 +147,17 @@ class AsyncStorageStore {
    */
   async clearItems(onlyWalletKeys = false) {
     const keys = await AsyncStorage.getAllKeys() || [];
-    this.hathorMemoryStorage = {};
```
r4mmer, line 150: The actual source of truth is the `AsyncStorage` so cleaning this will not erase the storage, unless we try to read before repopulating the `hathorMemoryStorage` which was happening, to fix this we only delete the required keys

src/screens/PinScreen.js
```diff
@@ -192,13 +192,26 @@ class PinScreen extends React.Component {
         return;
       }
 
-      const pinCorrect = cryptoUtils.checkPassword(accessData.words, pin);
+      let wordsEncryptedData = accessData.words;
+      if (!accessData.words.data) {
+        // This is from a previous version
```
tuliomir, line 197: I would suggest naming the version this refers to, like: "Versions before 0.26.0 didn't have the `data` property".

I think it would make it clearer when it's OK to remove this validation down the road.

tuliomir (approved):

src/screens/PinScreen.js
r4mmer: It may never be ok to remove this validation because anyone with a version prior to v0.26.0 may upgrade to the most recent version and not having this would require the person to import the wallet from scratch

andreabadesso (approved):

PR #298, opened 2023-07-15
Fix: Navigation background colors

tuliomir:
### Acceptance Criteria
- The navigators from React Navigator 5 should have a background the same color as the light theme on the native OS

### Additional refactors
Centralized two other colors that were being used throughout many files. These will help with an eventual adaptation of the whole application to allow dark mode later.
- A light background, but noticeably darker than white for giving a sense of container border
- A red color used in error messages and the offline panel background

### Security Checklist

### DIFF
--- a/src/styles/init.js
+++ b/src/styles/init.js
@@ -5,10 +5,15 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import { StyleSheet } from 'react-native';
+import { Platform, StyleSheet } from 'react-native';
 import { PRIMARY_COLOR } from '../constants';
 
-const defaultBackgroundColor = '#fafafa';
+/**
+ * TODO: Obtain color from the OS scheme
+ * @see: https://reactnavigation.org/docs/5.x/themes/#using-the-operating-system-preferences
+ * @type {string}
+ */
+const defaultBackgroundColor = (Platform.OS === 'ios') ? '#fff' : '#fafafa';
 
 const baseStyle = StyleSheet.create({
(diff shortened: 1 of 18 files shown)


src/styles/init.js
```diff
@@ -5,10 +5,15 @@
...
+import { Platform, StyleSheet } from 'react-native';
 import { PRIMARY_COLOR } from '../constants';
 
-const defaultBackgroundColor = '#fafafa';
+/**
+ * TODO: Obtain color from the OS scheme
+ * @see: https://reactnavigation.org/docs/5.x/themes/#using-the-operating-system-preferences
+ * @type {string}
+ */
+const defaultBackgroundColor = (Platform.OS === 'ios') ? '#fff' : '#fafafa';
```
tuliomir, line 16: This is the actual bugfix: changing the color from plain white to a light gray depending on the OS's native color ( when on the light theme )

alexruzenhack, line 16: 🚲 Shouldn't these colors `#fff` and `#fafafa` be in the `../constants` file as well?

alexruzenhack (approved):

src/styles/init.js
tuliomir: Good point. Refactored on [e4eb918](https://github.com/HathorNetwork/hathor-wallet-mobile/pull/298/commits/e4eb9185b88bd150db7ccf2057b3763aa8622705).

andreabadesso (approved):

PR #299, opened 2023-07-15
fix: Navigator header top padding

tuliomir:
When bumping the Navigation lib to v5 and adding the recommended `react-native-safe-area-context` lib, the previous visual adjustments for the app's top margin became too wide. This PR attempts to fix this with the best solution available.

Further below, two possible solutions are discussed for reference:
- Implementing a negative top margin
- Centralizing the header contents

The implemented solution on this PR's code is the header centralization. The images below explain the tradeoffs between these solutions.

### Acceptance Criteria
- The top padding for the navigators should not be excessively far from the screen top
- The `SafeAreaView` component should only be used on the outer views that interface with the OS directly, not on their children
  - For the current app state, this means the only time this component will be used is on the `RootStack` Stack Navigation instance, on the `App.js` file

### Header vertical alignment
This approach only improves the alignment, but leaves the insets sizes to the lib to decide.

The first row displays the Android exhibition, highlighting each of the header's containers:
- A dark yellow for the whole Header
- A blue for the screen title
- A purple for the right button

The second row displays the exhibition on iOS, with a purple color highlighting the "Safe Area" that is added by the OS on the screen top.
Mobile Padding-V-Align

### Negative top margin
This approach forces the content up against the Safe Area, potentially causing exhibition issues.

The first row shows the exhibition on Android devices. The second row shows the exhibition on iOS, with a purple highlighting the "Safe Area" that is added by the OS on the screen top.
Mobile Padding-Neg Margin

### Comparison
Mobile Padding-Comparison

### Security Checklist

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -575,7 +575,10 @@ NavigationService.setTopLevelNavigator(navigationRef);
 const App = () => (
   <SafeAreaProvider>
     <Provider store={store}>
-      <SafeAreaView style={{ flex: 1 }}>
+      <SafeAreaView
+        edges={['top', 'right', 'left']}
+        style={{ flex: 1 }}
+      >
         <NavigationContainer
           ref={navigationRef}
         >
--- a/src/components/HathorHeader.js
+++ b/src/components/HathorHeader.js
@@ -77,7 +77,7 @@ const styles = StyleSheet.create({
   wrapper: {
     height: 56,
     flexDirection: 'row',
-    alignItems: 'flex-end',
+    alignItems: 'center',
     borderColor: '#eee',
     paddingHorizontal: 16,
     backgroundColor: baseStyle.title.backgroundColor,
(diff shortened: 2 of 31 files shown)


src/App.js
```diff
@@ -575,7 +575,10 @@ NavigationService.setTopLevelNavigator(navigationRef);
 const App = () => (
   <SafeAreaProvider>
     <Provider store={store}>
-      <SafeAreaView style={{ flex: 1 }}>
+      <SafeAreaView
+        edges={['top', 'right', 'left']}
```
tuliomir, line 579: Avoids rendering on the bottom margin, as it consumes significant space in return for a user experience that does not seem to be improved.

src/components/HathorHeader.js
```diff
@@ -77,7 +77,7 @@ const styles = StyleSheet.create({
   wrapper: {
     height: 56,
     flexDirection: 'row',
-    alignItems: 'flex-end',
+    alignItems: 'center',
```
tuliomir, line 80: Proposed solution: centralization of header contents.

andreabadesso (approved):

PR #300, opened 2023-07-19
fix: Initial screen rendering

tuliomir:
The "Welcome" screen was being rendered while the storage fetched information about the loaded wallet. A blank screen should be shown instead, to avoid content flickering.

### Acceptance Criteria
- Nothing should be rendered until the application decides whether the Wallet is loaded or not

### Security Checklist

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -646,6 +646,10 @@ const RootStack = () => {
 
   const Stack = createStackNavigator();
 
+  /*
+   * XXX: Screens within the Root Stack have no transition animation, as they are processed before
+   * any user interface.
+   */
   return (
     <Stack.Navigator
       initialRouteName='Decide'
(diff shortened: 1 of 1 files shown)


src/App.js
```diff
@@ -518,13 +518,13 @@ const BlankScreen = () => null;
  */
 const RootStack = () => {
   const dispatch = useDispatch();
-  const [isLoaded, setIsLoaded] = useState(false);
+  const [appStatus, setAppStatus] = useState('initializing');
```
alexruzenhack, line 521: 💬 When the 'initializing' status is used?

alexruzenhack (approved):
I'm in doubt about the 'initializing' appStatus. However, I don't see any harm to approve it.

src/App.js
tuliomir: Previously there were only two states: `isLoaded: false` and `isLoaded: true`.

The problem was that the screen rendering was happening before the storage could answer the promise with the correct information. So, for a brief moment the rendering was incorrect and then the screen flickered to the correct one.

The state `appStatus: initializing` ensures no render will occur until the actual state of the wallet is fetched: either loaded or not.

As this only takes a second or two, I didn't bother adding a `loading` animation, but just a blank screen as was the behavior previously of our implementation on React Navigation 4.

andreabadesso:
@tuliomir It's good that the screen behind doesn't get rendered, but the flicker still happens:


https://github.com/HathorNetwork/hathor-wallet-mobile/assets/3586068/c7a65d2b-a2c1-408a-a5bd-dc78266ee225

andreabadesso (changes requested):
@tuliomir It's good that the screen behind doesn't get rendered, but the flicker still happens:


https://github.com/HathorNetwork/hathor-wallet-mobile/assets/3586068/c7a65d2b-a2c1-408a-a5bd-dc78266ee225

src/App.js
```diff
@@ -654,9 +658,27 @@ const RootStack = () => {
...
+        component={AppStackWrapper}
+        options={{
+          animationEnabled: false,
+        }}
+      />
+      <Stack.Screen
+        name='Init'
+        component={InitStack}
+        options={{
+          animationEnabled: false,
```
andreabadesso, line 679: Maybe do something like:

```javascript
const defaultOptions = { animationEnabled: true };
```

and use it on the screens

andreabadesso (approved):

PR #301, opened 2023-07-19
refactor: Saga camera permissions

tuliomir:
### Acceptance Criteria
- Camera permissions should be requested through saga instead of directly on a screen
- "Register Token" screen should go directly to "Manual Input" when camera permission is blocked

### Security Checklist

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -135,10 +135,12 @@ const SendStack = ({ navigation }) => {
   const dispatch = useDispatch();
 
   /*
-   * Request camera permission on initialization
+   * Request camera permission on initialization, if permission is not already set
    */
   useEffect(() => {
-    dispatch(requestCameraPermission());
+    if (isCameraAvailable === null) {
+      dispatch(requestCameraPermission());
+    }
   }, []);
 
   // Listen to camera permission changes from user input and navigate to the relevant screen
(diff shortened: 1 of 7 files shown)


src/components/QRCodeReader.js
```diff
@@ -62,8 +62,8 @@ export default ({
     setIsFocusedScreen(true);
 
     return () => {
-      focusEvent.remove();
-      blurEvent.remove();
+      focusEvent();
+      blurEvent();
```
tuliomir, line 66: React Navigation v5.x event listeners do not have the `remove` method, and can be removed by calling them directly. ( See [API Docs](https://reactnavigation.org/docs/5.x/navigation-events#navigationaddlistener) )

src/App.js
```diff
@@ -163,16 +202,63 @@ const CreateTokenStack = () => {
...
+   * @returns {string} Route name
+   */
+  const decideRouteByCameraAvailablity = (cameraStatus) => {
+    switch (isCameraAvailable) {
+      case true:
+        return 'RegisterToken';
+      case false:
+        return 'RegisterTokenManual';
+      default:
+        return 'RegisterCameraPermissionScreen';
```
alexruzenhack, line 222: ❓ Shouldn't it be just `'CameraPermissionScreen'`?

src/sagas/permissions.js
```diff
@@ -0,0 +1,54 @@
...
+ * For a complete description on all possible permition statuses,
+ * @see https://github.com/zoontek/react-native-permissions#permissions-statuses
+ */
+export function* requestCameraPermissions() {
+  const checkStatus = async (status) => {
+    switch (status) {
+      case RESULTS.GRANTED:
+        return true;
+      case RESULTS.DENIED: // The permission has not been requested / is denied but requestable
+        return checkStatus(await request(platformCameraPermission));
```
alexruzenhack, line 25: ❓ Can it fall in infinity loop? Have you tested the case in which the used denied the permission over and over again?

alexruzenhack (commented):
There is a possible bug in `src/sagas/permission`, I think it worths a double check.

src/sagas/permissions.js
tuliomir: Actually this implementation fixes an infinity loop that happened before it. The simplified permissions flow works like this:
- If the user has granted unlimited permission to the application, the status returned by `check()` is `GRANTED`
- In all other cases, the status is `DENIED`.

Upon `request()`, the result can then be either `GRANTED` or `BLOCKED`. The `DENIED` is never repeated.
So, there is no loop condition unless the `react-native-permissions` lib introduces a breaking change.

alexruzenhack: Great!

alexruzenhack (approved):

andreabadesso (approved):

PR #307, opened 2023-07-27
refactor: Remove callback from screen nav params

tuliomir:
Since the upgrade to navigation v5, some screen navigations trigger a [`Non-serializable values were found in the navigation state`](https://reactnavigation.org/docs/troubleshooting/#i-get-the-warning-non-serializable-values-were-found-in-the-navigation-state) alert.

The solution for this was to remove callbacks from the `ChooseToken` screen and replace it with existing saga flows:
- The `onItemPressed` callback function passed as `prop` was removed
- The `ChooseToken` screen updates directly the `selectedToken` redux state
- The screens that call it listen to that update directly from the state

### Acceptance Criteria
- All navigation parameters on the app should be serializable

### Security Checklist

### DIFF
--- a/src/components/TokenSelect.js
+++ b/src/components/TokenSelect.js
@@ -19,6 +19,22 @@ import { LIGHT_BG_COLOR, PRIMARY_COLOR } from '../constants';
 import { getLightBackground, renderValue, isTokenNFT } from '../utils';
 import { TOKEN_DOWNLOAD_STATUS } from '../sagas/tokens';
 
+/**
+ * @typedef TokenBalance
+ * @property {number} data.available
+ * @property {string} status
+ */
+
+/**
+ * @param {Object} props
+ * @param {Record<string,TokenBalance>} props.tokensBalance
+ * @param {{ uid: string }} props.selectedToken
+ * @param {unknown} props.tokenMetadata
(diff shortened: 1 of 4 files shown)


src/components/TokenSelect.js
```diff
@@ -19,6 +19,22 @@ import { LIGHT_BG_COLOR, PRIMARY_COLOR } from '../constants';
...
+/**
+ * @param {Object} props
+ * @param {Record<string,TokenBalance>} props.tokensBalance
+ * @param {{ uid: string }} props.selectedToken
+ * @param {unknown} props.tokenMetadata
+ * @param {unknown} props.tokens
+ * @param {unknown} props.header
+ * @param {boolean} props.renderArrow
+ * @param {function} props.onItemPress
+ */
```
tuliomir, line 37: Added documentation to aid the investigation process

src/screens/ChangeToken.js
```diff
@@ -55,7 +55,7 @@ class ChangeToken extends React.Component {
 
     return (
       <TokenSelect
-        header=<Header />
+        header={<Header />}
```
tuliomir, line 58: Lint fix

alexruzenhack (approved):

PR #309, opened 2023-08-03
fix: Pin screen adjustments

tuliomir:
### Acceptance Criteria
- Fix PinScreen lateral color difference on some devices
- Avoid flickering the Pin Digits while waiting for the biometry results from OS

### Security Checklist

### DIFF
--- a/src/screens/PinScreen.js
+++ b/src/screens/PinScreen.js
@@ -119,7 +119,7 @@ class PinScreen extends React.Component {
        * }
        */
       const credentials = await Keychain.getGenericPassword({
-        authenticationPrompt: this.biometryText,
+        authenticationPrompt: { title: this.biometryText },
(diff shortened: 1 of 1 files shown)


src/screens/PinScreen.js
```diff
@@ -116,13 +119,15 @@ class PinScreen extends React.Component {
        * }
        */
       const credentials = await Keychain.getGenericPassword({
-        authenticationPrompt: this.biometryText,
+        authenticationPrompt: { title: this.biometryText },
```
tuliomir, line 122: Passing a simple string on this parameter is now deprecated. Changed to the most up-to-date method signature.

alexruzenhack (approved):

andreabadesso (approved):

PR #319, opened 2023-08-30
chore: update api level to 33

alexruzenhack:
### Acceptance Criteria
- update the manifest to targed the api level 33

The only behavior affected by this API level is the [granular media-permission](https://developer.android.com/about/versions/13/behavior-changes-13#granular-media-permissions), however the default behavior provided by the platform is good enough.

### Security Checklist


alexruzenhack:
See https://github.com/HathorNetwork/hathor-wallet-mobile/pull/320

PR #322, opened 2023-09-04
fix: reset wallet

r4mmer:
### Acceptance Criteria
- Allow the wallet to be reset successfully

### Security Checklist

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -166,6 +166,10 @@ class AsyncStorageStore {
   async resetWallet() {
     // This should delete the access data and registered tokens, the only persisted data
     await AsyncStorage.multiRemove(walletKeys);
+    // Delete the same keys on the memory cache
+    for (const key of walletKeys) {
+      delete this.hathorMemoryStorage[key];
+    }
     // This will delete any wallet data of the legacy storage
     await this.clearItems(true);
   }


tuliomir (approved):
I feel the `this.clearItems(true)` on line 174 should already do what this new lines do.

But since this is an internal method within the `store`, it's ok to have bits of redundance. Or would it be a better idea to use the `clearItems` method to also clean these `walletKeys`?

r4mmer:
>I feel the this.clearItems(true) on line 174 should already do what this new lines do.
>
>But since this is an internal method within the store, it's ok to have bits of redundance. Or would it be a better idea to use the clearItems method to also clean these walletKeys?

The `resetWallet` method objective is to clear the wallet keys, so we can create a new wallet without losing other configuration (e.g. server, network, etc.), while `clearItems` is meant to erase all keys. The true argument is so we clear keys only from the legacy storage, this way we can reset a wallet from the locked screen even if the user did not unlock after the upgrade.

src/store.js
```diff
@@ -166,6 +166,10 @@ class AsyncStorageStore {
   async resetWallet() {
     // This should delete the access data and registered tokens, the only persisted data
     await AsyncStorage.multiRemove(walletKeys);
+    // Delete the same keys on the memory cache
```
pedroferreira1, line 169: I feel the `await this.clearItems(true);` line below is already deleting those keys, isn't it?

pedroferreira1: Actually it's not but it's confusing. `onlyWalletKeys` and `walletKeys` don't mean the same thing

r4mmer: Yes, onlyWalletKeys mean the legacy wallet keys which are prefixed with wallet:

pedroferreira1 (approved):

Issue #318, opened 2023-08-25
QA iOS Aug 25th v0.26.0-rc6


pedroferreira1:
Device: iPhone 13 mini
OS version: 16.6

Errors:

- After inactivity in background, wallet goes to Locked screen. Reset wallet in this case crashes the app (sent the error to Sentry). Actually the reset wallet is never working.
- Start wallet button stays frozen for some seconds before start the loading
- Loading forever, wallet without txs didn't load for more than a minute. (happened once)
- Gray color is too light, can't see lines with manual input, and placeholder (payment request)
- Payment request doesn't open keyboard automatically
- Different UI for modals (this is not a bug but the UI for modals has changed)
- Create token modal: X should close the whole sequence of modals, while back should go back to the previous
- Background color for payment request screen with QRCode and send tx is weird
- Each time I click to generate a payment request, the address changes. It should change only if the address is really used
- Payment request confirmation icon error
  - ![IMG_479D7A03D9DF-1](https://github.com/HathorNetwork/hathor-wallet-mobile/assets/3298774/b7e2f43d-8b0a-43ba-afca-2b49144e3ea8)
- Manual info, create token, ... click on screen should close keyboard?
- Locked Wallet screen when FaceID is enabled doesn't show, is full white
- Can't create token because keyboard for amount doesn't close, so I can't click the button
  - ![IMG_7BC82B8F8B88-1](https://github.com/HathorNetwork/hathor-wallet-mobile/assets/3298774/a68c3d96-bd60-4b17-bead-cbf21abc2e06)
- If wallet locks because of inactivity with number keyboard opened bugs
  - ![IMG_0B6E0B0F7F3C-1](https://github.com/HathorNetwork/hathor-wallet-mobile/assets/3298774/d264f6b5-35bb-45cf-a0ac-26462e39e037)

PR #325, opened 2023-09-11
Fix: Issues raised on iOS QA

tuliomir:
Solves items from #318 

### Acceptance Criteria
- Navigating to the "Receive" screen no longer marks an address as used
- "Receive Amount" screen automatically focuses on the Keyboard
- Clicking the "X" on the "Create Token" flow goes back to the dashboard screen
- Clicking "Back" on the "Receive Transaction Details" after receiving a tx, goes back to the dashboard screen
- Correct positioning of the "Receive Transaction Details" confirmation modal
- Correct "Safe Area" insets for modals
- Correct "Safe Area" insets for when the keyboard is being displayed
- Dismisses keyboard whenever the "Pin Screen" is displayed

Tested on simulators for devices:
- iPhone 13 mini
- iPhone 13
- iPhone SE (3rd gen)
- iPhone 14 Pro

### Security Checklist

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -361,7 +361,6 @@ const AppStack = () => {
     >
       <Stack.Navigator
         screenOptions={{
-          presentation: 'modal',
           headerShown: false,
         }}
       >
--- a/src/components/AmountTextInput.js
+++ b/src/components/AmountTextInput.js
@@ -17,7 +17,13 @@ class AmountTextInput extends React.Component {
   }
 
   focus = () => {
-    this.inputRef.current.focus();
+    /* After the focus method is called, the screen is still re-rendered at least once more.
+     * Requesting a delay before the focus command ensures it is executed on the final rendering
+     * of the component.
+     */
+    setTimeout(() => {
+      this.inputRef.current.focus();
+    }, 50);
   }
 
   onChangeText = (text) => {
--- a/src/components/HathorHeader.js
+++ b/src/components/HathorHeader.js
@@ -15,6 +15,7 @@ import Logo from './Logo';
 import chevronLeft from '../assets/icons/chevron-left.png';
 import closeIcon from '../assets/icons/icCloseActive.png';
 import baseStyle from '../styles/init';
+import { HEADER_HEIGHT } from '../constants';
 
 const HathorHeader = (props) => {
   const renderBackButton = () => {
@@ -75,7 +76,7 @@ const HathorHeader = (props) => {
 
 const styles = StyleSheet.create({
   wrapper: {
-    height: 56,
+    height: HEADER_HEIGHT,
     flexDirection: 'row',
     alignItems: 'center',
     borderColor: '#eee',
@@ -95,5 +96,4 @@ const styles = StyleSheet.create({
     alignItems: 'center',
   },
 });
-
 export default HathorHeader;
--- a/src/components/ModalConfirmation.js
+++ b/src/components/ModalConfirmation.js
@@ -11,6 +11,7 @@ import {
 } from 'react-native';
 
 import checkLogo from '../assets/images/icCheckBig.png';
+import { getKeyboardAvoidingViewTopDistance } from '../utils';
 
 class ModalConfirmation extends React.Component {
   /**
@@ -43,8 +44,10 @@ class ModalConfirmation extends React.Component {
 
     const { height, width } = Dimensions.get('window');
 
-    // Prevent the bottom message to be covered by some android phones
-    const marginBottom = Platform.OS === 'android' ? 40 : 16;
+    // Prevents the bottom message from being covered
+    const marginBottom = Platform.OS === 'ios'
+      ? getKeyboardAvoidingViewTopDistance() + 30
+      : 40;
 
     const styles = StyleSheet.create({
       modal: {
--- a/src/components/ModalConfirmation.js
+++ b/src/components/ModalConfirmation.js
@@ -7,11 +7,10 @@
 
 import React from 'react';
 import {
-  Dimensions, Image, Platform, StyleSheet, TouchableWithoutFeedback, View,
+  Dimensions, Image, StyleSheet, TouchableWithoutFeedback, View,
 } from 'react-native';
 
 import checkLogo from '../assets/images/icCheckBig.png';
-import { getKeyboardAvoidingViewTopDistance } from '../utils';
 
 class ModalConfirmation extends React.Component {
   /**
@@ -42,13 +41,9 @@ class ModalConfirmation extends React.Component {
       return null;
     }
 
-    const { height, width } = Dimensions.get('window');
+    const { width } = Dimensions.get('window');
 
     // Prevents the bottom message from being covered
-    const marginBottom = Platform.OS === 'ios'
-      ? getKeyboardAvoidingViewTopDistance() + 30
-      : 40;
-
     const styles = StyleSheet.create({
       modal: {
         position: 'absolute',
@@ -58,11 +53,11 @@ class ModalConfirmation extends React.Component {
         justifyContent: 'flex-end',
         alignItems: 'center',
         zIndex: 3,
-        height,
-        width,
+        height: '100%',
+        width: '100%',
       },
       innerModal: {
-        marginBottom,
+        marginBottom: 32,
         height: 270,
         backgroundColor: 'white',
         alignItems: 'center',
--- a/src/components/ReceiveMyAddress.js
+++ b/src/components/ReceiveMyAddress.js
@@ -23,8 +23,7 @@ export default function ReceiveMyAddress() {
   const lastSharedAddress = useSelector((state) => state.lastSharedAddress);
 
   const getNextAddress = async () => {
-    // Fetch the address but do not mark it as used yet.
-    const { address, index } = await wallet.getCurrentAddress();
+    const { address, index } = await wallet.getNextAddress();
 
     dispatch(sharedAddressUpdate(address, index));
   };
--- a/src/components/ReceiveMyAddress.js
+++ b/src/components/ReceiveMyAddress.js
@@ -23,7 +23,8 @@ export default function ReceiveMyAddress() {
   const lastSharedAddress = useSelector((state) => state.lastSharedAddress);
 
   const getNextAddress = async () => {
-    const { address, index } = await wallet.getNextAddress();
+    // Fetch the address but do not mark it as used yet.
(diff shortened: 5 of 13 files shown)


src/App.js
```diff
@@ -361,7 +361,6 @@ const AppStack = () => {
     >
       <Stack.Navigator
         screenOptions={{
-          presentation: 'modal',
```
tuliomir, line 364: Since the iOS modals exhibition have changed, this property is no longer desired.

src/screens/CreateTokenAmount.js
```diff
@@ -114,7 +114,7 @@ class CreateTokenAmount extends React.Component {
         <HathorHeader
           title={t`CREATE TOKEN`}
           onBackPress={() => this.props.navigation.goBack()}
-          onCancel={() => this.props.navigation.pop()}
+          onCancel={() => this.props.navigation.getParent()?.goBack()}
```
andreabadesso, line 117: What happens if there is no parent?

src/screens/PaymentRequestDetail.js
```diff
@@ -42,16 +43,22 @@ class PaymentRequestDetail extends React.Component {
...
+
     this.modalConfirmation = React.createRef();
   }
 
   async componentDidMount() {
-    // When we create a new payment request we update the address for a new one
-    await this.props.wallet.getNextAddress();
+    // When we create a new payment request, we don't update the address for a new one
+    // This will only happen when it receives a transaction and becomes a used address
+    await this.props.wallet.getCurrentAddress();
```
andreabadesso, line 56: I don't understand what's going on here, did we change the logic for `getCurrentAddress` and `getNextAddress`? When?

src/screens/CreateTokenAmount.js
tuliomir: This `?` operator was put there mostly as a coding best practice, but the idea is that there will always be a parent navigator there.

Not having a parent is a signal that there was a major refactor and our current hierarchy of navigators is different from expected. In that case, maybe throwing an exception would be a better way of informing the developer of the need to change this code as well.

src/screens/PaymentRequestDetail.js
tuliomir: This was requested on item n⁰ 9 on issue 318.

`getNextAddress()` not only fetches the next available address, but also marks the current one as used:

https://github.com/HathorNetwork/hathor-wallet-lib/blob/master/src/new/wallet.js#L563-L565

By changing the call to `getCurrentAddress()` we don't do this marking, and leave it to the wallet's new transaction websocket handler to do so.

src/screens/CreateTokenAmount.js
tuliomir: Removed the _optional chaining_ operator on [a38a087](https://github.com/HathorNetwork/hathor-wallet-mobile/pull/325/commits/a38a08757c5ed5741346ab6db2931fdc0b992d41) .

src/components/AmountTextInput.js
```diff
@@ -17,7 +17,13 @@ class AmountTextInput extends React.Component {
   }
 
   focus = () => {
-    this.inputRef.current.focus();
+    /* After the focus method is called, the screen is still re-rendered at least once more.
```
pedroferreira1, line 20: Why is it happening just now? It was working before, wasn't it?

src/components/HathorHeader.js
```diff
@@ -75,7 +76,7 @@ const HathorHeader = (props) => {
 
 const styles = StyleSheet.create({
   wrapper: {
-    height: 56,
+    height: HEADER_HEIGHT,
```
pedroferreira1, line 79: The other PR you stop using this constant

src/constants.js
```diff
@@ -100,6 +100,12 @@ export const MIN_JOB_ESTIMATION = 1;
  */
 export const PIN_SIZE = 6;
 
+/**
+ * Default Hathor Header height, used for recalculating layout offsets.
+ * @type {number}
+ */
+export const HEADER_HEIGHT = 56;
```
pedroferreira1, line 107: This should be removed in the iOS color PR

src/components/ReceiveMyAddress.js
```diff
@@ -23,7 +23,8 @@ export default function ReceiveMyAddress() {
   const lastSharedAddress = useSelector((state) => state.lastSharedAddress);
 
   const getNextAddress = async () => {
-    const { address, index } = await wallet.getNextAddress();
+    // Fetch the address but do not mark it as used yet.
+    const { address, index } = await wallet.getCurrentAddress();
```
pedroferreira1, line 27: I don't understand this change. This method is to get the next address, not the current address.

src/utils.js
```diff
@@ -243,10 +243,12 @@ function extractAddress(plainText) {
  * @return {number} The top distance
  */
 export const getKeyboardAvoidingViewTopDistance = () => {
-  if (Platform.OS === 'android') {
-    return getStatusBarHeight();
-  }
-  return 0;
+  const statusBarHeight = getStatusBarHeight();
+  const calculatedHeight = (Platform.OS === 'ios')
```
pedroferreira1, line 247: Why did this change for iOS? It was 0 before and now it's bigger than the Android

src/components/ModalConfirmation.js
```diff
@@ -43,8 +44,10 @@ class ModalConfirmation extends React.Component {
 
     const { height, width } = Dimensions.get('window');
 
-    // Prevent the bottom message to be covered by some android phones
-    const marginBottom = Platform.OS === 'android' ? 40 : 16;
+    // Prevents the bottom message from being covered
+    const marginBottom = Platform.OS === 'ios'
+      ? getKeyboardAvoidingViewTopDistance() + 30
```
pedroferreira1, line 49: Two questions here:

1. Why using `getKeyboardAvoidingViewTopDistance` when this confirmation appears in the bottom of the screen?
2. Why `+ 30`?

src/screens/PaymentRequestDetail.js
```diff
@@ -62,6 +69,18 @@ class PaymentRequestDetail extends React.Component {
     this.props.dispatch(clearInvoice());
   }
 
+  /**
+   * If the payment was not yet received, the user goes back to the "Select Amount" screen.
+   * Otherwise, goes directly to the Home screen to see the updated wallet.
+   */
+  onBackClick() {
+    if (this.state.paymentIsConfirmed) {
```
pedroferreira1, line 77: Why can't you use `if (this.props.payment !== null)` instead of creating a new state?

src/components/AmountTextInput.js
tuliomir: Yes, it was working before.
I couldn't find a decisive answer for why this stopped working, but it probably has something to do with the general changes to the [`focus` related functions from React Navigation 4 to 5](https://reactnavigation.org/docs/5.x/upgrading-from-4.x/#navigation-events).

src/components/HathorHeader.js
tuliomir: That's right: while this PR was already under review I came up with a better organization for this on 326.
This variable will be short lived on the `constants` file.

src/constants.js
tuliomir: The `utils.js` still uses it on the 326 color PR. It will be completely removed soon on the colors refactor:
- #327

src/components/ReceiveMyAddress.js
tuliomir: As mentioned above, the method [`getNextAddress()`](https://github.com/HathorNetwork/hathor-wallet-lib/blob/bfc290175b8ea113f8e5e624690d3aa15e2b6da8/src/new/wallet.js#L563-L565) not only fetches the next available address, but also marks the current one as used.
```ts
  async getNextAddress() {
    // First we mark the current address as used, then return the next
    await this.getCurrentAddress({ markAsUsed: true });
```

 That was the reason the QA #318 brought this issue up on item 9. By changing the call to `getCurrentAddress()` we don't do this marking and but otherwise keep the current behavior intact.

src/utils.js
tuliomir: I couldn't pinpoint the reason for this, but there are two major suspects for this need:
- The change on the React Navigation from 4 to 6
- The removal of the iOS modal wrapper, since it's now quite different from what it used to be on older versions.

The lack of this added height was causing the buttons to hide under the keyboard whenever it was opened.

src/components/ModalConfirmation.js
tuliomir: 1. Because the bottom offset needed for the modal is dependent on the same elements as the keyboard avoiding views: the status bar and the header height. The code could be reused on both situations.

Maybe it would be better to refactor this method's name to something like `getBottomSafeInset()`?

2. That was trial and error. Probably the same method used to the previous static value of `16`. Here is how it is shown after this code, on the iPhone 13 mini simulator:
Modal bottom offset

src/components/AmountTextInput.js
tuliomir: A few considerations about this:

- The `AmountTextInput` that needs to be focused here is mounted when the `My Address` tab is focused, and because of that is not available to be focused at mount time.
- The tab management component currently used does not use the React Navigation [`material-top-tab-navigator`](https://reactnavigation.org/docs/material-top-tab-navigator) wrapper, but instead uses directly its underlying [`react-native-tab-view`](https://reactnavigation.org/docs/tab-view/) lib.

That way, the `focus` event does not work as expected within the screen after the migration from Navigator 4 to 5 and 6. It only receives the `focus` event when changing from the _Send_ stack to the _Receive_ stack, for example.

I would suggest opening another PR to use the `material-top-tab-navigator` on this _Receive_ stack instead of the _tab view_, so that we could have the `focus` events available on the screens again through React _hooks_ and remove the need for this `setTimeout`.

src/components/ModalConfirmation.js
tuliomir: Managed to find a better solution by removing the numeric inputs on the modal height, and replacing them with `100%` calculations. Done on [d610eb4](https://github.com/HathorNetwork/hathor-wallet-mobile/pull/325/commits/d610eb4aeec81fa5d2c21ce469025598712db351).

src/components/AmountTextInput.js
pedroferreira1: Ok. Create another issue to handle this and add it to the upgrade dependencies project, we can do it later

andreabadesso (approved):

PR #332, opened 2023-09-19
fix: jest integration

alexruzenhack:
### Acceptance Criteria
- upgrade jest to v29.2.1
- introduce typescript for dev environment
- change test to .tsx
- mock saga-redux and redux to avoid integration with the underlying storage

### Security Checklist


__tests__/App.test.tsx
```diff
@@ -1,12 +1,19 @@
 /* eslint-disable */
-
 import 'react-native';
+import {it, jest} from '@jest/globals';
```
andreabadesso, line 3:
```suggestion
import { it, jest } from '@jest/globals';
```

alexruzenhack:
> Package-lock change is huge, can you please check the nodejs and npm version you're using?

We already use the [lockfileVersion: 2](https://github.com/HathorNetwork/hathor-wallet-mobile/blob/a4834fca2b5a40b10b515d37a9ae340c8339cc13/package-lock.json#L4C11-L4C11). Jest caused all this change. It is very aggressive, indeed.

package.json
```diff
@@ -18,6 +18,7 @@
     "@fortawesome/free-solid-svg-icons": "^6.1.1",
     "@fortawesome/react-native-fontawesome": "^0.2.7",
     "@hathor/wallet-lib": "^1.0.1",
+    "@jest/globals": "^29.7.0",
```
r4mmer, line 21: This should be a dev-dependency, right?

jest.config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  preset: 'react-native',
+  transformIgnorePatterns: [
+    '<rootDir>/node_modules/(?!(@react-native|react-native|@notifee/react-native|react-native-device-info|react-native-gesture-handler|react-native-keychain|@react-navigation|react-native-status-bar-height|@react-native-firebase|@sentry/react-native|react-native-version-number|unleash-proxy-client|react-native-permissions|react-native-modal|react-native-animatable|react-native-camera-kit|@fortawesome/react-native-fontawesome|react-native-qrcode-svg)/)'
+  ],
+  setupFiles: [
+    '<rootDir>/jestMockSetup.js',
+  ],
+  cacheDirectory: '.jest/cache'
+};
```
r4mmer, line 10: Do you think we should move this to the package.json file? [reference](https://jest-bot.github.io/jest/docs/configuration.html)

r4mmer (changes requested):
Since we are updating the package-lock should we add some specifications on the minimum node and npm versions?
There is a way to do this on the package.json ([reference](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#engines)) and we can also use the .nvmrc file.

package.json
alexruzenhack: I checkout the commit `@8120e600` and verified that @jest/globals is a dependency for jest-runtime, which is already a dependency for jest:

```sh
❯ npm ls \@jest/globals
HathorMobile@0.26.0-rc6 /Users/alexrusenhack/repos/hathor/hathor-wallet-mobile
└─┬ jest@29.2.1
  └─┬ @jest/core@29.7.0
    └─┬ jest-runtime@29.7.0
      └── @jest/globals@29.7.0
```

I'm removing this commit 591c6af71425e0e6a9cbf90bc0bdf2fcb95b8021

jest.config.js
alexruzenhack: Moved config to package.json at: 4ae3816c6918dde0bbfe2d92947e16e2fe6c8923

jestMockSetup.js
```diff
@@ -0,0 +1,106 @@
+import { jest } from '@jest/globals';
```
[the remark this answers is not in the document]
alexruzenhack: Oh. Ok!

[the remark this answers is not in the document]
alexruzenhack: Added at f8b97d376805f08ee546cfbaf009a2ba7f7b1aaa

r4mmer (approved):

Issue #328, opened 2023-09-13
[pre-settings] Add predefined network settings


alexruzenhack:
It should contain:
- testnet settings
- mainnet settings

Issue #330, opened 2023-09-13
[pre-settings] Implement the networkSettings saga


alexruzenhack:
Implements the networkSettings saga to handle network change events.

Issue #329, opened 2023-09-13
[pre-settings] Add default network setting


alexruzenhack:
It should define a default network pre-settings to be used in wallet initialization if no other pre-settings in selected:
- mainnet is the default pre-settings

PR #333, opened 2023-09-22
(1) feat: add network settings pre-settings and saga

alexruzenhack:
### Acceptance Criteria

- Refactor reducer by extracting initialization to reducer.init.js
- Refactor feature toggle actions by adding action types
- Add mainnet and testnet pre-settings
- Define a mainnet pre-settings as default in the initial state of networkSettings
- Implement network settings saga:
  - process a custom network request
  - persist a processed custom network
  - clean store when wallet reset
- Implement the test for the network settings saga
- Add a hook in the feature toggle update to disable wallet service and push notification in case the wallet service url is not present in the network settings state

Closes: #328 #329 #330

### Security Checklist


src/reducer.init.js
```diff
@@ -0,0 +1,15 @@
+import { applyMiddleware, createStore } from 'redux';
```
andreabadesso, line 1: _Comment unrelated to line number_

I like the idea of a init reducer file, but maybe a better organization would be to create a `reducers/` folder with a `index.js` with this file contents

Later on we can start organizing the reducers in different files instead of putting them all in a single `reducer.js` file

__tests__/sagas/networkSettings.test.ts
```diff
@@ -0,0 +1,260 @@
...
+            }
+            return next(effect);
+        }
+
+        const middleware = createSagaMiddleware({ effectMiddlewares: [effectMiddleware] });
+        const store = createStore(reducer, {}, applyMiddleware(middleware));
+
+        const task = middleware.run(defaultSaga);
+
+        Promise.resolve()
```
andreabadesso, line 48: Why do you need this first `Promise.resolve()`?

__tests__/sagas/networkSettings.test.ts
```diff
@@ -0,0 +1,260 @@
+import * as nsSaga from '../../src/sagas/networkSettings';
```
andreabadesso, line 1: Great job on adding tests for the sagas

src/sagas/networkSettings.js
```diff
@@ -0,0 +1,179 @@
...
+
+  // TODO: get network as the best effort
+  // - walletServiceUrl has precedence
+  // - nodeUrl as fallback
+
+  let network;
+  let walletServiceWsUrl;
+  if (walletServiceUrl) {
+    const _url = new URL(walletServiceUrl);
+    if (_url.protocol.includes('https')) {
```
andreabadesso, line 66: The websocket endpoint might have a different URL as it's a different lambda, so this might not work for everyone, specially during development (which is our current usecase, right?)

andreabadesso (changes requested):

src/reducer.init.js
alexruzenhack: Refactored at e34cf6d152ef61784c3786e9f6805bdf8fb5052c

__tests__/sagas/networkSettings.test.ts
alexruzenhack: Look at this article ["Understanding setImmediate()"](https://nodejs.dev/en/learn/understanding-setimmediate/). To cut short I use it just to resume the promises before the end of the call stack, before the next event loop, which is possible the execution of another test. Without that, we risk a memory leak or jest zombie(open) handler. See caution in the jest ["Async/Await"](https://jestjs.io/docs/asynchronous#asyncawait).

src/actions.js
```diff
@@ -101,20 +101,33 @@ export const types = {
...
   SET_WALLET_CONNECT_MODAL: 'SET_WALLET_CONNECT_MODAL',
   SET_WALLET_CONNECT_SESSIONS: 'SET_WALLET_CONNECT_SESSIONS',
   SET_UNLEASH_CLIENT: 'SET_UNLEASH_CLIENT',
   WC_URI_INPUTTED: 'WC_URI_INPUTTED',
   WC_CANCEL_SESSION: 'WC_CANCEL_SESSION',
   WC_SET_CONNECTION_FAILED: 'WC_SET_CONNECTION_FAILED',
+  // Network Settings actions
+  NETWORKSETTINGS_UPDATE: 'NETWORK_SETTINGS_UPDATE',
+  NETWORKSETTINGS_UPDATE_SUCCESS: 'NETWORK_SETTINGS_UPDATE_SUCCESS',
+  NETWORKSETTINGS_UPDATE_FAILURE: 'NETWORK_SETTINGS_UPDATE_FAILURE',
```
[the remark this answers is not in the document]
alexruzenhack: Actually I used a pattern here to frame the the names of actions in a taxonomy. See this commit message 6980cac80982b4437f57fc654920f27c38aa38eb. I can replicate it here. Basically we use a triple to communicate the `<domain>` or `<subject>`, then we inform the performed `<action>` in the domain, and we finish with the expected <result> of the action, being it an optional token in the action taxonomy.

```txt
I used the following taxonomy for the actions:
- <subject>_<action>[_<result>]

The `<subject>` and the `<result>` must be nouns and `<action>`
must be a verb.

For instance:

- NETWORKSETTINGS_UPDATE
- NETWORKSETTINGS_UPDATE_SUCCESS
- NETWORKSETTINGS_UPDATE_FAILURE
```

If we agree with this taxonomy, I will later refactor all the actions to match this rule. Otherwise I can follow your suggestion and change this to the current pattern.

I did it to try to bring more order to the actions names.

src/sagas/helpers.js
```diff
@@ -175,3 +176,47 @@ export async function getRegisteredTokens(wallet, excludeHTR = false) {
 
   return tokens;
 }
+
+export async function getFullnodeNetwork() {
+  // NOTE: Should we change the user-agent?
```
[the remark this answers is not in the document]
alexruzenhack: Added at 5a3d0e0de38253a85c2fd9225c8acc1184734050

src/sagas/networkSettings.js
```diff
@@ -0,0 +1,179 @@
...
+export function* updateNetworkSettings(action) {
+  const {
+    nodeUrl,
+    explorerUrl,
+    walletServiceUrl,
+  } = action.payload || {};
+
+  // validates input emptyness
+  if (isEmpty(action.payload)) {
+    yield put(networkSettingsUpdateFailure());
```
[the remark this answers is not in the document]
alexruzenhack: Sure. But I don't want to do it in this PR.

src/sagas/helpers.js
```diff
@@ -175,3 +176,47 @@ export async function getRegisteredTokens(wallet, excludeHTR = false) {
...
+  }
+}
+
+export function isWalletServiceUnavailable(networkSettings) {
+  return networkSettings.walletServiceUrl == null;
+}
+
+export function disableFeaturesIfNeeded(networkSettings, currentFeatureToggles) {
+  let featureToggles = { ...currentFeatureToggles };
+  if (isWalletServiceUnavailable(networkSettings)) {
```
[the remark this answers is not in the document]
alexruzenhack: I improved by adding docstrings at 5ba3e501c77cea9dc42389b122b5916bb19d2fe4

But cutting short, `networkSettings` is the new custom network settings and this function verifies if the wallet service URL is not there, if yes "it is not there" we need to forcefully disable the features that depends on the wallet-service.

This whole process is just to enforce the right feature toggles, because it is updated regularly by a polling process.

src/sagas/networkSettings.js
alexruzenhack: Implemented at e28fe89037a750a5958280be3099aded816220d3

src/sagas/networkSettings.js
```diff
@@ -0,0 +1,179 @@
...
+  });
+
+  if (timeout) {
+    yield put(reloadWalletRequested());
+  }
+}
+
+export function* cleanNetworkSettings() {
+  STORE.removeItem(networkSettingsKey.networkSettings);
+  yield 0;
```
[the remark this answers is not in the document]
alexruzenhack: Just to comply with lint:
```txt
/Users/alexrusenhack/repos/hathor/hathor-wallet-mobile/src/sagas/networkSettings.js
  175:8  error  This generator function does not have 'yield'  require-yield
```

src/actions.js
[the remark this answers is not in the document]
andreabadesso: I like it, but we should refactor the other ones

Maybe create an issue and also add a comment there referencing it, so we refactor it on the next time we change this file

src/sagas/networkSettings.js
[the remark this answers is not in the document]
andreabadesso: I think you can also use regular functions on redux-saga effects, can you try it?

```suggestion
export function cleanNetworkSettings() {
```

src/sagas/networkSettings.js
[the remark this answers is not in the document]
andreabadesso: Why?

This is just silently failing as is, right?

__tests__/sagas/networkSettings.test.ts
andreabadesso: Great, thanks

src/actions.js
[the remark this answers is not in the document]
alexruzenhack: Fair. I will do it.

src/sagas/networkSettings.js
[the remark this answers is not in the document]
alexruzenhack: Right! Next PR I will refactor wallet saga and then I will start the screens. In the screen's PRs I will be forced to specify the failure details to be handled as a feedback in the frontend. Does it make sense?

src/sagas/networkSettings.js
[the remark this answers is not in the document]
alexruzenhack: Oh. I didn't know... thanks!

src/actions.js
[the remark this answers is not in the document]
alexruzenhack: The issue: https://github.com/HathorNetwork/hathor-wallet-mobile/issues/334

src/sagas/networkSettings.js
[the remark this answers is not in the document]
alexruzenhack: The documentation says it should be a generator, [see](https://redux-saga.js.org/docs/api#takeeverypattern-saga-args). Can we left this unchanged? I will need to test this flow anyway latter, and I will have the opportunity to improve this.

andreabadesso (approved):

src/sagas/networkSettings.js
```diff
@@ -0,0 +1,186 @@
...
+    walletServiceUrl,
+    walletServiceWsUrl,
+    nodeUrl,
+    explorerUrl,
+  };
+
+  yield put(networkSettingsUpdateSuccess(customNetwork));
+}
+
+function invalidUrl(value) {
```
pedroferreira1, line 136: Missing docstring

src/sagas/networkSettings.js
```diff
@@ -0,0 +1,186 @@
...
+      return true;
+    }
+  } catch {
+    return true;
+  }
+
+  return false;
+}
+
+export function* persistNetworkSettings(action) {
```
pedroferreira1, line 154: Missing docstring

src/sagas/networkSettings.js
```diff
@@ -0,0 +1,186 @@
...
+  config.setExplorerServiceBaseUrl(explorerUrl);
+  config.setServerUrl(nodeUrl);
+
+  // - walletServiceUrl has precedence
+  // - nodeUrl as fallback
+  let network;
+  if (walletServiceUrl) {
+    const _url = new URL(walletServiceUrl);
+
+    if (!walletServiceWsUrl && _url.protocol.includes('https')) {
```
pedroferreira1, line 73: I think it's better to require the user to set the ws URL as well, instead of assuming it would be this format. What do you think?

package.json
```diff
@@ -18,6 +18,7 @@
     "@fortawesome/free-solid-svg-icons": "^6.1.1",
     "@fortawesome/react-native-fontawesome": "^0.2.7",
     "@hathor/wallet-lib": "^1.0.1",
+    "@jest/globals": "^29.7.0",
```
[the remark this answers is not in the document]
alexruzenhack: Oh... this will go to the dev after we rebase.

src/sagas/wallet.js
```diff
@@ -138,6 +138,9 @@ export function* startWallet(action) {
     dispatch = _dispatch;
   });
 
+  // Configure a custom user-agent in the axios calls
+  config.setUserAgent('wallet-mobile');
```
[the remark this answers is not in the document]
alexruzenhack: Humm... nice! I didn't know. I will remove it from here.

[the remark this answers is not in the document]
alexruzenhack: Removed at 3e0c5a11d689641ccd1dc9c69aed6d3dbb6f2e4c

src/sagas/networkSettings.js
```diff
@@ -0,0 +1,186 @@
...
+  // - should have a valid URL, if given
+  if (walletServiceUrl && invalidUrl(walletServiceUrl)) {
+    yield put(networkSettingsUpdateFailure());
+    return;
+  }
+
+  // validates walletServiceWsUrl
+  // - optional
+  // - should have a valid URL, if given
+  if (walletServiceWsUrl && invalidUrl(walletServiceWsUrl)) {
```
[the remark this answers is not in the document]
alexruzenhack: No. It didn't become required if walletServiceUrl is given. What happens is: if you inform walletServiceUrl, but didn't inform walletServiceWsUrl, we infer the later. However, if you don't inform the walletServiceUrl and inform the walletServiceWsUrl, we take the later as it is. We also take walletServiceWsUrl as is even if you also inform the walletServiceUrl.

src/sagas/networkSettings.js
```diff
@@ -0,0 +1,186 @@
...
+    yield put(networkSettingsUpdateFailure());
+    return;
+  }
+
+  // NOTE: Should we allow that all the URLs be equals?
+  // In practice they will never be equals.
+
+  const oldExplorerUrl = config.getExplorerServiceBaseUrl();
```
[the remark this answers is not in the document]
alexruzenhack: Oh! You right. I treated them as the same.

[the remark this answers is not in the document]
alexruzenhack: But actually I don't need both. I only need the URL from the frontend.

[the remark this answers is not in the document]
alexruzenhack: I confirmed here, the wallet didn't use the explorer service URL, so it is save not capture the explorer service URL, only the explorer URL.

src/sagas/networkSettings.js
alexruzenhack: Oh... We only infer the URL this way when they don't provides the WS URL, otherwise we use the URL from the input.

src/sagas/wallet.js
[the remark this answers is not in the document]
alexruzenhack: Removed at 3e0c5a11d689641ccd1dc9c69aed6d3dbb6f2e4c

src/sagas/networkSettings.js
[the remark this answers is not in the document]
alexruzenhack: Some refactor at 77d7ba2722015788bbc90a7dea429a06bfca3c59

src/sagas/networkSettings.js
alexruzenhack: Resolve at 995e2b78eff0653f6544e34d7bd14e6157843432

src/sagas/networkSettings.js
alexruzenhack: Resolve at 995e2b78eff0653f6544e34d7bd14e6157843432

package.json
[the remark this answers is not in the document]
pedroferreira1: But this PR (https://github.com/HathorNetwork/hathor-wallet-mobile/pull/332) does not update the package lock, only this one. We should update the package lock only after this rebase. Can you do it before merging?

src/sagas/networkSettings.js
pedroferreira1: Yes, I understand, but isn't it better to require the WS URL instead of inferring it?

src/sagas/networkSettings.js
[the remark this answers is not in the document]
pedroferreira1: Yes, we use the explorer service URL for the metadata API. You should have both URLs in the network settings

package.json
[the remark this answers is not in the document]
alexruzenhack: Oh! It don't change now because I created another PR only to update the package-lock.json, see https://github.com/HathorNetwork/hathor-wallet-mobile/pull/335

src/sagas/networkSettings.js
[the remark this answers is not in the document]
alexruzenhack: I saw this metadata API, but I couldn't identify when it is used in the wallet-mobile. Does it is really required?

src/sagas/networkSettings.js
alexruzenhack: Making it required? Can be, for me, it's ok and cheap to introduce. My decision drive was to put the least possible barriers to not bother too much the dev.

alexruzenhack: Changed at 562ee965ce5f0c68e0810fc16b81453893eece5c:
- explorerServiceUrl is required
- walletServiceWsUrl is conditionally required

package.json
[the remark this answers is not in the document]
alexruzenhack: It is resolved with the rebase.

src/actions.js
```diff
@@ -830,3 +848,44 @@ export const setWCConnectionFailed = (failed) => ({
   type: types.WC_SET_CONNECTION_FAILED,
   payload: failed,
 });
+
+/**
+ * Request the custom network settings input to be processed.
+ * @param {{
+ *   nodeUrl: string,
```
pedroferreira1, line 855: Is it missing `stage` and `network`?

src/constants.js
```diff
@@ -163,3 +163,50 @@ export const FEATURE_TOGGLE_DEFAULTS = {
...
+export const NODE_SERVER_MAINNET_URL = 'https://mobile.wallet.hathor.network/v1a/';
+export const EXPLORER_MAINNET_URL = 'https://explorer.hathor.network/';
+
+export const PRE_SETTINGS_MAINNET = {
+  stage: STAGE,
+  network: NETWORK,
+  walletServiceUrl: WALLET_SERVICE_MAINNET_BASE_URL,
+  walletServiceWsUrl: WALLET_SERVICE_MAINNET_BASE_WS_URL,
+  nodeUrl: NODE_SERVER_MAINNET_URL,
+  explorerUrl: EXPLORER_MAINNET_URL,
```
pedroferreira1, line 193: Where's the explorer service url for both mainnet and testnet?

src/sagas/networkSettings.js
```diff
@@ -0,0 +1,233 @@
...
+
+  return false;
+}
+
+/**
+ * Saves the network custom settings in the application storage.
+ * @param {object} action contains the payload with the network settings
+ * already saved in the redux store.
+ */
+export function* persistNetworkSettings(action) {
```
pedroferreira1, line 195: Why don't you update the redux state as well here? We would have a different information in the redux and in the storage?

src/actions.js
alexruzenhack: No. It is inferred automatically.

src/sagas/networkSettings.js
alexruzenhack: No. I separated both to distinguish the input treatment from the storing process. Between the update handover to the persist function, the redux stores the value in the state. That's when we have the opportunity to change the interface UI almost immediately. But of course, this can be changed, it is just a preference to segregate the processes.

andreabadesso (approved):

src/constants.js
```diff
@@ -163,3 +163,50 @@ export const FEATURE_TOGGLE_DEFAULTS = {
...
+};
+
+/**
+ * Network settings key collection to be used with the app
+ * storage.
+ */
+export const networkSettingsKey = {
```
andreabadesso, line 200: What do you think about renaming this to `networkSettingsKeyMap`?

`networkSettingsKey` implies imo that there is only one key, which is the case now, but might not be true in the future

andreabadesso (changes requested):

pedroferreira1 (approved):

src/constants.js
alexruzenhack: I used `networkSettingsKey` just keep consistency with [`unleashStorageKey`](https://github.com/HathorNetwork/hathor-wallet-mobile/blob/562ee965ce5f0c68e0810fc16b81453893eece5c/src/constants.js#L90) and [`pushNotificationKey`](https://github.com/HathorNetwork/hathor-wallet-mobile/blob/562ee965ce5f0c68e0810fc16b81453893eece5c/src/constants.js#L106). Can you pass this time and I open a new PR just to "fix" all the cases together?

alexruzenhack: I changed the name in this commit on PR2: c9c81e4559d8b63ede0b165a60e5af399809f613

alexruzenhack:
I made a rebase with `dev` branch to resolve the conflicts.

package-lock.json
```diff
@@ -5524,6 +5524,12 @@
       "integrity": "sha512-dRLjCWHYg4oaA77cxO64oO+7JwCwnIzkZPdrrC71jQmQtlhM556pwKo5bUzqvZndkVbeFLIIi+9TC40JNF5hNQ==",
       "dev": true
     },
+    "node_modules/@types/lodash": {
```
pedroferreira1, line 5527: Why do you have package-lock changes without package changes?

alexruzenhack: - Removed: 6f4005a621b4259770a6dbaa14b1716c258d87d1

src/constants.js
andreabadesso: Thanks!

PR #337, opened 2023-10-01
(3) refactor: hardcoded explorer link to networkSettings explorerUrl

alexruzenhack:
### Acceptance Criteria
- Use the networkSettings.explorerUrl as the link to the explorer button.

### Security Checklist

### DIFF
--- a/src/components/PublicExplorerListButton.js
+++ b/src/components/PublicExplorerListButton.js
@@ -5,13 +5,15 @@ import icShareActive from '../assets/icons/icShareActive.png';
 import { ListButton } from './HathorList';
 import { COLORS } from '../styles/themes';
 import { useSelector } from 'react-redux';
+import { combineURLs } from '../utils';
 
 export function PublicExplorerListButton(props) {
   const { txId } = props;
   const explorerIcon = <Image source={icShareActive} width={24} height={24} />;
-  const explorerUrl = useSelector((state) => state.networkSettings.explorerUrl);
+  const baseExplorerUrl = useSelector((state) => state.networkSettings.explorerUrl);
+  const txUrl = `transaction/${txId}`;
   // XXX: maybe we should have this on the constants or utils to check the network
-  const explorerLink = `${explorerUrl}transaction/${txId}`;
+  const explorerLink = combineURLs(baseExplorerUrl, txUrl);
 
   return (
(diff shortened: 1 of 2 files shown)


src/components/PublicExplorerListButton.js
```diff
@@ -4,12 +4,14 @@ import { Image, Linking } from 'react-native';
...
 import { COLORS } from '../styles/themes';
+import { useSelector } from 'react-redux';
 
 export function PublicExplorerListButton(props) {
   const { txId } = props;
   const explorerIcon = <Image source={icShareActive} width={24} height={24} />;
+  const explorerUrl = useSelector((state) => state.networkSettings.explorerUrl);
   // XXX: maybe we should have this on the constants or utils to check the network
-  const explorerLink = `https://explorer.hathor.network/transaction/${txId}`;
+  const explorerLink = `${explorerUrl}transaction/${txId}`;
```
pedroferreira1, line 14: I think we could use 

```
new URL(`transaction/${txId}`, `${explorerUrl}`).href
```

This way we will prevent any errors if the user sets the explorer URL without the last `/`.

What do you think?

alexruzenhack: I realize that URL is provided by Node, but it is not provided by Hermes, the ReactNative engine. We have two options, either verify the `/` in the end, or install a polyfill lib. I prefer the first option, to verify and adapt.

andreabadesso: I'm ok with both options

andreabadesso (approved):

src/components/PublicExplorerListButton.js
pedroferreira1: This is sad :(

But I prefer we avoid installing new packages just because of that. I saw that axios has a helper to do it but we don't have this in the mobile, only in the lib.

You can create a method exactly the same for our code and use it: https://github.com/axios/axios/blob/fe7d09bb08fa1c0e414956b7fc760c80459b0a43/lib/helpers/combineURLs.js

alexruzenhack: - added combineURLs method: 8b8b716180d0f19043f19ab49debb10a0f91dbcd
- refactored URLs combination: f7fd7aecb0657269e315f71a38a8d449480cde7f

alexruzenhack:
I rebased with the previous PR to resolve conflicts.

pedroferreira1 (approved):

andreabadesso (approved):

PR #354, opened 2023-10-11
refactor: remove blank on pin screen

andreabadesso:
## Motivation

Currently we are displaying a blank screen while the faceid is being asked and only displaying the PIN when it's successful or cancelled, the expected behavior is to display the face id indicator on top of the pin screen

### Acceptance Criteria
- Remove the logic that hides the PIN while the faceid is being requested

### Security Checklist

### DIFF
--- a/src/screens/PinScreen.js
+++ b/src/screens/PinScreen.js
@@ -61,7 +61,6 @@ class PinScreen extends React.Component {
       pin: '',
       pinColor: COLORS.textColor,
       error: null,
-      waitingForBiometry: false,
     };
 
     this.canCancel = false;
@@ -112,8 +111,6 @@ class PinScreen extends React.Component {
   handleBackButton = () => true;
 
   askBiometricId = async () => {
-    // Displaying an empty screen while waiting for the OS biometry response
-    this.setState({ waitingForBiometry: true });
     try {
       /* getGenericPassword will return either `false` or UserCredentials:
        *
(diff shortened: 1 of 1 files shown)


tuliomir:
Upon investigating this issue on Android, it was found that the screen being rendered at the time the biometry is requested is not the `PinScreen`, as it happens with iOS.

On Android the screen being displayed is the `Decide` from the `RootStack` navigator

https://github.com/HathorNetwork/hathor-wallet-mobile/blob/a4834fca2b5a40b10b515d37a9ae340c8339cc13/src/App.js#L708-L709

This happens before the `AppStack` is even rendered and `PinScreen` can be displayed. The `getBiometry()` call is then executed at `componentDidMount` time for the AppStack

https://github.com/HathorNetwork/hathor-wallet-mobile/blob/a4834fca2b5a40b10b515d37a9ae340c8339cc13/src/App.js#L480-L481

From there on, the order of the rendering appears to be different. What follows are videos displaying the exhibition on Android 11, Pixel 4 emulated phone on the current `dev` version ( v0.26.0-rc7 ) and the current Google App version ( v0.25.0 ).

**v0.26.0-rc7** - The `Decide` screen was painted blue ( `#55F` ) for didactic visualization:

https://github.com/HathorNetwork/hathor-wallet-mobile/assets/1299409/f7ef8496-9d97-4aaa-b4b1-f717710ee83d

**v0.25.0** - Original colors, current user experience

https://github.com/HathorNetwork/hathor-wallet-mobile/assets/1299409/0214e45f-9a36-4d51-a792-1a5609429382

tuliomir (approved):
Since this blank screen behavior on Android is not new to this version, and would require a greater refactor to be fixed, I consider the scope of this PR done.

pedroferreira1 (approved):

PR #380, opened 2023-11-06
Fix: Pin not working on update

tuliomir:
In a specific update condition, the app would not accept the user's PIN code:
- The user should have a wallet on and already installed `v0.25.0` app
- This wallet should have no registered custom tokens, only HTR
- The user upgrades to `v0.26.0-rc7`
- The Pin Screen does not accept the valid Pin anymore

This would happen because the migration code expected registered custom tokens to be present. This PR fixes this.

### Acceptance Criteria
- The app should accept the user's old PIN after migration on the identified edge case

### Security Checklist

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -370,11 +370,7 @@ class AsyncStorageStore {
    * @async
    */
   async handleMigrationOldRegisteredTokens() {
-    const oldTokens = this.getItem('wallet:tokens');
-    if (!oldTokens) {
-      // There were no registered custom tokens: skipping token migration.
-      return;
-    }
+    const oldTokens = this.getItem('wallet:tokens') ?? [];
 
     const newTokens = {};
     for (const token of oldTokens) {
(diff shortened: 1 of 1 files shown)


alexruzenhack (approved):

src/store.js
```diff
@@ -371,6 +371,11 @@ class AsyncStorageStore {
    */
   async handleMigrationOldRegisteredTokens() {
     const oldTokens = this.getItem('wallet:tokens');
+    if (!oldTokens) {
```
pedroferreira1, line 374: Is there any place in the app that we expect this key (`REGISTERED_TOKENS_KEY `) to exist in the storage?

Instead of doing this, why don't we just do this?

```
const oldTokens = this.getItem('wallet:tokens') ?? [];
```

So the code will continue working and will save an empty object in the storage in case there are no custom tokens.

tuliomir: Nowhere in the code we expect this key to exist: all `getItem` calls have a protection in place.

That said, I agree with your suggestion: it's more dev friendly to always have the key in storage. Refactored on [e3668d4](https://github.com/HathorNetwork/hathor-wallet-mobile/pull/380/commits/e3668d4737c70033854aa9d134bd37ec7dbc22e6)

pedroferreira1 (approved):

PR #387, opened 2023-11-14
Docs: QA steps to validate QR Code scanner

tuliomir:
Recently we found some errors ( fixed by #384 and #386 ) that only happened during QR Code scanning. This procedure should also be on the main QA.

### Acceptance Criteria
- There should be steps to validate the functionality of the QR Code Scanner

### Security Checklist

### DIFF
--- a/QA.md
+++ b/QA.md
@@ -83,6 +83,13 @@
     1. Generate a payment request of 2 TEST.
     1. Pay the Payment Request using your test wallet that has just received the TEST before and check the confirmation message.
 
+1. **Send HTR with QR Code**
+    1. Go to the Receive Screen and click on Payment Request.
+    1. Generate a payment request of 1 HTR
+    1. Take a screenshot of the QR Code and open that screenshot in another device
+    1. Use the Mobile Wallet QR code scanner to read this screenshot
+    1. Pay the Payment Request using the same wallet used to generate it
+
 1. **Register a token Tests**
     1. Go to the Dashboard Screen.
     1. Select the Test Token.
--- a/QA.md
+++ b/QA.md
@@ -84,11 +84,9 @@
     1. Pay the Payment Request using your test wallet that has just received the TEST before and check the confirmation message.
 
 1. **Send HTR with QR Code**
-    1. Go to the Receive Screen and click on Payment Request.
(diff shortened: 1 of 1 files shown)


QA.md
```diff
@@ -83,6 +83,13 @@
     1. Generate a payment request of 2 TEST.
     1. Pay the Payment Request using your test wallet that has just received the TEST before and check the confirmation message.
 
+1. **Send HTR with QR Code**
+    1. Go to the Receive Screen and click on Payment Request.
+    1. Generate a payment request of 1 HTR
+    1. Take a screenshot of the QR Code and open that screenshot in another device
+    1. Use the Mobile Wallet QR code scanner to read this screenshot
+    1. Pay the Payment Request using the same wallet used to generate it
```
andreabadesso, line 91: This sounds very confusing to me

Since we are validating the QR Code scan (the QR Code itself is already being tested in another step), we can just change it to something like:

```suggestion
    1. Generate a payment request requesting a payment of 1 HTR from another device
    1. Use the Mobile Wallet QR code scanner to read this screenshot
    1. Pay the Payment Request, the transaction should be successful
```

andreabadesso (changes requested):

QA.md
```diff
@@ -83,6 +83,11 @@
     1. Generate a payment request of 2 TEST.
     1. Pay the Payment Request using your test wallet that has just received the TEST before and check the confirmation message.
 
+1. **Send HTR with QR Code**
+    1. Generate a payment request requesting 1 HTR from another device
```
pedroferreira1, line 87: Does it make sense to generate a payment request of 1,000,000 HTR, so we can have an error handling of no funds? Then create this normal payment request of 1 and continue as you detailed?

tuliomir: For the current QR Code reading flow, the lack of funds is only validated when the actual `tx` is being built, because no utxos are found to complete its inputs. ( see screenshot below )


Maybe it would be better to add this step when we introduce a dedicated validation on the interface, such as the one we have with the manual inputs. What do you think?

No funds Screenshot

pedroferreira1 (approved):

andreabadesso (approved):

PR #390, opened 2023-11-21
feat: add txMiningServiceUrl to network settings

alexruzenhack:
### Acceptance Criteria
This change sets the txMiningUrl in the wallet lib configurations and a
presettings or custom network is selected.

- Add Tx Mining Service URL field in the form of Custom Network view
- Add TX_MINING_SERVICE_\<NETWORK\>_URL constants
- Add txMiningServiceUrl in presettings and in the redux networkSettings

https://github.com/HathorNetwork/hathor-wallet-mobile/assets/5992210/4d3b6ebf-80b4-445a-b141-ed9164b5599c

### Security Checklist

### DIFF
--- a/src/screens/NetworkSettings/CustomNetworkSettingsScreen.js
+++ b/src/screens/NetworkSettings/CustomNetworkSettingsScreen.js
@@ -226,7 +226,7 @@ export const CustomNetworkSettingsScreen = ({ navigation }) => {
 
         <SimpleInput
           containerStyle={styles.input}
-          label={t`Tx Mining Service URL`}
+          label={t`Transaction Mining Service URL`}
           autoFocus
           onChangeText={handleInputChange('txMiningServiceUrl')}
           error={invalidModel.txMiningServiceUrl}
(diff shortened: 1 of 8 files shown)


andreabadesso (approved):

src/screens/NetworkSettings/CustomNetworkSettingsScreen.js
```diff
@@ -217,6 +224,15 @@ export const CustomNetworkSettingsScreen = ({ navigation }) => {
           value={formModel.explorerServiceUrl}
         />
 
+        <SimpleInput
+          containerStyle={styles.input}
+          label={t`Tx Mining Service URL`}
```
pedroferreira1, line 229: I think I prefer the label as 'Transaction Mining Service URL'. We don't use the contraction 'tx' anywhere in the app, so to keep consistency I think we should continue using 'Transaction' here.

What do you think?

alexruzenhack: I agree! It is clearer, despite `Tx` be ubiquitous in the industry.

alexruzenhack: Done at: https://github.com/HathorNetwork/hathor-wallet-mobile/pull/390/commits/4abc76376f843e901f160581ce8e36bdabf78898

PR #396, opened 2023-11-29
fix: network assignment when handling custom network update

alexruzenhack:
### Motivation
The lib only accepts to values `mainnet`, `testnet` and `privatenet` to the `network` wallet param. However, a network name may come with a suffix, like the current "testnet" network, which name is `testnet-golf`. By assigning `testnet-golf` to `network` in the wallet makes the lib invalidates the configuration because `testnet-golf` is not equal `testnet`.

### Solution
Assign to `network` only the values the lib allows, by checking the network names. If a network name doesn't match with `mainnet`, `testnet` or `privatenet`, then throw an error. Avoid to call to get version information from wallet-service when it is disabled.

### Acceptance Criteria
- Assign the right network value: `mainnet`, `testnet`, `privatenet` to `network` wallet param or throw an error.
- Skip get version information from wallet-service when it is disabled.
- Show Wallet Service URLs only if wallet-service-mobile.rollout feature flag is enabled for the wallet
- Remove `dev` prefix in the wallet-service URLs

### Caveat
The custom network update happens in 2 phases:
1. Check the input and update the wallet configuration with the new settings
2. Restart the wallet to download the new history from the custom network

In step (2) we first try to get the history from wallet-service if it is available, otherwise we try to get the history from fullnode.

In step (1) we first try to get the network version information from wallet-service if its URL is available in the custom network input, otherwise we try to get the network version information from fullnode.

These steps seems right, but there is a flaw: the wallet-service was supposed to be off, however it is only disabled for the wallet-mobile. If we use the "testnet presettings" from custom network screen, the algorithm of step (1) calls to get version information from wallet-service because there is a wallet-service URL, and receive the `testnet` value; in the step (2) it falls back to load transaction history from fullnode, therefore all seems right. However, if we get rid of the wallet-service URL from custom network input, the step (1) falls back to call to get version from fullnode and receives the value `testnet-golf`, which is invalidate by the lib.

### Conditional Wallet Service URLs inputs
Show the inputs if the wallet-service-mobile.rollout feature flag is enabled for the wallet.

https://github.com/HathorNetwork/hathor-wallet-mobile/assets/5992210/3258b51a-1417-45f8-9cd9-d0d8a4ffc8ae

### Security Checklist


src/sagas/networkSettings.js
```diff
@@ -151,8 +154,9 @@ export function* updateNetworkSettings(action) {
 
   // - walletServiceUrl has precedence
   // - nodeUrl as fallback
+  let potentialNetwork;
   let network;
-  if (walletServiceUrl) {
+  if (walletServiceUrl && useWalletService) {
```
andreabadesso, line 159: _Comment unrelated to line number_

What if the user does not have the wallet-service feature flag active? We will allow him to change the URL in the dev settings but it won't do anything, as we will ignore it

Maybe we should hide the wallet-service URLs in the dev settings screen if he's not in the wallet-service

andreabadesso (changes requested):

src/sagas/networkSettings.js
alexruzenhack: You right! I changed the input to conditional rendering: 9d25ef9207363b3bac88a75737b20fbf6e6a331c

src/screens/NetworkSettings/CustomNetworkSettingsScreen.js
```diff
@@ -233,23 +236,27 @@ export const CustomNetworkSettingsScreen = ({ navigation }) => {
...
+        {walletServiceEnabled && (
+          <SimpleInput
+            containerStyle={styles.input}
+            label={t`Wallet Service WS URL (optional)`}
+            autoFocus
+            onChangeText={handleInputChange('walletServiceWsUrl')}
+            error={invalidModel.walletServiceWsUrl}
+            value={formModel.walletServiceWsUrl}
+          />
+        )}
```
andreabadesso, line 259:
```suggestion
        {walletServiceEnabled && (
        <>
          <SimpleInput
            containerStyle={styles.input}
            label={t`Wallet Service URL (optional)`}
            autoFocus
            onChangeText={handleInputChange('walletServiceUrl')}
            error={invalidModel.walletServiceUrl}
            value={formModel.walletServiceUrl}
          />
          <SimpleInput
            containerStyle={styles.input}
            label={t`Wallet Service WS URL (optional)`}
            autoFocus
            onChangeText={handleInputChange('walletServiceWsUrl')}
            error={invalidModel.walletServiceWsUrl}
            value={formModel.walletServiceWsUrl}
          />
        </>
        )}
```

andreabadesso (approved):

pedroferreira1 (approved):

PR #400, opened 2023-12-01
fix: firebase initialization

alexruzenhack:
The changes in this PR solves all the problems related to the Firebase initialization. In the first place, it was caused by the update of FirebaseCoreInternal from v10.8.0 to v10.18.0. There was introduced some changes related the authorization module, which introduced the behavior observed by us and documented in this issue [RNFB#4428](https://github.com/invertase/react-native-firebase/issues/4428).

In summary, the problem happens on iOS devices, which must be registered on APNS prior the registration on FCM. This requirement was introduced in the Firebase implementation because an FCM token without the permissions from APNS is useless. Also, the APNS registration request can fail for some reason and it is the application responsibility to properly handle this situation, as we can see in the Apple documentation statement bellow.

>In addition to handling successful registrations with APNs, prepare your app to handle unsuccessful registrations [...]. Registration might fail if the user’s device isn’t connected to the network, if the APNs server is unreachable for any reason, or if the app doesn’t have the proper code-signing entitlement. When a failure occurs, set a flag and try to register again at a later time.
– [See the reference](https://developer.apple.com/documentation/usernotifications/registering_your_app_with_apns#2942135)

This statement was reinforced by the RNFB maintainer at [this comment](https://github.com/invertase/react-native-firebase/issues/6829#issuecomment-1431854924).

Also, simulators can't be registered on APNS. See this issue [RNFB#3174](https://github.com/invertase/react-native-firebase/issues/3174). This is a limitation imposed by Apple, and there are other methods to test push notification on iOS, take [this article](https://www.avanderlee.com/workflow/testing-push-notifications-ios-simulator/) as a reference.

### Acceptance Criteria
- Sets firebase on AppDelegate.mm
  - This configuration is important to let Firebase framework uses the credentials during the application bootstrap.
- Changes minimal iphoneos target to 12
  - The Firebase framework updated its minimal target
- Migrates Firebase configuration from `Info.plist` to `firebase.json`
  - The React Native Firebase package provides this file to concentrate Firebase setups in one place and turn it more ingrained in the development workflow
- Updates the podfile.lock to reflect the updates in the npm packages
- Upgrades cocoapods from v1.12.0 to v1.13.0
  - This upgrade is not required, however it solves warnings in the `pod install` execution
- Improves the logs from push notification, both in saga and background scripts
- Adds the `deviceRegistered` flag
  - Avoids throw errors in the push notification initialization
  - Avoids proceed with device registration on wallet-service
  - Determine if the Push Notification option will show up to users in the Settings screen. It will appears if Push Notification is available for the device AND the device is successfully registered on FCM, and APNS when iOS. Otherwise, it will not appear.

>[!NOTE]
> The following error message related to the supported version of iphoneos 12.
>```txt
><root>/ios/Pods/Pods.xcodeproj: >warning: The iOS Simulator deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is >set to 11.0, but the range of supported deployment target versions is 12.0 to 17.0.99. (in >target 'FirebaseCore' from project 'Pods')
>```

### Security Checklist

### DIFF
--- a/src/logger.js
+++ b/src/logger.js
@@ -1,3 +1,19 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * Wraps the built-in console to add scope to messages.
+* @param {string} scope - prefix each message.
+*
+* @example
+* const log = logger('my-scope');
+* log.debug('a custom message');
+* // output: [my-scope] a custom message
+ */
 export const logger = (scope) => ({
   log(msg) {
     console.log(`[${scope}] ${msg}`);


tuliomir (approved):
:star: I like the idea of scoping the logs, especially with a context that has so many nuances such as push notifications

firebase.json
```diff
@@ -0,0 +1,7 @@
+{
+  "$schema": "./node_modules/@react-native-firebase/app/firebase-schema.json",
+  "react-native": {
+    "messaging_auto_init_enabled": true,
+    "messaging_ios_auto_register_for_remote_messages": true
```
pedroferreira1, line 5: You didn't have this configuration before, right? Why do we need this now?

ios/Podfile.lock
```diff
@@ -874,4 +874,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: c422aedd0c3b236578801088c8afef28030c3c83
 
-COCOAPODS: 1.13.0
+COCOAPODS: 1.12.0
```
pedroferreira1, line 877: Why downgrade?

src/logger.js
```diff
@@ -0,0 +1,11 @@
+export const logger = (scope) => ({
```
pedroferreira1, line 1: Missing file header

ios/Podfile.lock
alexruzenhack: Oh! My brain has bugged. I inverted the understanding here. I will double check.

src/sagas/pushNotification.js
```diff
@@ -125,15 +134,18 @@ function* confirmDeviceRegistrationOnFirebase(showErrorModal = true) {
   try {
     // Make sure deviceId is registered on the FCM
     if (!messaging().isDeviceRegisteredForRemoteMessages) {
-      console.debug('Device not registered on FCM. Registering device on FCM...');
+      log.debug('Device not registered on FCM. Registering device on FCM...');
       yield call(messaging().registerDeviceForRemoteMessages);
```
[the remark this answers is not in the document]
alexruzenhack: It checks for registration on FCM only. But it is required that the iOS devices must be registered on APNS first.

firebase.json
alexruzenhack: You right, we didn't have this explicit configuration before, but it was the default behavior of RNFB. Now we are enforcing this configuration because the RNFB implementation requires to register the iOS devices on APNS prior to register the device on FCM, and this causes a point of failure for the application to handle, that's why this configuration must be explicit now.

ios/Podfile.lock
alexruzenhack: I removed this update.

ios/Podfile.lock
```diff
@@ -515,7 +515,7 @@ PODS:
     - React-Core
   - ReactNativeExceptionHandler (2.10.10):
     - React-Core
-  - RNCAsyncStorage (1.19.3):
+  - RNCAsyncStorage (1.19.0):
```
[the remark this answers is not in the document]
alexruzenhack: I couldn't find an explanation for this downgrade. [This downgrade](https://github.com/firebase/firebase-ios-sdk/pull/11846) seems not related to this dependency.

pedroferreira1 (approved):

PR #420, opened 2024-01-22
feat: add SES to wallet-mobile

andreabadesso:
## Motivation

We want to defend against supply chain attacks as those kind of attacks have already hit the cryptocurrency ecosystem and present a significant risk for our developers and users of our wallets and apps.

This PR is part of a series of PRs using a set of tools called LavaMoat to improve security on our Javascript projects with a set of good defaults, preventing us from having to rewrite them from scratch

_Important:_ SES is not yet fully compatible with react-native, there is a compatibility tracker issue [here](https://github.com/LavaMoat/docs/issues/12)

What is not yet supported (that affect us)?
- [ ] Lockdown for react-native dependencies
- [ ] Building the app using [`lavamoat-node`](https://github.com/LavaMoat/LavaMoat/tree/main/packages/node)
- [ ] Hermes
- [ ] Policy generation
- [ ] allow-list for react-native dependencies

Also, to prevent multiple patches in react-native, we had to inject SES in a patch to react-native's `initializeCore` method, so it gets loaded before anything else in our bundle

### Acceptance Criteria
- We should be able to defend agains't prototype pollution and poisoning attacks coming from the installed libs and their dependencies by using [SES](https://github.com/endojs/endo/blob/master/packages/ses/README.md)
- We should have control over which dependencies are able to run `postinstall` scripts, blocking all the rest by using LavaMoat [allow-scripts](https://github.com/LavaMoat/LavaMoat/tree/main/packages/allow-scripts)
- We should use the hathor-unleash-client instead of the official unleash javascript client, which is a watered-down version of it, with only the features we use
- We should inject SES as soon as possible, so we should patch react-native to inject it before react-native initialization

### Security Checklist


package.json
```diff
@@ -131,5 +134,20 @@
       "<rootDir>/jestMockSetup.js"
     ],
     "cacheDirectory": ".jest/cache"
+  },
+  "lavamoat": {
+    "allowScripts": {
+      "$root$": true,
```
msbrogli, line 140: Why allow in root? Shouldn't we use false as default so we have to manually authorize scripts?

README.md
```diff
@@ -4,7 +4,13 @@ See the instructions to run on (Windows)[./WINDOWS.md].
 
 ## Install
 
-`npm install`
```
msbrogli, line 7: Any way to make `npm install` fail?

msbrogli (changes requested):

lockdown.umd.js
tuliomir, line 1: question: Should we have our Hathor file headers here, or is this generated automatically?

tuliomir (commented):
question: It looks good to me, but I feel there's something missing on the documentation on how to maintain the code from now on.

When we add new dependencies, should we change anything on the lavamoat files, or only on the "lavamoat" config inside the `package.json`?

thought: Since this will be a primary concern for any modifications on our dependencies, I think it's important to document it better on the readme file. Both for Hathor Labs devs and for the community members that wish to contribute.

andreabadesso:
> Hi. I think we could chat about what we're working on in LavaMoat. Get in touch if you think so too :)

Hey @naugtur, thanks for reaching out!

I've just sent you a DM on twitter, please tell me if this is the best way to contact you

README.md
```diff
@@ -4,7 +4,13 @@ See the instructions to run on (Windows)[./WINDOWS.md].
...
+`npm run setup`
+
+### Check SES integrity
+
+The SES lockdown file should match the commited sha256 sum
```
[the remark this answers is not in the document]
andreabadesso: Yes, we should update it when we update SES

Added a new markdown file explaining how to update it and why we do it in 9f42133a1f66b66e5e3db455597a546e90a537c5

lockdown.umd.js
andreabadesso: This is fetched from the SES project

I'm replicating what metamask is doing, which is to download it and commit it in the repository instead of installing it from npm

package.json
andreabadesso: The idea was to allow our `postinstall` script to run, as we need to patch our packages with nodejs-compatible shims

But allowscripts won't allow it, even with `$root$: true`, so I removed it.

README.md
andreabadesso: I don't think there is

`lavamoat/allow-scripts` blocks `preinstall` and `postinstall` scripts, so we can't make it fail

naugtur:
Update on our progress:
- we're making changes to SES lockdown that will let it work under Hermes. (Currently only async support is problematic)
- of you want runtime protections and can switch away from Hermes, any other engine can work. Current release of the webpack plugin is working for web targets. If you use the webpack based build tool from Callstack you could try out lavamoat plugin for webpack. We're planning to test.it ourselves after the work on SES above.

BTW 
Consider just switching to socketsupply.co

src/sagas/ses.js
```diff
@@ -0,0 +1,102 @@
...
+  take,
+  all,
+} from 'redux-saga/effects';
+import { checkForFeatureFlag } from './helpers';
+import { SES_FEATURE_TOGGLE, SHOULD_ENABLE_SES_STORAGE_KEY } from '../constants';
+
+const storage = new MMKV();
+
+function verifySesEnabled() {
+  return Object.isFrozen([].prototype);
```
tuliomir, line 22: suggestion: Explain the reason for this unusual syntax on this function's docstrings.

src/sagas/ses.js
```diff
@@ -0,0 +1,102 @@
...
+  if (restart) {
+    RNRestart.restart();
+  }
+}
+
+function enableSes(restart = true) {
+  storage.set(SHOULD_ENABLE_SES_STORAGE_KEY, true);
+
+  if (restart) {
+    RNRestart.restart();
```
tuliomir, line 38: question: Should we expect any negative side effects from a React Native restart? Any notes on how we should organize our code not to trigger those side effects by accident?

At a quick glance over the React Native Restart docs I found nothing noteworthy, but I think it's a question important enough to be brought up.

src/sagas/ses.js
```diff
@@ -0,0 +1,102 @@
...
+ *
+ */
+function* init() {
+  const storageEnabled = storage.getBoolean('should-enable-ses');
+  const unleashEnabled = yield call(checkForFeatureFlag, SES_FEATURE_TOGGLE);
+  const sesEnabled = verifySesEnabled();
+
+  if (unleashEnabled && storageEnabled && !sesEnabled) {
+    // This is an issue, the environment is not secure, we should issue a fatal
+    // error.
```
tuliomir, line 53: question(non-blocking): Why don't we throw here at the current code?

src/sagas/ses.js
```diff
@@ -0,0 +1,102 @@
...
+
+    if (oldSesEnabled && !newSesEnabled) {
+      // SES was disabled on the feature toggle provider
+      disableSes();
+    }
+
+    if (newSesEnabled && !oldSesEnabled) {
+      // SES was enabled on the feature-toggle provider, enable it without
+      // restarting the bundle
+      enableSes(false);
```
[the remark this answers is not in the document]
andreabadesso: The idea here was to avoid restarting the bundle if the user is in the middle of an operation, like sending a transaction or creating a token

src/sagas/ses.js
andreabadesso: Refactored the method and added a better jsDoc in ea15c73ebde51c0a19ec914d2b9519f581086820

src/sagas/ses.js
andreabadesso: I think there are negative side effects in usability, as if this happens while the user is doing something like creating a token or sending a transaction, the app will restart and he will lose his progress.. But I think this is acceptable 

We don't need to do this though, if we just remove this restart, SES will be enabled the next time he opens the app..

src/sagas/ses.js
andreabadesso: Sorry, I removed the error report from the code and forgot to re-add it

Fixed in 3c7a1286436cbea8a5e1ed5fde1ced0c0a6740c9

src/sagas/ses.js
```diff
@@ -0,0 +1,102 @@
...
+  if (unleashEnabled && storageEnabled && !sesEnabled) {
+    // This is an issue, the environment is not secure, we should issue a fatal
+    // error.
+  }
+
+  if (!unleashEnabled && storageEnabled) {
+    // SES was enabled, we should disable it in storage which gets read in the
+    // react-native initialization (more on this in patches/react-native+0.72.5.patch)
+    // and restart the react-native bundle.
+    disableSes();
```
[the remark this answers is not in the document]
tuliomir: I would suggest adding a comment or opening an issue/KTLO to address this in the future.

I agree that the QA would get this situation, but our QA is pretty big already and there is a possibility of human error.

In my opinion, this is not critical enough to demand a fix right now, but it's a nice to have for future improvements as a KTLO.

src/sagas/ses.js
```diff
@@ -0,0 +1,103 @@
...
+import { checkForFeatureFlag } from './helpers';
+import { SES_FEATURE_TOGGLE, SHOULD_ENABLE_SES_STORAGE_KEY } from '../constants';
+import { logger } from '../logger';
+import { onExceptionCaptured } from '../actions';
+import { verifySesEnabled } from '../utils';
+
+const storage = new MMKV();
+
+function disableSes(restart = true) {
+  logger.debug('Disabling SAS');
```
tuliomir, line 26:
```suggestion
  logger.debug('Disabling SES');
```

tuliomir (approved):

SES_AND_LAVAMOAT.md
```diff
@@ -0,0 +1,21 @@
...
+
+SES does not yet work with Hermes, so we had to disable it on both platforms https://github.com/facebook/hermes/issues/957
+
+### SES injection
+
+LavaMoat does not yet fully support react-native, there is a compatibility tracker [here](https://github.com/LavaMoat/docs/issues/12). Until then, we're using [SES](https://github.com/endojs/endo/tree/master/packages/ses) directly, injecting it directly into the InitializeCore which is the entrypoint for the react-native's bundle so it hardens our app as soon as possible in the process.
+
+This is done through a patch on the react-native package, using `patch-package`.
+
+We are currently using version `1.0.1` of SES, which is the latest version that works with react-native, it is loaded from the project root, `lockdown.umd.js`.
```
pedroferreira1, line 13: Is this still correct? It's been 6 months since you did it.

patches/react-native+0.72.5.patch
```diff
@@ -0,0 +1,52 @@
...
++if (isSesEnabled === undefined) {
++  isSesEnabled = true;
++  storage.set(SHOULD_ENABLE_SES_STORAGE_KEY, true);
++}
++
++const Platform = require('../Utilities/Platform');
++
++// This piece of code was taken from this PR: https://github.com/MetaMask/metamask-mobile/pull/6586/files
++// which is the reference implementation of SES On react-native
++if (isSesEnabled && Platform.OS === 'ios' && !global?.HermesInternal) {
```
pedroferreira1, line 26: Does it only work in iOS?

package.json
```diff
@@ -132,5 +138,20 @@
...
+      "@fortawesome/fontawesome-svg-core>@fortawesome/fontawesome-common-types": false,
+      "@fortawesome/free-regular-svg-icons": false,
+      "@fortawesome/free-regular-svg-icons>@fortawesome/fontawesome-common-types": false,
+      "@fortawesome/free-solid-svg-icons": false,
+      "@fortawesome/free-solid-svg-icons>@fortawesome/fontawesome-common-types": false,
+      "@hathor/wallet-lib>level>classic-level": false,
+      "@lavamoat/preinstall-always-fail": false,
+      "@sentry/react-native>@sentry/cli": false,
+      "ttag-cli": false
+    }
```
pedroferreira1, line 155: Why did you set this? Is this a test or these are real validations?

package.json
```diff
@@ -53,12 +56,14 @@
     "react-native-gesture-handler": "2.13.1",
     "react-native-keychain": "8.1.1",
     "react-native-localize": "2.2.6",
+    "react-native-mmkv": "2.12.2",
```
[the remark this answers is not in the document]
andreabadesso: I need mmkv to be able to synchronously read the flag to enable or disable SES. We use this so we can disable SES through unleash

package.json
```diff
@@ -53,12 +56,14 @@
...
     "react-native-modal": "13.0.1",
     "react-native-pager-view": "5.4.25",
     "react-native-permissions": "3.8.3",
     "react-native-qrcode-svg": "6.2.0",
     "react-native-randombytes": "3.6.1",
     "react-native-reanimated": "3.5.4",
+    "react-native-restart": "0.0.27",
```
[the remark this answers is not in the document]
andreabadesso: This is used to restart the bundle to reload it with/without SES (when we activate it through unleash)

src/sagas/ses.js
```diff
@@ -0,0 +1,103 @@
...
+  const sesEnabled = verifySesEnabled();
+
+  if (unleashEnabled && storageEnabled && !sesEnabled) {
+    // This is an issue, the environment is not secure, we should issue a fatal
+    // error.
+    yield put(onExceptionCaptured(new Error('SES should be enabled but environment is not secure, failing!'), true));
+  }
+
+  if (!unleashEnabled && storageEnabled) {
+    // SES was enabled, we should disable it in storage which gets read in the
```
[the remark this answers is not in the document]
andreabadesso: FIxed, thanks

src/sagas/ses.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+}
+
+/**
+ * `init` will run on the app initialization.
+ *
+ */
+function* init() {
+  const storageEnabled = storage.getBoolean('should-enable-ses');
+  const unleashEnabled = yield call(checkForFeatureFlag, SES_FEATURE_TOGGLE);
```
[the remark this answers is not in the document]
andreabadesso: Yes we should, thanks, done

patches/react-native+0.72.5.patch
andreabadesso: No, fixed

SES_AND_LAVAMOAT.md
andreabadesso: Updated this document with a better explanation on SES

package.json
andreabadesso: These are real validations, it will disallow those libraries (which are the ones with postinstall scripts) to execute the postinstall

pedroferreira1: Is there any configuration to `disallowAll`?

andreabadesso: It actually denies all scripts by default, this config was auto generated by using `allow-scripts auto`, I think it's nice to have so we know which dependencies had scripts.

Here is the output of `allow-scripts list`:

```
❯ npx allow-scripts list

# allowed packages with bin scripts
  (none)

# allowed packages with lifecycle scripts
  (none)

# disallowed packages with lifecycle scripts
- $root$ [1 location(s)]
- @fortawesome/fontawesome-svg-core [1 location(s)]
- @fortawesome/fontawesome-svg-core>@fortawesome/fontawesome-common-types [1 location(s)]
- @fortawesome/free-regular-svg-icons [1 location(s)]
- @fortawesome/free-regular-svg-icons>@fortawesome/fontawesome-common-types [1 location(s)]
- @fortawesome/free-solid-svg-icons [1 location(s)]
- @fortawesome/free-solid-svg-icons>@fortawesome/fontawesome-common-types [1 location(s)]
- @hathor/wallet-lib>level>classic-level [1 location(s)]
- @lavamoat/preinstall-always-fail [1 location(s)]
- @sentry/react-native>@sentry/cli [1 location(s)]
- ttag-cli [1 location(s)]
```

package.json
```diff
@@ -13,8 +13,10 @@
     "start:clean": "react-native start --reset-cache",
     "lint": "eslint .",
     "test": "jest",
+    "setup": "npm install && ./node_modules/.bin/allow-scripts && npm run postinstall",
     "locale-update-pot": "ttag extract -o ./locale/texts.pot ./src/",
-    "postinstall": "rn-nodeify --install stream,process,path,events,crypto,console,buffer --hack"
+    "preinstall": "exit 1",
```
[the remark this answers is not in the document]
andreabadesso: I don't, removed!

patches/intl+1.2.5.patch
```diff
@@ -0,0 +1,45 @@
...
+         for (var k in ls.Date) {
+-            if (hop.call(ls.Date, k)) defineProperty(Date.prototype, k, { writable: true, configurable: true, value: ls.Date[k] });
++            if (hop.call(ls.Date, k)) {
++                localeHelpers[`date_${k}`] = function(date) {
++                    return ls.Date[k].call(date);
++                };
++            }
+         }
++
++        return localeHelpers;
```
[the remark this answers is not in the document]
andreabadesso: Refactored this to just return `null`

src/sagas/ses.js
andreabadesso: Fixed! Thanks

pedroferreira1 (approved):

package.json
msbrogli: Ok. Thanks!

README.md
msbrogli: Great! Thanks!

Issue #415, opened 2024-01-12
Pin Screen taking time to confirm Push Notification registration


alexruzenhack:
The Pin Screen takes some time to resume:

https://github.com/HathorNetwork/hathor-wallet-mobile/assets/5992210/dba2bb19-4de9-414c-8f57-b3ee24634d2e

We need to debug this issue either to improve performance or to make it open the loading screen while processing.

Related to https://github.com/HathorNetwork/hathor-wallet-mobile/issues/404

tuliomir:
This behavior similar to one found on android on:
- #369 

Probably refactoring the `PinScreen` to a functional component would help rewrite the exhibition logic, but could have impact on the navigation of other screens.

PR #423, opened 2024-02-01
fix: push notification registration latency on PinScreen

alexruzenhack:
### Acceptance Criteria
- fix: PinScreen dismiss response
- fix: exits script after dispatch failure effect
- fix: replace BLOCKED to DENIED enum value
- fix: push notification authorization logic

PinScreen response latency now in miliseconds:

https://github.com/HathorNetwork/hathor-wallet-mobile/assets/5992210/97403ac8-0cf8-4cf3-b24a-d688483d5858

Closes: https://github.com/HathorNetwork/hathor-wallet-mobile/issues/415

---

# Performance Profiling

We have 2 major offenders:
- `generateAccessDataFromSeed` consuming ~4s
- `generateCreateWalletAuthData` consuming ~14s

Under `generateCreateWalletAuthData` we have other 3 offenders:
- `getXPrivKeyFromSeed` using ~4s
- `getXPubKeyFromSeed` using ~4s
- `getAuthXPubKeyFromSeed` using ~4s

Both `getXPubKeyFromSeed` and `getAuthXPubKeyFromSeed` calls `getXPrivKeyFromSeed` under the hood, which took almost all execution time. Therefore the real offender for this group is `getXPrivKeyFromSeed`.

image

## Wallet `start`

Both methods `generateAccessDataFromSeed` and `generateCreateWalletAuthData` are called from wallet `start`. We start an instance of `HathorWalletServiceWallet` when a user is operating a fullnode's wallet facade.

There are 4 options to try to solve or reduce the effect of this latency, either:

1. We implement a common wallet API to mount an authentication payload that can be used to call wallet-service API without the need to instantiate `HathorWalletServiceWallet`.
1. We store the `HathorWalletServiceWallet` instance once it is started for the first time, which can decrease the latency for subsequent interactions with push notification operations.
1. We remove support of push notification for users operating a fullnode's wallet facade and make it a feature strict to wallet-service users; which would allow us to remove this logic to instantiate a `HathorWalletServiceWallet`.
1. We implement a lightweight wallet just for the purpose to call wallet-service APIs.

We can discuss other possibilities, these ones were the ones I identified quickly without thinking too much.

>[!NOTE]
> PM me to get the profiler.

---

### Security Checklist

### DIFF
--- a/src/sagas/pushNotification.js
+++ b/src/sagas/pushNotification.js
@@ -378,6 +378,7 @@ export function* loadWallet() {
       });
     } catch (error) {
       yield put(pushLoadWalletFailed({ error }));
+      return;
     }
   } else {
     walletService = yield select((state) => state.wallet);
--- a/src/sagas/pushNotification.js
+++ b/src/sagas/pushNotification.js
@@ -362,7 +362,7 @@ export function* loadWallet() {
 
     // Delay 300ms to resume script execution in the next loop.
     // This solution liberates the PinScreen to dismiss.
-    yield race([delay(300)]);
+    yield delay(300);
 
     const seed = yield STORE.getWalletWords(pin);
     walletService = new HathorWalletServiceWallet({
--- a/src/sagas/pushNotification.js
+++ b/src/sagas/pushNotification.js
@@ -15,6 +15,7 @@ import {
   take,
   takeLatest,
   debounce,
+  delay,
 } from 'redux-saga/effects';
 import messaging from '@react-native-firebase/messaging';
 import notifee from '@notifee/react-native';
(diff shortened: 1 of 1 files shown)


andreabadesso (approved):

src/sagas/pushNotification.js
```diff
@@ -358,6 +359,11 @@ export function* loadWallet() {
     const network = new Network(networkSettings.network);
 
     const pin = yield call(showPinScreenForResult, dispatch);
+
+    // Delay 300ms to resume script execution in the next loop.
+    // This solution liberates the PinScreen to dismiss.
+    yield delay(300);
```
pedroferreira1, line 365: Does it work if you use `delay(0)`?

pedroferreira1 (approved):

src/sagas/pushNotification.js
alexruzenhack: Good point. I will test.

Issue #411, opened 2024-01-12
Reset ignore wallet service flag at some point


andreabadesso:
We should define the policy for reseting the flag

We should also display an error to the user to inform him that the wallet-service load failed and we will fall back to the fullnode

Related to #404

andreabadesso:
Fixed in https://github.com/HathorNetwork/hathor-wallet-mobile/pull/426

PR #426, opened 2024-02-05
refactor: reset wallet service ignore flag

andreabadesso:
Solves https://github.com/HathorNetwork/hathor-wallet-mobile/issues/411

### Acceptance Criteria
- We should refactor the `LoadHistoryScreen` to be a function
- We should display our standard retry screen when the wallet-service fails
- If the error thrown from `wallet.start()` is not a request error, we should display the modal to send the error to our sentry
- We should have a mechanism to expire the ignore wallet service flag
- We should fix an issue preventing errors from being sent to Sentry

### Security Checklist

### DIFF
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -201,7 +201,6 @@ export function* startWallet(action) {
   yield put(setUniqueDeviceId(uniqueDeviceId));
 
   try {
-    throw new Error('a new test error has appeared');
     yield call(wallet.start.bind(wallet), {
       pinCode: pin,
       password: pin,
@@ -216,7 +215,6 @@ export function* startWallet(action) {
     }
 
     if (useWalletService) {
-
       // Wallet Service start wallet will fail if the status returned from
       // the service is 'error' or if the start wallet request failed.
       //
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -61,6 +61,7 @@ import {
   setAvailablePushNotification,
   resetWalletSuccess,
   setTokens,
+  setShowWalletServiceLoadFailedModal,
 } from '../actions';
 import { fetchTokenData } from './tokens';
 import {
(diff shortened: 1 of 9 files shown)


src/sagas/wallet.js
```diff
@@ -194,11 +206,18 @@ export function* startWallet(action) {
     });
   } catch (e) {
     if (useWalletService) {
+      yield put(setShowWalletServiceLoadFailedModal(true));
+      yield take(types.WALLETSERVICE_RENDER_FAILED_ACK);
```
alexruzenhack, line 210: But if a user dismiss the modal without tapping on "ok" button? Should we consider it an acknowledge? Maybe the user wants to keep trying to use the wallet-service...

alexruzenhack (approved):

alexruzenhack (approved):

src/screens/LoadHistoryScreen.js
```diff
@@ -21,66 +21,58 @@ import Spinner from '../components/Spinner';
...
+        fontSize: 18, lineHeight: 22, width: 200, textAlign: 'center'
+      }}
+      >
+        There&apos;s been an error connecting to the server
+      </Text>
+      <SimpleButton
+        containerStyle={{ marginTop: 12 }}
+        textStyle={{ fontSize: 18 }}
+        onPress={() => dispatch(onWalletReload())}
+        title='Try again'
```
pedroferreira1, line 51: Missing translation tag

src/sagas/wallet.js
```diff
@@ -97,12 +100,22 @@ export function* isPushNotificationEnabled() {
  * @returns {Generator<unknown, boolean>}
  */
 export function* isWalletServiceEnabled() {
+  // Users might have had issues with the wallet-service in the past, we can detect
```
pedroferreira1, line 103: If the `IGNORE_WS_TOGGLE_FLAG` changed, how can we detect old flags with this code?

src/sagas/wallet.js
```diff
@@ -193,20 +206,28 @@ export function* startWallet(action) {
       password: pin,
     });
   } catch (e) {
+    // WalletRequestError can either be an network error making the request
```
pedroferreira1, line 209:
```suggestion
    // WalletRequestError can either be a network error making the request
```

andreabadesso: Fixed, thanks

src/screens/LoadHistoryScreen.js
andreabadesso: Fixed, thanks

src/sagas/wallet.js
andreabadesso: You are right, I shouldn't change the flag, fixed in 2e37c2cccc323d9b6239d835e723d80fa26dd040

pedroferreira1 (approved):

PR #441, opened 2024-03-12
feat: add nano contract feature toggle

alexruzenhack:
### Acceptance Criteria
- Should add Nano Contract feature toggle

### Security Checklist

### DIFF
--- a/src/reducers/reducer.js
+++ b/src/reducers/reducer.js
@@ -232,7 +232,7 @@ const initialState = {
   networkSettingsStatus: NETWORKSETTINGS_STATUS.READY,
   nanoContract: {
     /**
-     * @property {boolean} 
+     * @property {boolean}
      * @description Reflects Nano Contract's unleash feature toggle value.
      */
     available: false,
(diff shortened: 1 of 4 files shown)


src/reducers/reducer.js
```diff
@@ -230,6 +230,13 @@ const initialState = {
   networkSettings: PRE_SETTINGS_MAINNET,
   networkSettingsInvalid: {},
   networkSettingsStatus: NETWORKSETTINGS_STATUS.READY,
+  nanoContract: {
+    /**
+     * @property {boolean}
+     * @description Reflects Nano Contract's unleash feature toggle value.
+     */
+    available: false,
```
pedroferreira1, line 238: We already have a redux state `featureToggles` with all the feature flags updated. Is this the same value? If it is, I prefer keeping only one place with feature flags updated, so we don't get confused when accessing and updating them.

alexruzenhack: Sure!

andreabadesso (approved):

pedroferreira1 (approved):

PR #444, opened 2024-03-16
fix: wallet-service netowrk error

alexruzenhack:
Closes: 

### Acceptance Criteria
- Should fix a network error when processing handleTx

### Security Checklist

### DIFF
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -28,6 +28,7 @@ import {
   take,
   fork,
   spawn,
+  delay,
 } from 'redux-saga/effects';
 import { eventChannel } from 'redux-saga';
 import { getUniqueId } from 'react-native-device-info';
@@ -447,6 +448,43 @@ export function* listenForWalletReady(wallet) {
   }
 }
 
+/**
+ * An abstraction effect to call wallet's checkAddressMine method.
+ * This effect applies a retry strategy and throws a fatal error
+ * after max retry attempts.
+ *
+ * @param {Object} wallet Either fullnode or wallet-service wallet.
+ * @param {string[]} txAddresses A collection of addresses to check.
+ * @returns {{ [address: string]: boolean }} A map of type address:isMine.
+ */
+function* checkAddressMine(wallet, txAddresses) {
+  const maxRetries = 5;
+  const resultHandler = async (cb) => {
+    try {
+      const success = await cb();
+      return { success };
+    } catch (error) {
+      return { error }
+    }
+  };
+
+  let result = null;
+  for (let i = 0; i < maxRetries; i++) {
+    result = yield call(resultHandler, async () => await wallet.checkAddressesMine.bind(wallet)([...txAddresses]));
+    if (result.success) {
+      return result.success;
+    }
+    // attemp 0: 300ms
+    // attemp 1: 330ms
+    // attemp 2: 420ms
+    // attemp 3: 570ms
+    // attemp 4: 780ms
+    yield delay(300 + 30 * (i ** i));
+  }
+
+  yield put(onExceptionCaptured(result.error, true));
+}
+
 export function* handleTx(action) {
   const tx = action.payload;
   const wallet = yield select((state) => state.wallet);
@@ -488,9 +526,9 @@ export function* handleTx(action) {
     return acc;
   }, [{}, new Set([])],);
 
-  const txWalletAddresses = yield call(wallet.checkAddressesMine.bind(wallet), [...txAddresses]);
-  const tokensToDownload = [];
+  const txWalletAddresses = yield call(checkAddressMine, wallet, [...txAddresses]);
 
+  const tokensToDownload = [];
   for (const [tokenUid, addresses] of Object.entries(tokenAddressesMap)) {
     for (const [address] of addresses.entries()) {
       // txWalletAddresses should always have the address we requested, but we should double check


src/sagas/wallet.js
```diff
@@ -447,6 +448,43 @@ export function* listenForWalletReady(wallet) {
...
+  for (let i = 0; i < maxRetries; i++) {
+    result = yield call(resultHandler, async () => await wallet.checkAddressesMine.bind(wallet)([...txAddresses]));
+    if (result.success) {
+      return result.success;
+    }
+    // attemp 0: 300ms
+    // attemp 1: 330ms
+    // attemp 2: 420ms
+    // attemp 3: 570ms
+    // attemp 4: 780ms
```
andreabadesso, line 481:
```suggestion
    // attempt 0: 300ms
    // attempt 1: 330ms
    // attempt 2: 420ms
    // attempt 3: 570ms
    // attemp 4: 780ms
```

src/sagas/wallet.js
```diff
@@ -447,6 +448,43 @@ export function* listenForWalletReady(wallet) {
...
+ * An abstraction effect to call wallet's checkAddressMine method.
+ * This effect applies a retry strategy and throws a fatal error
+ * after max retry attempts.
+ *
+ * @param {Object} wallet Either fullnode or wallet-service wallet.
+ * @param {string[]} txAddresses A collection of addresses to check.
+ * @returns {{ [address: string]: boolean }} A map of type address:isMine.
+ */
+function* checkAddressMine(wallet, txAddresses) {
+  const maxRetries = 5;
```
andreabadesso, line 461: suggestion: Make this and the increase factor below constants

src/sagas/wallet.js
```diff
@@ -447,6 +448,43 @@ export function* listenForWalletReady(wallet) {
...
+      const success = await cb();
+      return { success };
+    } catch (error) {
+      return { error }
+    }
+  };
+
+  let result = null;
+  for (let i = 0; i < maxRetries; i++) {
+    result = yield call(resultHandler, async () => await wallet.checkAddressesMine.bind(wallet)([...txAddresses]));
```
andreabadesso, line 473: Why do you need this resultHandler? Can't you just try/catch here?

pedroferreira1 (commented):
The issue you mention in the PR description has two bugs, one in the wallet service and the other in the full node facade. Which of the bugs does this PR fixes?

One other thing, after reading the issue again, I think the wallet service bug is not clear to me, I'm missing some details of it in the issue. Any received tx causes this bug? Or just if the app is in background? In which conditions does the bug happen?

I have some more questions regarding the implementation itself but it depends on the answers above.

src/sagas/wallet.js
```diff
@@ -447,6 +448,43 @@ export function* listenForWalletReady(wallet) {
...
+    // attemp 2: 420ms
+    // attemp 3: 570ms
+    // attemp 4: 780ms
+    yield delay(300 + 30 * (i ** i));
+  }
+
+  yield put(onExceptionCaptured(result.error, true));
+}
+
 export function* handleTx(action) {
```
andreabadesso, line 488: So, from what I understood from the issue, this gets called while the phone is in background so the request fails and throws, is that correct?

Add a comment here explaining this potential issue

src/sagas/wallet.js
```diff
@@ -488,9 +526,9 @@ export function* handleTx(action) {
     return acc;
   }, [{}, new Set([])],);
 
-  const txWalletAddresses = yield call(wallet.checkAddressesMine.bind(wallet), [...txAddresses]);
-  const tokensToDownload = [];
+  const txWalletAddresses = yield call(checkAddressMine, wallet, [...txAddresses]);
```
andreabadesso, line 529: Instead of only retrying the first http request, why don't we retry the entire handleTx?

src/sagas/wallet.js
alexruzenhack: What happens is simple, the client request the wallet-service and a network issue happen, but the error is not handled. It do not happens all the time a wallet is on background, just sometimes.

src/sagas/wallet.js
alexruzenhack: I don't see why to retry the entire handle effect because this is an issue of the client that don't handle a possible network error.

alexruzenhack:
> The issue you mention in the PR description has two bugs, one in the wallet service and the other in the full node facade. Which of the bugs does this PR fixes?
> 
> One other thing, after reading the issue again, I think the wallet service bug is not clear to me, I'm missing some details of it in the issue. Any received tx causes this bug? Or just if the app is in background? In which conditions does the bug happen?
> 
> I have some more questions regarding the implementation itself but it depends on the answers above.

I answer it here: https://github.com/HathorNetwork/internal-issues/issues/257#issuecomment-2007626717

But in summary, this issue solves an unhandled network error, that can happen as a normal consequence of an HTTP request. It is not related to any transaction type of the background condition of the app.

alexruzenhack:
> > The issue you mention in the PR description has two bugs, one in the wallet service and the other in the full node facade. Which of the bugs does this PR fixes?
> > One other thing, after reading the issue again, I think the wallet service bug is not clear to me, I'm missing some details of it in the issue. Any received tx causes this bug? Or just if the app is in background? In which conditions does the bug happen?
> > I have some more questions regarding the implementation itself but it depends on the answers above.
> 
> I answer it here: [HathorNetwork/internal-issues#257 (comment)](https://github.com/HathorNetwork/internal-issues/issues/257#issuecomment-2007626717)
> 
> But in summary, this issue solves an unhandled network error, that can happen as a normal consequence of an HTTP request. It is not related to any transaction type or the background condition of the app.

src/sagas/wallet.js
alexruzenhack: Done at https://github.com/HathorNetwork/hathor-wallet-mobile/pull/444/commits/89c53c7f1c73034086e9cc5e4f59375df32fcf26

src/sagas/wallet.js
alexruzenhack: Done at https://github.com/HathorNetwork/hathor-wallet-mobile/pull/444/commits/89c53c7f1c73034086e9cc5e4f59375df32fcf26

src/sagas/wallet.js
alexruzenhack: Done at https://github.com/HathorNetwork/hathor-wallet-mobile/pull/444/commits/89c53c7f1c73034086e9cc5e4f59375df32fcf26

src/sagas/wallet.js
andreabadesso: A suggestion here would be to abstract this further and create a generic "retry saga" in helpers which receives a promise and a number of retries as a parameter and retries that promise until the max number of retries is reached

What I don't like here is adding an abstraction to this specific method just because it is the first http request in this saga

What do you think?

src/sagas/wallet.js
andreabadesso: I see, ok

src/sagas/wallet.js
alexruzenhack: Humm... It is good for me. I will implement this helper.

alexruzenhack: Done at https://github.com/HathorNetwork/hathor-wallet-mobile/pull/444/commits/dbbb4582a0ea6d03c5969b26ee5cb86421d906ae

src/constants.js
```diff
@@ -246,3 +246,7 @@ export const HTTP_REQUEST_TIMEOUT = 3000;
  * Any network that is not mainnet or testnet should be a privatenet.
  */
 export const NETWORK_PRIVATENET = 'privatenet';
+
+export const MAX_RETRIES_WS_CALL = 5;
```
pedroferreira1, line 250: Why `WS_CALL`? Why not only `MAX_RETRIES`?

src/constants.js
```diff
@@ -246,3 +246,7 @@ export const HTTP_REQUEST_TIMEOUT = 3000;
  * Any network that is not mainnet or testnet should be a privatenet.
  */
 export const NETWORK_PRIVATENET = 'privatenet';
+
```
pedroferreira1, line 249: Maybe add a comment above that the 3 constants below belong to the retry mechanism of sagas method calls. What do you think?

src/sagas/helpers.js
```diff
@@ -275,3 +279,54 @@ export function getNetworkSettings(state) {
...
+    // skip delay for last call
+    if (i === maxRetries - 1) {
+      continue;
+    }
+
+    // attempt 0: 300ms
+    // attempt 1: 330ms
+    // attempt 2: 420ms
+    // attempt 3: 570ms
+    // attempt 4: 780ms
```
pedroferreira1, line 328: You are trying 5 times in less than 2.5s. If the user had a network instability, this might not be enough for him to recover. I understand that we can't try forever and it doesn't matter the value we choose, it might not be enough anyway for the network to recover. But I think we could increase this amount, maybe even increase 1s in each attempt, then you give the user network 5s to recover (1s - 1s - 1s - 1s - 1s). 

What do you think?

src/sagas/helpers.js
```diff
@@ -275,3 +279,54 @@ export function getNetworkSettings(state) {
...
   return STORE.getItem(networkSettingsKeyMap.networkSettings) ?? state.networkSettings;
 }
+
+/**
+ * A request abstraction that applies a progressive retry strategy.
+ * One can define how many retries it should make or use the default value.
+ *
+ * @param {Promise<any>} request The async callback function to be executed.
+ * @param {number} maxRetries The max retries allowed, with default value.
+ * Notice this param should be at least 2 to make sense.
```
pedroferreira1, line 289: Why? The minimum should be 1.

If the parameter name contains 'retries', then 1 should be ok. You try once, then if it fails, you retry once.
 
However the implementation is different, if you set `maxRetries = 5`, then you try once and retries 4 times only, which should be 5.

src/sagas/helpers.js
alexruzenhack: Well, my experience was that these network flash instabilities didn't take long, actually. I believe 2.5s is enough for now imo.

src/sagas/helpers.js
alexruzenhack: Oh. Got it. You right, the implementation should account only the retries.

src/constants.js
alexruzenhack: Done https://github.com/HathorNetwork/hathor-wallet-mobile/pull/444/commits/c552aa618f7da09b9353666901f6bfe4d6ca441e

src/sagas/helpers.js
alexruzenhack: Changed to 8 attempts by default, the effect almost the same. Done at https://github.com/HathorNetwork/hathor-wallet-mobile/pull/444/commits/c552aa618f7da09b9353666901f6bfe4d6ca441e

src/sagas/helpers.js
alexruzenhack: Done at https://github.com/HathorNetwork/hathor-wallet-mobile/pull/444/commits/c552aa618f7da09b9353666901f6bfe4d6ca441e

src/constants.js
alexruzenhack: Done at https://github.com/HathorNetwork/hathor-wallet-mobile/pull/444/commits/d5af3d6c10af31743d2deb82b4fbb7289201851c

src/sagas/wallet.js
```diff
@@ -447,6 +448,25 @@ export function* listenForWalletReady(wallet) {
...
+ *
+ * @param {Object} wallet Either fullnode or wallet-service wallet.
+ * @param {string[]} txAddresses A collection of addresses to check.
+ * @returns {{
+ *   success?: { [address: string]: boolean };
+ *   error?: any;
+ * }} If success, it returns a map of type address:isMine,
+ * otherwise it will return the error definition.
+ */
+function* checkAddressMine(wallet, txAddresses) {
```
andreabadesso, line 464: Why don't you just pass wallet.checkAddressMine to the `progressiveRetryRequest` method it in the `handleTx` method instead of implementing this abstraction with the same name?

pedroferreira1 (approved):

src/sagas/wallet.js
andreabadesso: I would agree that it improves code readability if this wrapper was also being used somewhere else and it was a pattern in our error handling, but imho it only adds complexity to an already clear construct which is the try/catch

My suggestion would be to just add the try catch here until it reaches maxRetries, throw if it reaches it and let the caller handle the error as this is what all other methods in the code do

src/sagas/wallet.js
alexruzenhack: Removed.

andreabadesso (approved):

src/sagas/wallet.js
```diff
@@ -488,9 +489,17 @@ export function* handleTx(action) {
     return acc;
   }, [{}, new Set([])],);
 
-  const txWalletAddresses = yield call(wallet.checkAddressesMine.bind(wallet), [...txAddresses]);
-  const tokensToDownload = [];
+  let txWalletAddresses = null;
+  try {
+    const request = async () => wallet.checkAddressesMine.bind(wallet)([...txAddresses]);
+    txWalletAddresses = yield call(progressiveRetryRequest, request, 1);
```
pedroferreira1, line 495: Why do you have this parameter `1`?

pedroferreira1 (approved):

PR #456, opened 2024-04-01
feat: add components for nano contract transactions [9]

alexruzenhack:
### Motivation

This is a PR on a sequence to make the review easier.

### Acceptance Criteria
- Should add all components to be used on nano contract transactions screen

>[!NOTE]
>It depends on:
>- https://github.com/HathorNetwork/hathor-wallet-mobile/pull/455

### Security Checklist
- [x] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

PR #457, opened 2024-04-01
feat: add unregister nano contract saga effect [10]

alexruzenhack:
### Motivation

This is a PR on a sequence to make the review easier.

### Acceptance Criteria
- It should implement the unregister saga effect for nano contract
- It should add unregisterNanoContract to HybridStore
- It should clean registered nano contracts on cleanTokens
- It should update payload of `nanoContractAddressChangeRequest` action
- It should implement reducer `onNanoContractAddressChangeRequest`

>[!NOTE]
>It depends on:
>- https://github.com/HathorNetwork/hathor-wallet-mobile/pull/455
>- https://github.com/HathorNetwork/hathor-wallet-mobile/pull/456

### Security Checklist

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -149,6 +149,12 @@ export const types = {
   NANOCONTRACT_HISTORY_SUCCESS: 'NANOCONTRACT_HISTORY_SUCCESS',
   /* It indicates a Nano Contract history failed to load. */
   NANOCONTRACT_HISTORY_FAILURE: 'NANOCONTRACT_HISTORY_FAILURE',
+  /* It initiates an unregistration process of a Nano Contract. */
+  NANOCONTRACT_UNREGISTER_REQUEST: 'NANOCONTRACT_UNREGISTER_REQUEST',
+  /* It signals a successful completion of unregistration process. */
+  NANOCONTRACT_UNREGISTER_SUCCESS: 'NANOCONTRACT_UNREGISTER_SUCCESS',
+  /* It initiates a process to change the address on registered Nano Contract. */
+  NANOCONTRACT_ADDRESS_CHANGE_REQUEST: 'NANOCONTRACT_ADDRESS_CHANGE_REQUEST',
 };
 
 export const featureToggleInitialized = () => ({
@@ -1054,4 +1060,38 @@ export const nanoContractHistorySuccess = (payload) => ({
 export const nanoContractHistoryFailure = (payload) => ({
   type: types.NANOCONTRACT_HISTORY_FAILURE,
   payload,
+});
+
+/**
+ * Request unregistration of a Nano Contract by its key.
+ * @param {{
+ *   ncId: string,
+ * }} Nano Contract ID to unregister.
+ */
+export const nanoContractUnregisterRequest = (unregisterRequest) => ({
+  type: types.NANOCONTRACT_UNREGISTER_REQUEST,
+  payload: unregisterRequest,
+});
+
+/**
+ * Unregistration of a Nano Contract has finished with success.
+ * @param {{
+ *   ncId: string,
+ * }} Nano Contract ID unregistered.
+ */
+export const nanoContractUnregisterSuccess = (unregistered) => ({
+  type: types.NANOCONTRACT_UNREGISTER_SUCCESS,
+  payload: unregistered,
+});
+
+/**
+ * Request a change on the Nano Contract registered.
+ * @param {{
+ *   ncId: string;
+ *   newAddress: string;
+ * }} changeAddressRequest
+ */
+export const nanoContractAddressChangeRequest = (changeAddressRequest) => ({
+  type: types.NANOCONTRACT_ADDRESS_CHANGE_REQUEST,
+  payload: changeAddressRequest,
 });
--- a/src/reducers/reducer.js
+++ b/src/reducers/reducer.js
@@ -1460,6 +1460,7 @@ export const onNanoContractAddressChangeRequest = (state, { payload }) => {
   return {
     ...state,
     nanoContract: {
+      ...state.nanoContract,
       registered: {
         ...state.nanoContract.registered,
         [payload.ncId]: newRegisteredNc,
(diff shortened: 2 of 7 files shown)


src/sagas/nanoContract.js
```diff
@@ -252,9 +253,36 @@ export function* requestHistoryNanoContract({ payload }) {
...
+ * }} action with request payload.
+ */
+export function* unregisterNanoContract({ payload }) {
+  const { ncId } = payload;
+
+  const wallet = yield select((state) => state.wallet);
+  if (!wallet.isReady()) {
+    log.debug('Fail unregistering Nano Contract because wallet is not ready yet.');
+    // This will show user an error modal with the option to send the error to sentry.
+    yield put(onExceptionCaptured(new Error(failureMessage.walletNotReadyError), true));
```
andreabadesso, line 271: Is this really a fatal error?

src/store.js
```diff
@@ -144,6 +144,19 @@ class HybridStore extends MemoryStore {
...
+   * Unregister a nano contract.
+   *
+   * @param {string} ncId Nano Contract ID.
+   * @returns {Promise<void>}
+   * @async
+   */
+  async unregisterNanoContract(ncId) {
+    await super.unregisterNanoContract(ncId);
+    const contracts = STORE.getItem(REGISTERED_NANO_CONTRACTS_KEY) || {};
+    delete contracts[ncId];
```
pedroferreira1, line 157: Don't you need to set the new object to `STORE`?

src/actions.js
```diff
@@ -1055,3 +1061,37 @@ export const nanoContractHistoryFailure = (payload) => ({
...
+export const nanoContractUnregisterSuccess = (unregistered) => ({
+  type: types.NANOCONTRACT_UNREGISTER_SUCCESS,
+  payload: unregistered,
+});
+
+/**
+ * Request a change on the Nano Contract registered.
+ * @param {{
+ *   newAddress: string;
+ *   oldNc: Object;
```
pedroferreira1, line 1091: Why do you need the full object, not only the `ncId`?

src/actions.js
```diff
@@ -1055,3 +1061,37 @@ export const nanoContractHistoryFailure = (payload) => ({
...
+ * }} changeAddressRequest
+ */
+export const nanoContractAddressChangeRequest = (changeAddressRequest) => ({
```
pedroferreira1, line 1094: This is not currently in use, right?

src/reducers/reducer.js
```diff
@@ -1329,6 +1331,37 @@ export const onNanoContractRegisterSuccess = (state, { payload }) => ({
...
+ * @param {{
+ *   payload: {
+ *     ncId: string,
+ *   }
+  * }} action
+ */
+export const onNanoContractUnregisterSuccess = (state, { payload }) => {
+  const { ncId } = payload;
+
+  const newRegisteredContracts = { ...state.nanoContract.registered, };
```
pedroferreira1, line 1345:
```suggestion
  const newRegisteredContracts = { ...state.nanoContract.registered };
```

src/store.js
alexruzenhack: No, I don't. There aren't a new object to set, `contracts` is a reference to the object, and we only need remove an entry from this object to set it right.

src/actions.js
alexruzenhack: Changed to ncId.

src/sagas/nanoContract.js
```diff
@@ -254,9 +256,36 @@ export function* requestHistoryNanoContract({ payload }) {
...
+    return;
+  }
+
+  yield call(wallet.storage.unregisterNanoContract.bind(wallet.storage), ncId);
+
+  log.debug(`Success unregistering Nano Contract. ncId = ${ncId}`);
+  // emit action NANOCONTRACT_REGISTER_SUCCESS
```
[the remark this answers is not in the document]
alexruzenhack: Removed the comment.

src/sagas/nanoContract.js
```diff
@@ -80,7 +81,8 @@ export function* registerNanoContract({ payload }) {
 
   let ncState = null;
   try {
-    ncState = yield call(ncApi.getNanoContractState.bind(ncApi), ncId);
+    const response = yield call(ncApi.getNanoContractState.bind(ncApi), ncId);
```
andreabadesso, line 84: Also check if `response.(data?).success` is `true`

andreabadesso (changes requested):

src/store.js
pedroferreira1: Yes, I agree we should do it.

pedroferreira1 (approved):

alexruzenhack:
>My suggestion is that we avoid mutating the object in this PR, use STORE.setItem to store it even though it's redundant and open a KTLO to return immutable objects in the Storage

Ok. I will open a PR. However, to achieve immutability we need more than just use the spread operator.

src/store.js
alexruzenhack: > My suggestion is that we avoid mutating the object in this PR, use STORE.setItem to store it even though it's redundant and open a KTLO to return immutable objects in the Storage

Ok. I will open a PR. However, to achieve immutability we need more than just use the spread operator.

src/sagas/nanoContract.js
```diff
@@ -254,9 +256,35 @@ export function* requestHistoryNanoContract({ payload }) {
...
+ *     ncId: string;
+ *   }
+ * }} action with request payload.
+ */
+export function* unregisterNanoContract({ payload }) {
+  const { ncId } = payload;
+
+  const wallet = yield select((state) => state.wallet);
+  if (!wallet.isReady()) {
+    log.debug('Fail unregistering Nano Contract because wallet is not ready yet.');
```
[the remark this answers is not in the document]
alexruzenhack: I believe we shouldn't because there was no error, it is just a condition we don't tolerate.

src/store.js
andreabadesso: I understand, we can discuss this in the KTLO, but at the very least we should return a new instance of the object, to prevent it from being mutated by the api consumer

Issue #464, opened 2024-04-11
Issues found on Network Settings QA for v0.27.0-rc.3


alexruzenhack:
Related to: https://github.com/HathorNetwork/internal-issues/issues/249

## Network Settings

- [x] Issue: error on wallet initialization process trying to set wallet-service when it is enabled but the URL is empty
- [x] Issue: wallet behavior inconsistency after quit and reopen the wallet because service URLs aren't reapplied when a custom network settings is set
- [ ] Issue: after send a new network settings and let the app be blocked, the success feedback modal shows up on the locked scree

PR #465, opened 2024-04-11
fix: custom network settings processing

alexruzenhack:
### Motivation

Issues found on QA for v0.27.0-rc.3

Closes: https://github.com/HathorNetwork/hathor-wallet-mobile/issues/464

### Acceptance Criteria
- It should fix network settings processing on wallet
- It should fix wallet behavior consistency regarding custom network settings

### Security Checklist

### DIFF
--- a/src/sagas/networkSettings.js
+++ b/src/sagas/networkSettings.js
@@ -6,7 +6,6 @@ import {
   networkSettingsPersistStore,
   networkSettingsUpdateInvalid,
   networkSettingsUpdateFailure,
-  networkSettingsUpdateState,
   networkSettingsUpdateSuccess,
   networkSettingsUpdateWaiting,
   types,
--- a/src/sagas/networkSettings.js
+++ b/src/sagas/networkSettings.js
@@ -156,7 +156,7 @@ export function* updateNetworkSettings(action) {
   // - nodeUrl as fallback
   let potentialNetwork;
   let network;
-  if (walletServiceUrl && useWalletService) {
+  if (useWalletService && !isEmpty(walletServiceUrl)) {
     config.setWalletServiceBaseUrl(walletServiceUrl);
     config.setWalletServiceBaseWsUrl(walletServiceWsUrl);
 
@@ -222,6 +222,7 @@ export function* updateNetworkSettings(action) {
     nodeUrl,
     explorerUrl,
     explorerServiceUrl,
+    txMiningServiceUrl,
     walletServiceUrl,
     walletServiceWsUrl,
   };
--- a/src/sagas/networkSettings.js
+++ b/src/sagas/networkSettings.js
@@ -36,11 +36,6 @@ import { isWalletServiceEnabled } from './wallet';
  * Initialize the network settings saga when the wallet starts successfully.
  */
 export function* initNetworkSettings() {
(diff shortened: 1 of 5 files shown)


src/sagas/networkSettings.js
```diff
@@ -148,15 +144,15 @@ export function* updateNetworkSettings(action) {
...
   config.setExplorerServiceBaseUrl(explorerServiceUrl);
   config.setServerUrl(nodeUrl);
-  config.setTxMiningUrl(txMiningServiceUrl);
 
   // - walletServiceUrl has precedence
   // - nodeUrl as fallback
   let potentialNetwork;
   let network;
-  if (walletServiceUrl && useWalletService) {
+  if (useWalletService && !isEmpty(walletServiceUrl)) {
```
andreabadesso, line 155: Oh, we were setting the url to null?

Ok, so this will now just fallback to the URL in the constants file

src/sagas/wallet.js
```diff
@@ -151,10 +153,29 @@ export function* startWallet(action) {
...
+    // If the wallet is initialized from quit state it must
+    // update the network settings on redux state
+    yield put(networkSettingsUpdateState(networkSettings));
+  } else {
+    networkSettings = yield select(getNetworkSettings);
+  }
 
   let wallet;
-  if (useWalletService) {
+  if (useWalletService && !isEmpty(networkSettings.walletServiceUrl)) {
```
andreabadesso, line 178: I don't understand this... What if he never used the network settings but is using the wallet-service (which is now 100% of our userbase)? He will have an empty networkSettings value, right? Is this a bug?

andreabadesso (changes requested):

src/sagas/networkSettings.js
```diff
@@ -36,11 +36,6 @@ import { isWalletServiceEnabled } from './wallet';
  * Initialize the network settings saga when the wallet starts successfully.
  */
 export function* initNetworkSettings() {
-  const customNetwork = STORE.getItem(networkSettingsKeyMap.networkSettings);
-  if (customNetwork) {
-    yield put(networkSettingsUpdateState(customNetwork));
```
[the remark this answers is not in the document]
alexruzenhack: Removed

src/sagas/networkSettings.js
alexruzenhack: No, if `walletServiceUrl` is empty we will set an empty value in redux state. When the wallet reloads and wallet service is enabled, it will try to connect but will fail, then it will fallback to fullnode.

src/sagas/wallet.js
alexruzenhack: No. The `walletServiceUrl` has a value in redux state by default. Beyond that, the pre-settings for network settings, `mainnet` and `testnet`, also have default values for `walletServiceUrl`. So, wallet-service will work fine. It will not work only if the user customize the network settings removing the `walletServiceUrl`.

andreabadesso: Oh, I see!

I was confusing `devSettings` with `networkSettings`.

So at this point, `networkSettings` will always have a value, either populated from the `PRE_SETTINGS` or from the `devSettings`

andreabadesso (approved):

pedroferreira1 (approved):

PR #470, opened 2024-04-22
fix: useWalletService by forcefully disabling it when wallet service URL is empty

alexruzenhack:
### Acceptance Criteria
- It should fix wallet-service usage when it should not be used: when a user empty wallet service URL

### Security Checklist

### DIFF
--- a/src/sagas/wallet.js
+++ b/src/sagas/wallet.js
@@ -124,7 +124,7 @@ export function* isWalletServiceEnabled() {
 
   let walletServiceEnabled = yield call(checkForFeatureFlag, WALLET_SERVICE_FEATURE_TOGGLE);
 
-  // At this point the networkSettings already have been set by startWallet.
+  // At this point, the networkSettings have already been set by startWallet.
   const networkSettings = yield select(getNetworkSettings);
   if (walletServiceEnabled && isEmpty(networkSettings.walletServiceUrl)) {
     // In case of an empty value for walletServiceUrl, it means the user
@@ -142,6 +142,13 @@ export function* startWallet(action) {
     words,
     pin,
   } = action.payload;
+
+  // clean memory storage and metadata before starting the wallet.
+  // This should be cleaned when stopping the wallet,
+  // but the wallet may be closed unexpectedly
+  const storage = STORE.getStorage();
+  yield call([storage.store, storage.store.cleanMetadata]); // clean metadata on memory
+  yield call([storage, storage.cleanStorage], true); // clean transaction history
 
   // As this is a core setting for the wallet, it should be loaded first.
   // Network settings either from store or redux state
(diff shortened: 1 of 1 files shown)


src/sagas/wallet.js
```diff
@@ -122,7 +122,17 @@ export function* isWalletServiceEnabled() {
     yield call(() => AsyncStorage.removeItem(IGNORE_WS_TOGGLE_FLAG));
   }
 
-  const walletServiceEnabled = yield call(checkForFeatureFlag, WALLET_SERVICE_FEATURE_TOGGLE);
+  let walletServiceEnabled = yield call(checkForFeatureFlag, WALLET_SERVICE_FEATURE_TOGGLE);
+
+  // At this point the networkSettings already have been set by startWallet.
```
andreabadesso, line 127:
```suggestion
  // At this point, the networkSettings have already been set by startWallet.
```

src/sagas/wallet.js
```diff
@@ -122,7 +122,17 @@ export function* isWalletServiceEnabled() {
...
+  const networkSettings = yield select(getNetworkSettings);
+  if (walletServiceEnabled && isEmpty(networkSettings.walletServiceUrl)) {
```
andreabadesso, line 129: I think the user experience for this is bad

I think we should add a checkbox in the network settings screen, "Use wallet-service if available?", validate the wallet-service URL if it is checked while saving

andreabadesso: But since this is a bug, I think we can do this in a future PR, if you both agree with my suggestion

@pedroferreira1 @alexruzenhack

andreabadesso (changes requested):

src/sagas/wallet.js
```diff
@@ -174,6 +165,26 @@ export function* startWallet(action) {
...
+
+  yield put(setUseWalletService(useWalletService));
+  yield put(setAvailablePushNotification(usePushNotification));
+
+  // clean storage and metadata before starting the wallet
+  // this should be cleaned when stopping the wallet,
+  // but the wallet may be closed unexpectedly
+  const storage = STORE.getStorage();
+  yield storage.store.cleanMetadata();
+  yield storage.cleanStorage(true);
```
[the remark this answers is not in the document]
alexruzenhack: > I know this is not being changed in this PR, but is this actually awaiting those promises? I think it's just yielding the promise, I don't think redux-saga will actually await them

You right, it is not awaiting.

src/sagas/wallet.js
alexruzenhack: Here the KTLO: https://github.com/HathorNetwork/hathor-wallet-mobile/issues/471

pedroferreira1 (approved):

andreabadesso (approved):

PR #473, opened 2024-04-30
fix: allows push notification settings only when using mainnet network

alexruzenhack:
### Acceptance Criteria
- It should add a condition to allow use the push notification settings only when using the mainnet network
- It should closes https://github.com/HathorNetwork/internal-issues/issues/285

### Security Checklist

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -13,7 +13,7 @@ import { Linking, Platform, Text } from 'react-native';
 import { getStatusBarHeight } from 'react-native-status-bar-height';
 import { isEmpty } from 'lodash';
 import baseStyle from './styles/init';
-import { KEYCHAIN_USER } from './constants';
+import { KEYCHAIN_USER, NETWORK_MAINNET } from './constants';
 import { STORE } from './store';
 import { TxHistory } from './models';
 import { COLORS, STYLE } from './styles/themes';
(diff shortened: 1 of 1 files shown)


src/utils.js
```diff
@@ -383,5 +383,12 @@ export function combineURLs(baseURL, relativeURL) {
  * @returns {Boolean} true if available, false otherwise.
  */
 export const isPushNotificationAvailableForUser = (state) => (
-  state.pushNotification.available && state.pushNotification.deviceRegistered
+  state.pushNotification.available
+    // On iOS a simulator can't register a device token on APNS
+    && state.pushNotification.deviceRegistered
+    // TODO: We should drop this condition when we add support other networks
```
pedroferreira1, line 389: Do we have an issue to discuss this?

pedroferreira1 (approved):

src/utils.js
alexruzenhack: This one: https://github.com/HathorNetwork/internal-issues/issues/285

We can keep it open until we find a definite solution. What do you think?

andreabadesso (approved):

src/utils.js
alexruzenhack: Here is is a PR to track this TODO: https://github.com/HathorNetwork/internal-issues/issues/288

PR #475, opened 2024-04-30
fix: remove registered tokens when changing network settings

alexruzenhack:
### Acceptance Criteria
- It should remove registered tokens when a user changes the network settings
- It should closes https://github.com/HathorNetwork/internal-issues/issues/286

It depends on:
- https://github.com/HathorNetwork/hathor-wallet-lib/pull/648

### Security Checklist

### DIFF
--- a/src/sagas/networkSettings.js
+++ b/src/sagas/networkSettings.js
@@ -299,10 +299,8 @@ export function* persistNetworkSettings(action) {
     return;
   }
 
-  // Remove registered tokens only
-  yield call([wallet.storage, wallet.storage.cleanStorage], false, false, true)
   // Stop wallet and clean its storage without clean its access data.
-  wallet.stop({ cleanStorage: true, cleanAddresses: true });
+  wallet.stop({ cleanStorage: true, cleanAddresses: true, cleanTokens: true });
   // This action should clean the tokens history on redux.
   // In addition, the reload also clean the inmemory storage.
   yield put(reloadWalletRequested());
--- a/src/sagas/networkSettings.js
+++ b/src/sagas/networkSettings.js
@@ -29,7 +29,7 @@ import {
   getFullnodeNetwork,
   getWalletServiceNetwork,
 } from './helpers';
-import { STORE } from '../store';
+import { REGISTERED_TOKENS_KEY, STORE } from '../store';
 import { isWalletServiceEnabled } from './wallet';
 import { logger } from '../logger';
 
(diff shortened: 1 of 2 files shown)


src/sagas/networkSettings.js
```diff
@@ -289,6 +289,8 @@ export function* persistNetworkSettings(action) {
     return;
   }
 
+  // Remove registered tokens
+  STORE.removeItem(REGISTERED_TOKENS_KEY);
```
pedroferreira1, line 293: Does it remove from the wallet storage as well, in the `wallet.storage`?

alexruzenhack: This one: https://github.com/HathorNetwork/internal-issues/issues/285

alexruzenhack: Yes, it does.

alexruzenhack: I refactored it.

andreabadesso (approved):

src/sagas/networkSettings.js
```diff
@@ -289,6 +289,8 @@ export function* persistNetworkSettings(action) {
     return;
   }
 
+  // Remove registered tokens only
+  yield call([wallet.storage, wallet.storage.cleanStorage], false, false, true)
   // Stop wallet and clean its storage without clean its access data.
   wallet.stop({ cleanStorage: true, cleanAddresses: true });
```
pedroferreira1, line 295: The method `stop` from the wallet already calls the `cleanStorage`. With this change we are calling it twice.

We should add a new parameter to the `stop` function to support `cleanTokens` as well.

alexruzenhack: Oh, ok! I will open an issue for it.

alexruzenhack: See https://github.com/HathorNetwork/internal-issues/issues/290

pedroferreira1: I think we should do this in this PR, it's a problem to merge like this

andreabadesso: So, are we going to do this in this PR? Was it solved in a different way?

andreabadesso (changes requested):

src/sagas/networkSettings.js
alexruzenhack: Done

andreabadesso (approved):

src/store.js
```diff
@@ -143,6 +143,22 @@ class HybridStore extends MemoryStore {
...
+
+  /**
+   * Clean the storage.
+   * @param {boolean} cleanHistory if we should clean the transaction history.
+   * @param {boolean} cleanAddresses if we should clean the addresses.
+   * @param {boolean} cleanTokens if we should clean the registered tokens.
+   * @async
+   * @returns {Promise<void>}
+   */
+  async cleanStorage(cleanHistory = false, cleanAddresses = false, cleanTokens = false) {
```
pedroferreira1, line 155: Why do you need this method?

alexruzenhack: Oh, this is needed to also clean the app's AsyncStorage managed by the `HybridStore`.

pedroferreira1 (approved):

PR #476, opened 2024-05-03
feat: add firstBlock to TxHistory model

alexruzenhack:
### Acceptance Criteria
- It should add `firstBlock` property to TxHistory model

### Security Checklist


pedroferreira1:
It's missing the screenshot with this new label

src/models.js
```diff
@@ -110,6 +116,11 @@ export class TxHistory {
     return this.version === constants.NANO_CONTRACTS_VERSION;
   }
 
+  hasFirstBlock() {
+    // not null, not undefined and not string empty
+    return this.firstBlock != null && this.firstBlock.trim() !== '';
```
tuliomir, line 121: todo: It's important that this validation is made with `!==` instead of `!=`.
The behavior of comparing something with `!=` null can be hard to predict.
```suggestion
    return this.firstBlock !== null && this.firstBlock.trim() !== '';
```

alexruzenhack: The use of `!= null` was intentional. What it does check of the value is not null and not undefined.

tuliomir: As we've talked, I just feel that the behavior of the `==` / `!=` operators [is a bit hard to predict](https://blog.webdevsimplified.com/2020-08/==-vs-===/#:~:text=Conclusion,equals%20is%20the%20preferred%20method.).
But if this has been well tested I see no problem in keeping it.

alexruzenhack: This is the precise case it is good to use, indeed:
image

tuliomir (approved):

src/sagas/tokens.js
```diff
@@ -136,7 +136,7 @@ function* fetchTokenHistory(action) {
       return;
     }
 
-    const response = yield call(wallet.getTxHistory.bind(wallet), { token_id: tokenId });
+    const response = yield call([wallet, wallet.getTxHistory], { token_id: tokenId });
```
alexruzenhack, line 139: This is the way to call with context using saga-redux `call` effect.

pedroferreira1 (approved):

PR #483, opened 2024-05-15
feat: add nano contract feature toggle to dashboard

alexruzenhack:
### Acceptance Criteria
- it should add Nano Contract feature toggle in the Dashboard

Nano Contract `disabled`

Nano Contract `enabled`

### Security Checklist

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -12,7 +12,7 @@ import { t } from 'ttag';
 import { Linking, Platform, Text } from 'react-native';
 import { getStatusBarHeight } from 'react-native-status-bar-height';
 import baseStyle from './styles/init';
-import { KEYCHAIN_USER } from './constants';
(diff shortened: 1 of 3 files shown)


src/utils.js
```diff
@@ -392,3 +392,14 @@ export function combineURLs(baseURL, relativeURL) {
...
 );
+
+/**
+ * Get Nano Contract feature toggle state from redux.
+ *
+ * @param {Object} state Redux store state
+ *
+ * @returns {boolean} the Nano Contract feature toggle state.
+ */
+export const getNanoContractFeatureToggle = (state) => (
```
r4mmer, line 403: Should we create a new file for feature flag selectors?

r4mmer (approved):

src/utils.js
alexruzenhack: Hummm... I think it is nice, but doing it for all selectors. We have a folder for `reducers`, maybe we can create a folder for `selectors`. What do you think?

Issue #334, opened 2023-09-29
Standardize actions with a taxonomy


alexruzenhack:
I used in this commit https://github.com/HathorNetwork/hathor-wallet-mobile/commit/6980cac80982b4437f57fc654920f27c38aa38eb the following taxonomy for the action name.

```txt
I used the following taxonomy for the actions:
- <subject>_<action>[_<result>]

The `<subject>` and the `<result>` must be nouns and `<action>`
must be a verb.

For instance:

- NETWORKSETTINGS_UPDATE
- NETWORKSETTINGS_UPDATE_SUCCESS
- NETWORKSETTINGS_UPDATE_FAILURE
```

@andreabadesso has agree with it in this comment https://github.com/HathorNetwork/hathor-wallet-mobile/pull/333#discussion_r1340646133. Now we should apply this pattern to all the other actions.

PR #495, opened 2024-06-10
feat: nano contract integration with wallet connect

alexruzenhack:
### Acceptance Criteria

- Add new nano contract modal
  - It should appear when the wallet connect receive an RPC call to create a new nano contract transaction
- Add new nano contract details screen
  - It should allow the user to change the caller address
  - It should allows the user to decline the transaction
  - It should allows the user to accept the transaction
  - It should show a loading feedback content when transaction information is loading
  - It should show a loading feedback content when transaction is processing after user accepts it
  - It should show a success feedback content when transaction is sent with success
  - It should show a failed feedback content when transaction is not sent with success
  - It should show a feedback content instruction user to register Nano Contract when Nano Contract is not registered
    - It should allows the user to decline the transaction only
   - It should show a loading value on `Blueprint Name` and `Caller` on `initialize` call while background requests are being processed
   - It should show a warning value on `Blueprint Name` or `Caller` on `initialize` call if any request fails
   - It should instruct the user to select an address on `Caller` if request fails
   - It should allows the user to select an address on `Caller` if requests fails
   - It should show argument's name if a method call has arguments
   - It should show `Position $` for each positional argument as a fallback for argument's name when blueprint info fails to load
   - It should load the details of a not registered token used in actions
     - It should show the symbol for the loaded token details
     - It should show an error message of feedback in case it fails to load some token details
     - It should load a cut of token uid as a fallback for not loading the token details

##### Loading tx info

https://github.com/HathorNetwork/hathor-wallet-mobile/assets/5992210/25ce4323-4aa8-4cfd-917e-6a782b9a454e

##### Declining tx

https://github.com/HathorNetwork/hathor-wallet-mobile/assets/5992210/5bde3bbc-1d13-4601-96c8-adcf8148604c

##### Changing caller address

https://github.com/HathorNetwork/hathor-wallet-mobile/assets/5992210/5909358d-27bc-4f0b-b74b-addfe357fee4

##### Accepting tx with failure

https://github.com/HathorNetwork/hathor-wallet-mobile/assets/5992210/686f2f98-a47c-40d8-a717-4f6d0ab78e36

##### Accepting tx with failure and trying again

https://github.com/HathorNetwork/hathor-wallet-mobile/assets/5992210/b5c26358-38b8-4313-bc29-d0aa8ab09e87

#### Accepting tx with success

https://github.com/HathorNetwork/hathor-wallet-mobile/assets/5992210/6b8a3c8f-73fa-4a14-b13d-d00fbe9343f3

#### Handling many types of transactions for Nano Contract registered

https://github.com/HathorNetwork/hathor-wallet-mobile/assets/5992210/a9b84488-9c13-4cb6-b934-f7c40ca37340

#### Handling `initialize` with requests failure

https://github.com/HathorNetwork/hathor-wallet-mobile/assets/5992210/2e3b738e-b79d-4d83-8a44-732b7b46bcb3

#### Handling requests for Nano Contract not registered

https://github.com/HathorNetwork/hathor-wallet-mobile/assets/5992210/7a5f317e-a453-45d4-b07b-497ae6b53fb3

#### Showing UID cut as fallback for not registered token and feedback error message

---

>[!WARNING]
>It depends on:
>- [ ] An API update on hathor-core, related to the nano contract state.
>- [x] Proper implementation of nano contract RPC handler

### Security Checklist


src/components/WalletConnect/NanoContract/NanoContractMethodArgs.js
```diff
@@ -0,0 +1,92 @@
...
+import { t } from 'ttag';
+import { get } from 'lodash';
+import { COLORS } from '../../../styles/themes';
+import { commonStyles } from '../theme';
+
+/**
+ * It renders a list of method arguments for when the Nano Contract executes.
+ *
+ * @param {Object} props
+ * @param {string[]} props.ncArgs A list of method's argument.
```
andreabadesso, line 24: Missing method and blueprintId on docs

src/components/WalletConnect/NanoContract/NewNanoContractTransactionModal.js
```diff
@@ -0,0 +1,177 @@
...
+import { ModalBase } from '../../ModalBase';
+import SimpleButton from '../../SimpleButton';
+import { setNewNanoContractTransaction, walletWalletReject } from '../../../actions';
+import { WALLET_STATUS } from '../../../sagas/wallet';
+import { NANO_CONTRACT_INFO_URL } from '../../../constants';
+
+export const NewNanoContractTransactionModal = () => {
+  const navigation = useNavigation();
+  const dispatch = useDispatch();
+  const readMoreUrl = NANO_CONTRACT_INFO_URL;
```
andreabadesso, line 28: Is it published already?

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
```diff
@@ -0,0 +1,313 @@
...
+import { DappContainer } from './DappContainer';
+import { NanoContractExecInfo } from './NanoContractExecInfo';
+import { NanoContractActions } from './NanoContractActions';
+import { NanoContractMethodArgs } from './NanoContractMethodArgs';
+import { DeclineModal } from './DeclineModal';
+
+/**
+ * @param {Object} props
+ * @param {Object} props.ncTxRequest
+ * @param {Object} props.ncTxRequest.nc
```
andreabadesso, line 48: Also document the `nc` prop attributes?

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
```diff
@@ -0,0 +1,313 @@
...
+  const registeredTokensMetadata = useSelector((state) => state.tokenMetadata);
+  // Use it to add loading feedback
+  const metadataLoaded = useSelector((state) => state.metadataLoaded);
+  const blueprintInfo = useSelector((state) => state.nanoContract.blueprint[nc.blueprintId]);
+
+  const [showSelectAddressModal, setShowSelectAddressModal] = useState(false);
+  const [showDeclineModal, setShowDeclineModal] = useState(false);
+  // If nano-contract's method is 'initialize' then the expression
+  // should be resolved to firstAddress value by default.
+  // However, the wallet can have trouble to
```
andreabadesso, line 72: Trouble to what?

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
```diff
@@ -0,0 +1,313 @@
...
+  };
+
+  const onDeclineTransaction = () => {
+    setShowDeclineModal(true);
+  };
+  const onDeclineConfirmation = () => {
+    setShowDeclineModal(false);
+    dispatch(walletWalletReject());
+    navigation.goBack();
+    // TODO: It is not being dimissed automatically
```
andreabadesso, line 101: I refactored this in my PR, so leave this as a TODO

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
```diff
@@ -0,0 +1,313 @@
...
+  // If nano-contract's method is 'initialize' then the expression
+  // should be resolved to firstAddress value by default.
+  // However, the wallet can have trouble to
+  const [ncAddress, setNcAddress] = useState(registeredNc?.address || firstAddress.address);
+  const ncToAccept = useMemo(() => ({
+    ...nc,
+    caller: ncAddress,
+  }), [ncAddress])
+
+  // Controle SelectAddressModal
```
andreabadesso, line 79:
```suggestion
```

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
```diff
@@ -0,0 +1,313 @@
...
+
+    // Request blueprint info if not present to feed the components:
+    // - NanoContractExecInfo, and
+    // - NanoContractMethodArgs
+    if (!blueprintInfo) {
+      dispatch(nanoContractBlueprintInfoRequest(nc.blueprintId));
+    }
+
+    // Get tokens metadata
+    // actions can be null?
```
andreabadesso, line 134: I don't know, can it?

We already have lodash in the project, it doesn't hurt to defend against it:

```javascript
    const tokensUid = get(nc, 'actions', []).map((each) => each.token) || [];
```

src/sagas/wallet.js
```diff
@@ -350,13 +350,22 @@ export function* loadTokens() {
...
+
 /**
  * The wallet needs each token metadata to show information correctly
  * So we fetch the tokens metadata and store on redux
  */
 export function* fetchTokensMetadata(tokens) {
-  // No tokens to load
+  // No tokens to load, set metadata as loaded
   if (!tokens.length) {
+    yield put(tokenMetadataUpdated({}));
```
andreabadesso, line 368: Why are you passing an empty object?

src/sagas/wallet.js
```diff
@@ -381,8 +390,7 @@ export function* fetchTokensMetadata(tokens) {
   const tokenMetadatas = {};
   for (const response of responses) {
     if (response.type === types.TOKEN_FETCH_METADATA_FAILED) {
-      // eslint-disable-next-line
-      console.log('Error downloading metadata of token', response.tokenId);
+      log(`Error downloading metadata of token ${response.tokenId}.`);
```
andreabadesso, line 393: This will crash, log is an object with `warn`, `error` and `debug` methods

I've seen this in other files, can you please review?

src/screens/WalletConnect/NewNanoContractTransactionScreen.js
```diff
@@ -0,0 +1,44 @@
...
+  StyleSheet,
+  View,
+} from 'react-native';
+import { t } from 'ttag';
+import HathorHeader from '../../components/HathorHeader';
+import OfflineBar from '../../components/OfflineBar';
+import { NewNanoContractTransactionRequest } from '../../components/WalletConnect/NanoContract/NewNanoContractTransactionRequest';
+import { COLORS } from '../../styles/themes';
+
+export function NewNanoContractTransactionScreen({ route }) {
```
andreabadesso, line 19: Why don't you use `useRoute()`?

andreabadesso (changes requested):

src/sagas/wallet.js
pedroferreira1: We should just return

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
```diff
@@ -0,0 +1,313 @@
...
+    const tokensUid = nc.actions?.map((each) => each.token) || [];
+    const tokensMetadataToDownload = [];
+
+    tokensUid.forEach((uid) => {
+      if (uid !== DEFAULT_TOKEN.uid && !(uid in registeredTokensMetadata)) {
+        tokensMetadataToDownload.push(uid);
+      }
+    });
```
pedroferreira1, line 142: Why do you need this? If I'm not wrong, this token metadata is used to get NFT and metadata information. Why do you need to get these from the actions tokens?

src/components/HathorFlatList.js
```diff
@@ -15,7 +15,7 @@ import { COLORS } from '../styles/themes';
  * param {FlatListProps} props
  */
 export const HathorFlatList = (props) => (
-  <View style={styles.wrapper}>
+  <View style={[styles.wrapper, props.wrapperStyle]}>
```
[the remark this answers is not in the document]
alexruzenhack: I've improved documentation.

src/components/WalletConnect/NanoContract/NanoContractExecInfo.js
```diff
@@ -0,0 +1,168 @@
...
+          </TextValue>
+          {blueprintName && (
+            <FrozenTextValue>{blueprintName}</FrozenTextValue>
+          )}
+          {isBlueprintInfoLoading && (
+            <WarnTextValue>{t`Loading...`}</WarnTextValue>
+          )}
+          {hasBlueprintInfoFailed && (
+            <WarnTextValue>{blueprintInfo.error}</WarnTextValue>
+          )}
```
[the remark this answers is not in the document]
alexruzenhack: I've solved it by adding a condition in `isBlueprintInfoLoading` and `hasBlueprintInfoFailed`, which is "only assess the blueprint state if the nano contract is not registered".

src/components/WalletConnect/NanoContract/NanoContractActions.js
```diff
@@ -0,0 +1,156 @@
...
+    value: [commonStyles.text, commonStyles.value],
+  });
+
+  return (
+    <View>
+      <View>
+        <Text style={commonStyles.sectionTitle}>{t`Action List`}</Text>
+      </View>
+      <HathorFlatList
+        scrollEnabled={false} // it avoids nearest scrolls
```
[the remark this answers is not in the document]
alexruzenhack: I've used `HathorFlatList` just for visual consistency, otherwise I would have to build another component only for this list. I don't know for sure what happens if we have 30 actions, I will try here and check. But `scrollEnabled=false` avoids pass the parent scroll control to the this component. It means if the user tap over an action and swipe up or down, the control will still be kept in the parent component, thus moving the parent pane.

src/components/WalletConnect/NanoContract/NewNanoContractTransactionModal.js
alexruzenhack: Yes, it is.

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
alexruzenhack: I don't see benefit to use lodash here compared to use the **[optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) operator** `?`.

src/sagas/wallet.js
alexruzenhack: This use of `tokenMetadataUpdated` results the state `metadataLoaded` to be `true`. We need this state in `NewNanoContractTransactionRequest` to [evaluate some flags](https://github.com/HathorNetwork/hathor-wallet-mobile/pull/495/files#diff-f4c181ecb3766d7252d1d68de29c6e0431178ae696108807b41011b64feef35fR163-R169).

alexruzenhack: Passing an empty object has no effect for the `tokenMetadata` state as we can [see here](https://github.com/HathorNetwork/hathor-wallet-mobile/blob/12e4539d85afdc2d61a2e99813a516414e5cf9cc/src/reducers/reducer.js#L913).

src/screens/WalletConnect/NewNanoContractTransactionScreen.js
alexruzenhack: This is a screen component and it is served by the navigation itself, which make it possible to get the route in its prop directly. I don't see why not get it.

src/constants.js
```diff
@@ -240,6 +240,20 @@ export const NANOCONTRACT_REGISTER_STATUS = {
...
+  LOADING: 'loading',
+  SUCCESSFUL: 'successful',
+};
+
+export const NANOCONTRACT_BLUEPRINTINFO_STATUS = {
+  READY: 'ready',
+  FAILED: 'failed',
+  LOADING: 'loading',
+  SUCCESSFUL: 'successful',
+};
```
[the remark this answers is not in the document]
alexruzenhack: I think you are right to raise this discussion and I believe it would fit a refactoring KTLO. I've opened this issue https://github.com/HathorNetwork/hathor-wallet-mobile/issues/522.

src/components/FrozenTextValue.js
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * @param {Object} props
+ * @param {boolean} props.title It sets font weight to bold and a larger font size
+ * @param {boolean} props.bold It sets font weight to bold
+ * @param {boolean} props.oneline It sets numberOfLines to 1
+ * @param {boolean} props.shrink It sets flexShrink to 1
+ * @param {boolean} props.pb4 It sets padding bottom to 4
+ */
+export const FrozenTextValue = ({ title, bold, oneline, shrink, pb4, children }) => (
```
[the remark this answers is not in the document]
alexruzenhack: I've created the component to improve the semantics of the code, we also have the `WarnTextValue`.

src/components/WalletConnect/theme.js
```diff
@@ -0,0 +1,71 @@
...
+export const commonStyles = StyleSheet.create({
+  // Card
+  card: {
+    paddingVertical: 16,
+    paddingHorizontal: 16,
+    backgroundColor: COLORS.backgroundColor,
+    borderTopLeftRadius: 16,
+    borderTopRightRadius: 16,
+    borderBottomLeftRadius: 16,
+    borderBottomRightRadius: 16,
```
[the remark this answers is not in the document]
alexruzenhack: You right, we can use `borderRadius` but I prefer to be explicit, especially because we can add variations in the card component over time and the borders may become asymmetric.

src/components/WalletConnect/theme.js
```diff
@@ -0,0 +1,71 @@
...
+  // General
+  text: {
+    fontSize: 14,
+    lineHeight: 20,
+    color: COLORS.black,
+  },
+  bold: {
+    fontWeight: 'bold',
+  },
+  value: {
```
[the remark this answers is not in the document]
alexruzenhack: This is a "qualifier" for the text component. It means a text that delivers a "value" from an object, not only information or instruction. See its usage at [NanoContractActions](https://github.com/HathorNetwork/hathor-wallet-mobile/pull/495/files#diff-d66e90db765ee3efa25024efcfc80ebfe4a82e1e983761f0226b2ac122bfe908R71). Maybe the is a poor name.

src/actions.js
```diff
@@ -175,6 +188,8 @@ export const types = {
   FIRSTADDRESS_SUCCESS: 'FIRSTADDRESS_SUCCESS',
   /* It signals a fetch failure due to an error. */
   FIRSTADDRESS_FAILURE: 'FIRSTADDRESS_FAILURE',
+  /* It updates the redux state of new nano contract transaction status on wallet connect register. */
+  WALLETCONNECT_NEW_NANOCONTRACT_STATUS: 'WALLETCONNECT_NEW_NANOCONTRACT_STATUS',
```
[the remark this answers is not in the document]
alexruzenhack: Sure, I believe `WALLET_CONNECT` was chosen by @andreabadesso, but I [proposed a taxonomy](https://github.com/HathorNetwork/hathor-wallet-mobile/issues/334) in this issue and I'm following it.

src/components/WalletConnect/NanoContract/NanoContractActions.js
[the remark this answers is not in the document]
alexruzenhack: A large list of actions would just expand the component without create a scroll context in the child component, which is what I was expecting, and that's good. See:

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
alexruzenhack: Never mind. The user will see a feedback message instructing it to select an address for the transaction in case the first address request fails.

src/sagas/wallet.js
alexruzenhack: Never mind. I refactored it and the `metadataLoaded` is not being used anymore. I implement now a flag `knownTokens.isLoading` to evaluate those flags and control component rendering.

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
alexruzenhack: You right! I refactored it to use the right API.

src/components/WalletConnect/theme.js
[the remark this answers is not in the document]
pedroferreira1: To be honest I still don't understand the meaning of this, even looking into the code. 

> Maybe the is a poor name.

Maybe you are correct, it's not clear. What other name do you suggest to improve this?

src/actions.js
```diff
@@ -113,6 +115,9 @@ export const types = {
   SET_WALLET_CONNECT: 'SET_WALLET_CONNECT',
   SET_WALLET_CONNECT_MODAL: 'SET_WALLET_CONNECT_MODAL',
   SET_WALLET_CONNECT_SESSIONS: 'SET_WALLET_CONNECT_SESSIONS',
+  WALLET_CONNECT_ACCEPT: 'WALLET_CONNECT_ACCEPT',
+  WALLET_CONNECT_REJECT: 'WALLET_CONNECT_REJECT',
+  SET_NEW_NANO_CONTRACT_TRANSACTION: 'SET_NEW_NANO_CONTRACT_TRANSACTION',
```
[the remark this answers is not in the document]
pedroferreira1: Yes but down below you have `WALLETCONNECT*` subjects, so even if you don't want to do the refactor right now, I believe all action types related to wallet connect should be together.

locale/pt-br/texts.po
```diff
@@ -1336,85 +1353,222 @@ msgstr ""
...
+msgid "Stay vigilant and protect your data from potential phishing attempts."
+msgstr "Fique atento e proteja seus dados de potenciais tentativas de phising."
+
+#: src/components/WalletConnect/NanoContract/DeclineModal.js:26
+msgid "Decline transaction"
+msgstr "Recusar transação"
+
+#: src/components/WalletConnect/NanoContract/DeclineModal.js:29
+msgid "Are you sure you want to decline this transaction?"
+msgstr "Está certo de recursar esta transação?"
```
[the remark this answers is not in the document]
pedroferreira1: This is still wrong: `recursar`

src/sagas/walletConnect.js
```diff
@@ -517,6 +524,45 @@ export function* onSessionDelete(action) {
...
+    try {
+      const { tokenInfo: { symbol, name } } = yield call([wallet, wallet.getTokenDetails], uid);
+      const token = { uid, symbol, name };
+      tokens[uid] = token;
+    } catch (e) {
+      log.error(`Fail getting token data for token ${uid}.`, e);
+      someError = true;
+      // continue
+    }
+  }
```
andreabadesso, line 559: You can use `all` to make the requests in parallel, also here goes a suggestion for legibility:

```suggestion
const tokenDetailsRequests = uids.map((uid) =>
  call(function* () {
    try {
      const { tokenInfo: { symbol, name } } = yield call([wallet, wallet.getTokenDetails], uid);

      return [false, { uid, symbol, name }];
    } catch (e) {
      log.error(`Fail getting token data for token ${uid}.`, e);
      someError = true;
      return [true, null];
    }
  })
);

const [someError, tokenDetails] = (yield all(tokenDetailsRequests)).reduce(
  (acc, tokenDetailRequest) => {
    const [error, tokenObj] = tokenDetailRequest;
    const [lastError, tokenDetails] = acc;

    return [
      lastError || error,
      tokenDetail ? {
        ...tokenDetails,
        [tokenDetail.uid]: tokenDetail
      } : tokenObj,
    ];
  }, [false, {}]
);
```

Also, can you move this code to `src/sagas/tokens.js` and call it here?

andreabadesso (changes requested):

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
```diff
@@ -0,0 +1,317 @@
...
+        <FeedbackModal
+          icon={(<Image source={checkIcon} style={styles.feedbackModalIcon} resizeMode='contain' />)}
+          text={t`Transaction successfully sent.`}
+          onDismiss={onFeedbackModalDismiss}
+          action={(<NewHathorButton discrete title={t`Ok, close`} onPress={onNavigateToDashboard} />)}
+        />
+      )}
+      {isTxFailed() && (
+        <FeedbackModal
+          icon={(<Image source={errorIcon} style={styles.feedbackModalIcon} resizeMode='contain' />)}
```
andreabadesso, line 275: `styles.feedbackModalIcon` is not defined

andreabadesso (changes requested):

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
```diff
@@ -0,0 +1,317 @@
...
+  const notRegistered = notInitialize && registeredNc == null;
+  // It results in true for registered nc and initialize request
+  const showRequest = !notRegistered;
+
+  // This effect should run only once because firstAddress is kept on state when loaded
+  useEffect(() => {
+    if (firstAddress.address) {
+      setNcAddress(firstAddress.address);
+    }
+  }, [firstAddress]);
```
pedroferreira1, line 127: I didn't understand this `useEffect`. If it's a registered nano, shouldn't we have the registered address as the `ncAddress`?

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
```diff
@@ -0,0 +1,317 @@
...
+    navigation.navigate('Dashboard');
+  };
+
+  const onTryAgain = () => {
+    dispatch(setNewNanoContractStatusReady());
+  };
+
+  const isTxInfoLoading = () => knownTokens.isLoading;
+  const isTxInfoLoaded = () => (
+    !knownTokens.isLoading && newTxStatus !== WALLETCONNECT_NEW_NANOCONTRACT_TX_STATUS.LOADING
```
pedroferreira1, line 168:
```suggestion
    !isTxInfoLoading() && newTxStatus !== WALLETCONNECT_NEW_NANOCONTRACT_TX_STATUS.LOADING
```

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
```diff
@@ -0,0 +1,317 @@
...
+  );
+  const isTxProcessing = () => (
+    !knownTokens.isLoading && newTxStatus === WALLETCONNECT_NEW_NANOCONTRACT_TX_STATUS.LOADING
```
pedroferreira1, line 171:
```suggestion
    !isTxInfoLoading() && newTxStatus === WALLETCONNECT_NEW_NANOCONTRACT_TX_STATUS.LOADING
```

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
alexruzenhack: Added.

src/actions.js
```diff
@@ -1205,3 +1246,114 @@ export const firstAddressFailure = (failurePayload) => ({
   type: types.FIRSTADDRESS_FAILURE,
   payload: failurePayload,
 });
+
+/**
+ * Request the download of token metadata for a list of tokens.
+ * @param {string[]} tokens A list of token uid
+ */
+export const tokensFetchMetadataRequest = (tokens) => ({
```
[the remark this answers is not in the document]
alexruzenhack: No, removed.

src/sagas/wallet.js
```diff
@@ -350,12 +350,20 @@ export function* loadTokens() {
   return registeredUids;
 }
 
+/**
+ * It handles the request to fetch tokens metadata and open a thread to process it.
+ * @param {{ tokens: string[] }}
+ */
+export function* fetchTokensMetadataRequest(action) {
```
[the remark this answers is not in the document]
alexruzenhack: No, removed.

src/sagas/walletConnect.js
```diff
@@ -516,6 +527,48 @@ export function* onSessionDelete(action) {
...
+  let someError = false;
+  for (const uid of uids) {
+    try {
+      const { tokenInfo: { symbol, name } } = yield call([wallet, wallet.getTokenDetails], uid);
+      const token = { uid, symbol, name };
+      tokens[uid] = token;
+    } catch (e) {
+      log.error(`Fail getting token data for token ${uid}.`, e);
+      someError = true;
+      // continue
```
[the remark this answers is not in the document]
alexruzenhack: I don't want to interrupt the loop. I want it to pass by all itens but flag if there was an error during some token details download.

src/components/WalletConnect/NanoContract/NanoContractMethodArgs.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  const blueprintInfo = useSelector((state) => state.nanoContract.blueprint[blueprintId]);
+  const argEntries = useMemo(() => {
+    const _methodInfo = get(blueprintInfo?.data, `public_methods.${method}`, null);
+    if (_methodInfo) {
+      return ncArgs.map((arg, idx) => [_methodInfo.args[idx].name, arg]);
+    }
+
+    return ncArgs.map((arg, idx) => [t`Position ${idx}`, arg]);
```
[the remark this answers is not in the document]
alexruzenhack: No, I don't believe we should throw an error here, but we can fire an error to sentry.

src/reducers/reducer.js
```diff
@@ -1683,3 +1828,199 @@ export const onFirstAddressSuccess = (state, { payload }) => ({
...
+
+/**
+ * @param {Object} state
+ * @param {{
+ *   payload: {
+ *     id: string;
+ *   };
+ * }} action
+ */
+export const onNanoContractBlueprintInfoReady = (state, { payload }) => ({
```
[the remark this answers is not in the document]
alexruzenhack: It is not need. Removed.

src/components/WalletConnect/NanoContract/NanoContractActions.js
```diff
@@ -0,0 +1,168 @@
...
+              <Text style={styles.action}>{getActionTitle(tokens, item)}</Text>
+              {item.address
+                && (
+                  <View>
+                    <Text style={styles.valueLabel}>{t`To Address:`}</Text>
+                    <Text style={styles.value}>{item.address}</Text>
+                  </View>
+                )}
+            </View>
+            <Amount amount={item.amount} isNft={false} />
```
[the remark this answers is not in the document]
alexruzenhack: You right. Added the proper check for NFT.

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
alexruzenhack: Yes, you right. I've added another condition for this assignment, the method should be `initialize`. Despite we have a the firstAddress as the default value for when the component is initializing, this state may not be initialized yet. This effect capture the change in the addressFirst when it is requested to load on `NanoContractExecInfo` child component.

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
alexruzenhack: Done.

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
alexruzenhack: Done.

src/components/WalletConnect/NanoContract/NanoContractMethodArgs.js
```diff
@@ -0,0 +1,94 @@
...
+  }, [blueprintInfo]);
+
+  return (
+    <View>
+      <View>
+        <Text style={commonStyles.sectionTitle}>{t`Argument List`}</Text>
```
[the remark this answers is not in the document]
alexruzenhack: Done.

src/sagas/walletConnect.js
```diff
@@ -100,6 +105,11 @@ const ERROR_CODES = {
 const Core = class {};
 const Web3Wallet = class {};
 
+const failureMessage = {
+  walletNotReadyError: t`Wallet is not ready yet to process a Wallet Connect effect.`,
+  someTokensNotLoaded: t`Some token details couldn't have been loaded.`,
```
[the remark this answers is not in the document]
alexruzenhack: Done.

src/sagas/walletConnect.js
alexruzenhack: Oh, thank you! Tbh, I have to think more about it. It involves network calls and may have some hidden pitfalls. I have to test it first. I have had some problems with "parallelism" using promises before. Lek me do some tests first. I'm opening this issue #531 to work on it later.

src/components/NanoContract/NanoContractTransactionHeader.js
```diff
@@ -20,6 +20,7 @@ import { COLORS } from '../../styles/themes';
 import { combineURLs, getShortContent, getShortHash, getTimestampFormat } from '../../utils';
 import SimpleButton from '../SimpleButton';
 import { ArrowDownIcon } from '../Icons/ArrowDown.icon';
+import { ArrowUpIcon } from '../Icons/ArrowUp.icon';
```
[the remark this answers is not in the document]
alexruzenhack: Done.

src/components/WalletConnect/NanoContract/DappContainer.js
```diff
@@ -0,0 +1,84 @@
...
+        <Image
+          source={{ uri: dapp.icon }}
+          width={48}
+          height={48}
+          style={styles.avatarIcon}
+        />
+      </View>
+      <View>
+        <Text style={styles.proposer}>{dapp.proposer}</Text>
+        <Text style={styles.network}>{'• '}{'mainnet'}</Text>
```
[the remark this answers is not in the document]
alexruzenhack: Done.

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
alexruzenhack: Done.

src/components/WalletConnect/NanoContract/NewNanoContractTransactionRequest.js
alexruzenhack: Removed.

src/sagas/wallet.js
alexruzenhack: Fixed.

locale/pt-br/texts.po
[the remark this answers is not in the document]
alexruzenhack: Fixed.

src/components/WalletConnect/theme.js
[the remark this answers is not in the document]
alexruzenhack: I refactored it to `field`.

src/sagas/walletConnect.js
```diff
@@ -516,6 +527,48 @@ export function* onSessionDelete(action) {
   yield call(onCancelSession, action);
 }
 
+/**
+ * Request tokens data to feed walletConnect's tokens.
+ * @param {Object} action
+ * @param {Object} action.payload
+ * @param {string[]} action.payload.uids
+ */
+export function* requestTokens(action) {
```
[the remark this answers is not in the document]
alexruzenhack: I've moved it to `saga/tokens`.

src/components/WalletConnect/NanoContract/NanoContractMethodArgs.js
```diff
@@ -0,0 +1,141 @@
...
+  const blueprintInfo = useSelector((state) => state.nanoContract.blueprint[blueprintId]);
+  // It results a in a list of entries like:
+  // >>> [['oracle_script', 'abc'], ['token_uid', '00'], ['date_last_bet', 123]]
+  // or a fallback like:
+  // >>> [['Position 0', 'abc'], ['Position 1', '00'], ['Position 2', 123]]
+  const argEntries = useMemo(() => {
+    // Retrieve the fallback entries while blueprint info is being loading.
+    if (blueprintInfo == null || blueprintInfo.status === STATUS.LOADING) {
+      return getFallbackArgEntries(ncArgs);
+    }
```
pedroferreira1, line 74: I don't think I like the fallback to be used while data is loading, only in case there's an error. I think I prefer a loading instead of having Position 0, then change it to the argument name.
What do you think?

alexruzenhack: Oh! I've removed it and now the screen keeps loading until the download of blueprint information.

src/components/WalletConnect/NanoContract/NanoContractMethodArgs.js
```diff
@@ -0,0 +1,154 @@
...
+  // Empty while downloading the bleuprint details
+  const isEmpty = argEntries.length === 0;
+  const notEmpty = !isEmpty;
+
+  return (
+    <View>
+      <View>
+        <Text style={commonStyles.sectionTitle}>{t`Arguments`}</Text>
+      </View>
+      {isEmpty /* This is a reduhdancy to the general loading */
```
pedroferreira1, line 100:
```suggestion
      {isEmpty /* This is a redundancy to the general loading */
```

src/components/WalletConnect/NanoContract/NanoContractMethodArgs.js
```diff
@@ -0,0 +1,154 @@
...
+    // Check any change in the lib or in the fullnode that could cause an impact here.
+    const errMsg = 'Error while getting the argument names of public_methods on blueprint';
+    dispatch(onExceptionCaptured(new Error(errMsg), false));
+
+    // Still render a fallback
+    return getFallbackArgEntries(ncArgs);
+  }, [method, ncArgs, blueprintInfo]);
+
+  // Empty while downloading the bleuprint details
+  const isEmpty = argEntries.length === 0;
```
pedroferreira1, line 92: We might have a method without arguments, so you can't show the feedback modal always when the argEntries is empty. You should check the loading status.

alexruzenhack: Oh! At the beginning of the method we have this:
```js
export const NanoContractMethodArgs = ({ blueprintId, method, ncArgs }) => {
  if (!ncArgs.length) {
    return null;
  }
```

It means we don't render the component if the transaction doesn't have arguments to call.

src/components/WalletConnect/NanoContract/NanoContractMethodArgs.js
alexruzenhack: Done.

src/components/FeedbackContent.js
```diff
@@ -65,6 +80,15 @@ const styles = StyleSheet.create({
     shadowColor: COLORS.textColor,
     shadowOpacity: 0.08,
   },
+  offMargin: {
+    marginTop: 0,
+    marginBottom: 0,
+    marginRight: 0,
+    marginLeft: 0,
```
andreabadesso, line 87:
```suggestion
    margin: 0,
```

src/screens/WalletConnect/NewNanoContractTransactionScreen.js
andreabadesso: You're right, just checked their documentation and it's preferable to get it from props whenever possible

src/sagas/walletConnect.js
andreabadesso: I don't think you'll have any problems here, even with a huge number of tokens

React-native's native HTTP module will queue them for you and only resolve when the request is done

But I agree that this should be tested, so OK to doing it in another PR

andreabadesso (approved):

src/components/FeedbackContent.js
alexruzenhack: It is not given that `margin: 0` will work here, see this issue I have opened about _flatten_ styles in React Native: https://github.com/HathorNetwork/hathor-wallet-mobile/issues/532

pedroferreira1 (approved):

PR #497, opened 2024-06-11
bump: current-project to 1.0.1

alexruzenhack:
### Acceptance Criteria
- This bump allows the same iOS version to be published

### Checklist
- [x] Make sure you updated the `CURRENT_PROJECT_VERSION` with the appropriate release-candidate version in `ios/HathorMobile.xcodeproj/project.pbxproj`
- [ ] Make sure you updated the `MARKETING_VERSION` with the appropriate version in `ios/HathorMobile.xcodeproj/project.pbxproj`
- [ ] Make sure you updated the version attribute in `package.json`
- [ ] Make sure you updated the version attribute in `package-lock.json`
- [ ] Make sure you incremented the `versionCode` attribute in `android/app/build.gradle`
- [ ] Make sure you updated the `versionName` with the appropriate version, including the release candidate number in `android/app/build.gradle`


ios/HathorMobile.xcodeproj/project.pbxproj
```diff
@@ -512,7 +512,7 @@
 				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				CLANG_ENABLE_MODULES = YES;
 				CODE_SIGN_ENTITLEMENTS = HathorMobile/HathorMobile.entitlements;
-				CURRENT_PROJECT_VERSION = 1.0.0;
+				CURRENT_PROJECT_VERSION = 1.0.1;
```
tuliomir, line 515: thought: I'm not sure this would work. The way we use the `PROJECT_VERSION` property is only by manipulating the major and  versions: one for each _release candidate_ version. For example: `rc.4` would have a project version `0.4.0`.

When we have a stable release we use the `1.0.0`.
So in this case the `1.0.1` would have no meaning and should not be used.

My suggestion is that we update the full application to a new patch: `v0.27.1`. This would revert the `CURRENT_PROJECT_VERSION` to `1.0.0` as it will be a new stable.

PR #505, opened 2024-07-04
chore: upgrade wallet lib to v1.8.0

r4mmer:
### Acceptance Criteria

- Bump wallet-lib v1.8.0
- Fix instances of any breaking changes, i.e. uses of `HATHOR_TOKEN_CONFIG`

### Security Checklist

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -37,7 +37,11 @@ export const _IS_MULTI_TOKEN = true;
  *   uid: '00'
  * }
  */
-export const _DEFAULT_TOKEN = hathorLib.constants.HATHOR_TOKEN_CONFIG;
+export const _DEFAULT_TOKEN = {
+  name: 'Hathor',
+  symbol: 'HTR',
+  uid: '00',
+};
 
 /**
  * App's primary color (Hathor purple)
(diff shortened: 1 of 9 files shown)


src/config.js
```diff
@@ -37,7 +37,11 @@ export const _IS_MULTI_TOKEN = true;
  *   uid: '00'
  * }
  */
-export const _DEFAULT_TOKEN = hathorLib.constants.HATHOR_TOKEN_CONFIG;
+export const _DEFAULT_TOKEN = {
+  name: 'Hathor',
+  symbol: 'HTR',
+  uid: '00',
+};
```
pedroferreira1, line 44: Why don't we use the default native token config from the lib here?

r4mmer: We can, but since this `_DEFAULT_TOKEN` is meant as a build parameter for a custom wallet I thought we should have the fields explicitly stated here instead of using constants from a dependency.
What do you think?

alexruzenhack: I think we can blend the solution, we can declare a literal object with references to native token from the lib.

alexruzenhack (approved):

pedroferreira1 (approved):

PR #506, opened 2024-07-04
fix: cleaning of tokens and nano-contract registered after reload

alexruzenhack:
### Acceptance Criteria
- It should set nanoContract redux state to its initial state after reload be requested

#### Changing network settings after register nano-contract

When we change the network settings, all registered tokens and nano-contracts are cleaned from storage, therefore the redux state should also resumes to its initial state.

https://github.com/HathorNetwork/hathor-wallet-mobile/assets/5992210/a3a1ed83-20c1-4598-93ab-7040f8bddeca

### Security Checklist

### DIFF
--- a/src/reducers/reducer.js
+++ b/src/reducers/reducer.js
@@ -1222,7 +1222,13 @@ export const onExceptionCaptured = (state, { payload }) => {
 };
 
 /**
- * On wallet reload, tokens data will be reloaded as well.
+ * On wallet reload, tokens and nano contract  data will be
+ * reloaded as well.
+ *
+ * Some flows can request a wallet reload such as:
+ *   - Disable wallet service
+ *   - Change network settings
+ *   - Loose network connection
  */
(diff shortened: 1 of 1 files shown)


andreabadesso (approved):

src/reducers/reducer.js
```diff
@@ -1229,6 +1229,7 @@ export const onReloadWalletRequested = (state) => ({
   tokensHistory: initialState.tokensHistory,
   tokensBalance: initialState.tokensBalance,
   loadHistoryStatus: initialState.loadHistoryStatus,
+  nanoContract: initialState.nanoContract,
```
pedroferreira1, line 1232: I think we should improve this docstring. I thought this reload was for any wallet reload, e.g. when the wallet loses internet connection. But this is only after you connect to a new network settings, right? I think this should be explicit in the docstring

pedroferreira1 (approved):

PR #431, opened 2024-02-16
fix: token duplication on dashboard

alexruzenhack:
### Issue
Closes: https://github.com/HathorNetwork/internal-issues/issues/242

@pedroferreira1 has investigated the these of migration be the culprit and has confirmed it. See the [investigation note](https://github.com/HathorNetwork/internal-issues/issues/242#issuecomment-1971507971).

Despite we didn't need a fix anymore, we will proceed to avoid this problem in the future.

### Acceptance Criteria
- Should fix token duplication by replacing token list to token map

The following recording shows operations with token, register, unregister and creation to demonstrate these operations were not effected by the fix/refactoring:

<video src="https://github.com/HathorNetwork/hathor-wallet-mobile/assets/5992210/4b03af1c-6d0c-451d-98d1-6846c2040163"></video>

### Security Checklist
- [x] Make sure you do not include new dependencies in the project unless strictly necessary and do not include dev-dependencies as production ones. More dependencies increase the possibility of one of them being hijacked and affecting us.

PR #507, opened 2024-07-11
fix: Unable to send token by reading a QR code

alexruzenhack:
### Context

I have fixed token duplication on the Dashboard screen by transforming the registered tokens list into a map, this way we would never have a duplication. However, I failed to cover the necessary change and left one case untouched, which caused this bug.

See the previous fix at: https://github.com/HathorNetwork/hathor-wallet-mobile/pull/431

### Acceptance Criteria
- It should restore the ability to read QR code when sending token

Closes: https://github.com/HathorNetwork/internal-issues/issues/345

### Security Checklist

### DIFF
--- a/src/screens/SendScanQRCode.js
+++ b/src/screens/SendScanQRCode.js
@@ -57,10 +57,7 @@ class SendScanQRCode extends React.Component {
     if (!qrcode.isValid) {
       this.showAlertError(qrcode.error);
     } else if (qrcode.token && qrcode.amount) {
-      const isTokenRegistered = this.props.tokens.some(
-        (stateToken) => stateToken.uid === qrcode.token.uid
-      );
-      if (isTokenRegistered) {
+      if (qrcode.token.uid in this.props.tokens) {
         const params = {
           address: qrcode.address,
           token: qrcode.token,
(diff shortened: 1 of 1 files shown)


andreabadesso (approved):

src/screens/SendScanQRCode.js
```diff
@@ -57,10 +57,7 @@ class SendScanQRCode extends React.Component {
     if (!qrcode.isValid) {
       this.showAlertError(qrcode.error);
     } else if (qrcode.token && qrcode.amount) {
-      const isTokenRegistered = this.props.tokens.some(
-        (stateToken) => stateToken.uid === qrcode.token.uid
-      );
-      if (isTokenRegistered) {
+      if (qrcode.token.uid in this.props.tokens) {
```
tuliomir, line 60: suggestion(non-blocking): Add a comment explaining that this is a search into the registered tokens.

Otherwise the reader would have to search what the `props.tokens` mean to understand what this rule is about. This was made clearer before by the old variable name itself.

tuliomir (approved):

pedroferreira1 (approved):

PR #512, opened 2024-07-22
fix: nano-contract history pagination and list order

alexruzenhack:
### Acceptance Criteria
- It should fix nano contract transaction loading for newer transactions
- Add 'Load More' in the footer of transactions list, which triggers the request for older transactions when pressed
  - Hides the 'Load More' button when the last transaction of the list is the 'initialize'

#### Loading newer transactions

https://github.com/user-attachments/assets/781e6759-6db2-4e81-b3bc-c20641c097ed

#### Loading older transactions

https://github.com/user-attachments/assets/68f99f4e-b8be-4cf2-a02f-63f1268bfa18

The video above shows a list of transactions initialized with the newest transaction only. When the button 'Load More' is pressed a list of older transactions is loaded. The button 'Load More' didn't show after load because the list has reached its end.

---

Related issues:
- [ ] https://github.com/HathorNetwork/hathor-wallet-mobile/issues/514

---

### Security Checklist

### DIFF
--- a/src/components/NanoContract/NanoContractDetails.js
+++ b/src/components/NanoContract/NanoContractDetails.js
@@ -33,7 +33,6 @@ import { FeedbackContent } from '../FeedbackContent';
  *   txHistory: Object[];
  *   isLoading: boolean;
  *   error: string;
- *   after: string;
  * }}
  */
 const getNanoContractDetails = (ncId) => (state) => {
@@ -42,14 +41,13 @@ const getNanoContractDetails = (ncId) => (state) => {
    * and let it step aside while coming back to Dashboard screen. This transition happens
    * quickly, therefore the user will not have time to see the default state.
    */
-  const defaultMeta = { isLoading: false, error: null, after: null };
+  const defaultMeta = { isLoading: false, error: null };
   const txHistory = state.nanoContract.history[ncId] || [];
-  const { isLoading, error, after } = state.nanoContract.historyMeta[ncId] || defaultMeta;
+  const { isLoading, error } = state.nanoContract.historyMeta[ncId] || defaultMeta;
   return {
     txHistory,
     isLoading,
     error,
-    after,
   };
 }
 
@@ -65,7 +63,11 @@ export const NanoContractDetails = ({ nc }) => {
   const dispatch = useDispatch();
   const navigation = useNavigation();
 
-  const { txHistory, isLoading, error, after } = useSelector(getNanoContractDetails(nc.ncId));
+  const {
+    txHistory,
+    isLoading,
+    error,
+  } = useSelector(getNanoContractDetails(nc.ncId));
   const [ncAddress, changeNcAddress] = useState(nc.address);
 
   const onAddressChange = (newAddress) => {
@@ -77,6 +79,7 @@ export const NanoContractDetails = ({ nc }) => {
     navigation.navigate('NanoContractTransactionScreen', { tx });
   };
 
+  // This effect runs only once when the component is first built.
   useEffect(() => {
     if (txHistory.length === 0) {
       /* The first time we load the Nano Contract details its transaction history is empty.
@@ -85,20 +88,18 @@ export const NanoContractDetails = ({ nc }) => {
        * For the first transaction history load we don't need to specify the `after` param,
        * it will be set during the load.
        */
-      dispatch(nanoContractHistoryRequest({ ncId: nc.ncId, after: null }));
+      dispatch(nanoContractHistoryRequest({ ncId: nc.ncId }));
     } else {
-      dispatch(nanoContractHistoryRequest({ ncId: nc.ncId, after }))
+      // Fetch new transactions when there are some transactions in the history.
+      dispatch(nanoContractHistoryRequest({ ncId: nc.ncId, before: txHistory[0].txId }));
     }
   }, []);
 
-  const handleMoreTransactions = () => {
-    if (after == null) {
-      /* This situation is unlikely to happen because on the first transactions history load
-       * the `after` is assigned with the hash of the last transaction in the list.
-       */
-      return;
-    }
-    dispatch(nanoContractHistoryRequest({ ncId: nc.ncId, after }));
+  /**
+   * Triggered when a user makes the pull gesture on the transaction history content.
+   */
+  const handleNewTransactions = () => {
+    dispatch(nanoContractHistoryRequest({ ncId: nc.ncId, before: txHistory[0].txId }));
   };
 
   /* If an error happens on loading transactions history, an error feedback
@@ -135,7 +136,7 @@ export const NanoContractDetails = ({ nc }) => {
             )}
             keyExtractor={(item) => item.txId}
             refreshing={isLoading}
-            onRefresh={handleMoreTransactions}
+            onRefresh={handleNewTransactions} // pull gesture
           />
         )}
     </Wrapper>
--- a/src/components/NanoContract/NanoContractDetails.js
+++ b/src/components/NanoContract/NanoContractDetails.js
@@ -137,7 +137,10 @@ export const NanoContractDetails = ({ nc }) => {
             )}
             keyExtractor={(item) => item.txId}
             refreshing={isLoading}
-            onRefresh={handleNewerTransactions} // pull gesture
+            // Enables the pull gesture to get newer transactions
+            onRefresh={handleNewerTransactions}
+            // Enables a button to load more of older transactions until the end
+            // By reaching the end, the button ceases to render
             ListFooterComponent={<LoadMoreButton lastTx={txHistory.slice(-1,).pop()} />}
             extraData={[isLoading, error]}
           />
--- a/src/components/NanoContract/NanoContractDetails.js
+++ b/src/components/NanoContract/NanoContractDetails.js
@@ -150,7 +150,7 @@ export const NanoContractDetails = ({ nc }) => {
 };
 
 /**
- * It show a button to 'Load More' transactions after the last one.
+ * It shows a button to 'Load More' transactions after the last one.
  * It hides the button when the last transaction is the initialize.
  *
  * @param {Object} prop Properties object
--- a/src/components/NanoContract/NanoContractDetails.js
+++ b/src/components/NanoContract/NanoContractDetails.js
@@ -23,6 +23,7 @@ import Spinner from '../Spinner';
 import errorIcon from '../../assets/images/icErrorBig.png';
 import SimpleButton from '../SimpleButton';
 import { FeedbackContent } from '../FeedbackContent';
+import NewHathorButton from '../NewHathorButton';
 
 /**
  * Retrieves Nano Contract details from Redux.
@@ -98,7 +99,7 @@ export const NanoContractDetails = ({ nc }) => {
   /**
    * Triggered when a user makes the pull gesture on the transaction history content.
    */
-  const handleNewTransactions = () => {
+  const handleNewerTransactions = () => {
     dispatch(nanoContractHistoryRequest({ ncId: nc.ncId, before: txHistory[0].txId }));
   };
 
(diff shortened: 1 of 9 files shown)


src/sagas/nanoContract.js
```diff
@@ -172,21 +171,37 @@ export function* registerNanoContract({ payload }) {
  * @param {string} after Transaction hash to start to get items
  * @param {Object} wallet Wallet instance from redux state
  *
+ * @description
+ * We use `after` because we look at time from present to future, therefore `after` means
```
andreabadesso, line 175: I don't get it... We are actually fetching transactions "before" a given transaction as the user "infinite scrolls" deeper, shouldn't we name the variable `before`?

src/sagas/nanoContract.js
```diff
@@ -172,21 +171,37 @@ export function* registerNanoContract({ payload }) {
...
   if (!success) {
     throw new Error('Failed to fetch nano contract history');
   }
-
-  const history = [];
-  for (const rawTx of rawHistory) {
+  // We are interested to produce a list of transactions in descending order
+  // because we want users to see newest txs first.
+  const historyReversed = new Array(rawHistory.length)
+  for (let idx = 0; idx < rawHistory.length; idx += 1) {
```
andreabadesso, line 203: suggestion(non-blocking): You could refactor this to use `Promise.all` so those multiple async requests would run concurrently

src/sagas/nanoContract.js
```diff
@@ -213,15 +228,18 @@ export async function fetchHistory(ncId, count, after, wallet) {
...
-    history.push(tx);
+    historyReversed[rawHistory.length - idx - 1] = tx;
   }
 
   let next = after;
-  if (history && history.length > 0) {
-    next = history[history.length - 1].txId;
+  if (historyReversed && historyReversed.length > 0) {
+    // We should get the first item of the list because we are looking
+    // from present -> past, and we want to paginate from preset toward future.
```
andreabadesso, line 237: I'm confused, we're actually paginating towards the past, right?

You scroll to see earlier transactions, i.e. the latest transaction is on top

andreabadesso (changes requested):

src/components/NanoContract/NanoContractDetails.js
```diff
@@ -86,6 +86,8 @@ export const NanoContractDetails = ({ nc }) => {
        * it will be set during the load.
        */
       dispatch(nanoContractHistoryRequest({ ncId: nc.ncId, after: null }));
+    } else {
+      dispatch(nanoContractHistoryRequest({ ncId: nc.ncId, after }))
```
andreabadesso, line 90: You should add `after` as a dependency for the `useEffect`

andreabadesso (changes requested):

src/sagas/nanoContract.js
alexruzenhack: Sorry for the confusion, let me try again:
When we call `fetchHistory` with `after` param this denotes that we want new transactions _after_ a certain tx, and we are looking from the present moment toward the future. This is a common knowledge of how to think in time, _after_ means the future.

{txs that we have (past)} **{after (the last tx that we have)}** {txs we are interested in (future)}

However the method `getNanoContractHistory` uses the following signature:
`getNanoContractHistory(ncId, count, after, before)`.

Here, `after` denotes all the transactions from a certain tx towards the **past**. I believe it has something to do with the way we walk the dag. However, this way to frame time is not like the way we usually do, and it can makes things really odd in my opinion, and that's why I'm swapping the words `before` to `after`. After all, we are interested in the future when we ask a new chunk of transactions.

src/sagas/nanoContract.js
alexruzenhack: But if we do this we can't use guarantee the order of execution, right? I need it ordered.

src/sagas/nanoContract.js
alexruzenhack: This is confusing, indeed. I believe I can remove this comment. We just need to assign the newest tx hash to `next`, that's it.

src/components/NanoContract/NanoContractDetails.js
alexruzenhack: If we do this, the `nanoContractHistoryRequest` will be dispatched twice. The other side effect would be the dispatch of `nanoContractHistoryRequest` until we react the newest tx, however it would happen without a user action, which is not good for usability. `txHistory` already provokes a new rendering if it change, I think.

src/sagas/nanoContract.js
pedroferreira1: The words `after` and `before` doesn't have any relation with future and past in this API. I think it's confusing to swap the words, we should use the same words as it's used in the API. `after` is the pagination parameter to get the history after a specific tx, and `before` the same. The API returns the transactions in the descending order, i.e., from the newest to the oldest, as it's expected in the screen.

src/sagas/nanoContract.js
andreabadesso: You could pass the index to the promise and return it like [index, response], then use it to assembly the array

But it's non-blocking at all, if you think it's not worth the effort, don't do it

alexruzenhack: Ok. I have opened an issue for it: https://github.com/HathorNetwork/hathor-wallet-mobile/issues/514

src/sagas/nanoContract.js
alexruzenhack: I have refactored it, can you take a new look?

src/components/NanoContract/NanoContractDetails.js
alexruzenhack: I have refactored it, can you take a new look?

src/sagas/nanoContract.js
alexruzenhack: Ok. I just refactored it. Can you both take a new look at it?

src/sagas/nanoContract.js
```diff
@@ -93,7 +93,10 @@ export function* registerNanoContract({ payload }) {
     return;
   }
 
-  const isAddressMine = yield call(wallet.isAddressMine.bind(wallet), address);
+  // XXX: Make sure to disable the Wallet Service feature if you are testing
+  // the nano-testnet network because Wallet Service Wallet doesn't implement
+  // `isAddressMine`.
```
andreabadesso, line 98: I think this should be done automatically as we might eventually want to control the nano-testnet in Unleash

Is there an issue for this? I don't think it should be done in this PR

andreabadesso: In this PR, you could just set it to false if the user is in the wallet-service, so it doesn't crash

src/sagas/nanoContract.js
```diff
@@ -167,31 +169,55 @@ export function* registerNanoContract({ payload }) {
...
+  const historyNewestToOldest = new Array(rawHistory.length)
+  for (let idx = 0; idx < rawHistory.length; idx += 1) {
+    const rawTx = rawHistory[idx];
     const network = wallet.getNetworkObject();
     const caller = addressUtils.getAddressFromPubkey(rawTx.nc_pubkey, network).base58;
     // XXX: Wallet Service Wallet doesn't implement isAddressMine.
     // It means this method can't run under wallet-service without
     // throwing an exception.
+    // XXX: Make sure to disable the Wallet Service feature if you are testing
+    // the nano-testnet network
```
andreabadesso, line 220: same here

src/sagas/nanoContract.js
```diff
@@ -263,24 +284,38 @@ export function* requestHistoryNanoContract({ payload }) {
...
+      count: NANO_CONTRACT_TX_HISTORY_SIZE,
+    };
+    const { history } = yield call(fetchHistory, req);
 
-    if (after != null) {
-      // The first load has always `after` equals null. The first load means to be fast,
-      // but the subsequent ones are all request by user and we want slow down multiple
-      // calls to this effect.
+    if (after != null || before != null) {
+      // We want slow down multiple calls to this effect.
```
andreabadesso, line 304: Why? If you dispatch multiple actions at the same time, they will still execute at the same time, just 1000ms later

src/components/NanoContract/NanoContractDetails.js
```diff
@@ -133,7 +136,7 @@ export const NanoContractDetails = ({ nc }) => {
             )}
             keyExtractor={(item) => item.txId}
             refreshing={isLoading}
-            onRefresh={handleMoreTransactions}
+            onRefresh={handleNewTransactions} // pull gesture
```
andreabadesso, line 139: Ok, so this is the pull to refresh, but where is the code for the infinite scroll?

andreabadesso (changes requested):

src/sagas/nanoContract.js
alexruzenhack: You right! I removed it.

src/sagas/nanoContract.js
alexruzenhack: I've opened an issue, see https://github.com/HathorNetwork/hathor-wallet-lib/issues/732.

src/components/NanoContract/NanoContractDetails.js
alexruzenhack: Sorry, I forgot to upload. Here it is: https://github.com/HathorNetwork/hathor-wallet-mobile/pull/512/commits/2040049f05e0125695c2ad5e612d54039da8b087

alexruzenhack:
>I missed this in the add component PR, but can you rename this component? It doesn't make sense to have a NewHathorButton, how is it different from the HathorButton? Can we merge both?
-- Comment by @andreabadesso on https://github.com/HathorNetwork/hathor-wallet-mobile/pull/515, which is now incorporated here.

I will open a KTLO for this.

src/sagas/nanoContract.js
```diff
@@ -167,33 +174,60 @@ export function* registerNanoContract({ payload }) {
...
-    // XXX: Wallet Service Wallet doesn't implement isAddressMine.
-    // It means this method can't run under wallet-service without
-    // throwing an exception.
-    // eslint-disable-next-line no-await-in-loop
-    const isMine = await wallet.isAddressMine(caller);
+    // XXX: Wallet Service doesn't implement isAddressMine.
+    // See issue: https://github.com/HathorNetwork/hathor-wallet-lib/issues/732
+    // Default to `false` if using Wallet Service.
+    let isMine = false;
+    const useWalletService = consumeGenerator(isWalletServiceEnabled());
```
andreabadesso, line 225: Why do you need that?

src/utils.js
```diff
@@ -415,6 +415,18 @@ export const getNanoContractFeatureToggle = (state) => (
  */
 export const getTimestampFormat = (timestamp) => moment.unix(timestamp).format(t`DD MMM YYYY [•] HH:mm`)
 
+/**
+ * Extract the result of a generator function when it is done.
+ */
+export const consumeGenerator = (generator) => {
```
andreabadesso, line 421: This is not "consuming" a generator, it's waiting until it's done and returning the last yielded value

Name is misleading

src/sagas/nanoContract.js
andreabadesso: Can't you just use

`const useWalletService = yield call(isWalletServiceEnabled);`

?

andreabadesso (changes requested):

src/utils.js
alexruzenhack: No, it is not returning the last yielded value, it is returning the return value of a generator after it has yield for the last time.

I believe it is "consuming" the generator because it is running the generator until its end.

src/sagas/nanoContract.js
alexruzenhack: The function isn't a generator, but a promise. As we can't have async generators the only option to call one until its end is by consuming it. But in this case I'm interested in the result of it, not the yield values.

src/components/NanoContract/NanoContractDetails.js
```diff
@@ -133,13 +137,53 @@ export const NanoContractDetails = ({ nc }) => {
...
+            ListFooterComponent={<LoadMoreButton lastTx={txHistory.slice(-1,).pop()} />}
+            extraData={[isLoading, error]}
           />
         )}
     </Wrapper>
   );
 };
 
+/**
+ * It show a button to 'Load More' transactions after the last one.
```
pedroferreira1, line 153:
```suggestion
 * It shows a button to 'Load More' transactions after the last one.
```

src/components/NanoContract/NanoContractDetails.js
```diff
@@ -133,13 +137,53 @@ export const NanoContractDetails = ({ nc }) => {
...
+ *
+ * @param {Object} prop Properties object
+ * @param {{
+ *   ncId: string;
+ *   txId: string;
+ * }} prop.lastTx A transaction item from transaction history
+ */
+const LoadMoreButton = ({ lastTx }) => {
+  const dispatch = useDispatch();
+  const isInitializeTx = lastTx.ncMethod === 'initialize';
```
pedroferreira1, line 164: You can continue using this to check if it already loaded the last tx, but now the API returns a new value in the response `has_more`

pedroferreira1 (approved):

src/components/NanoContract/NanoContractDetails.js
alexruzenhack: Got it. Well, I prefer the current way. But if you want, I can open a KTLO to adapt the pagination to use the `has_more` property.

PR #535, opened 2024-08-21
chore: upgrade actions to a version that use node20 (LTS)

r4mmer:
### Acceptance Criteria
- upgrade actions to a version that use node20 (LTS)

### Security Checklist

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,9 +20,9 @@ jobs:
       matrix:
         node-version: [14.x, 16.x]
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
     - name: Use Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -20,9 +20,9 @@ jobs:
       matrix:
         node-version: [14.x, 16.x]
```
andreabadesso, line 21:
```suggestion
        node-version: [18.x, 20.x]
```

tuliomir: Should we still use NodeJS 18 in the matrix? Our other repositories are using only version 20.

tuliomir: Should we still use NodeJS 18 in the matrix? Our other repositories are using only version 20.

pedroferreira1 (approved):

andreabadesso (approved):

PR #542, opened 2024-08-22
refactor: default network is nano testnet

andreabadesso:
### Acceptance Criteria
- We should include the nano testnet settings and default to it.
- We should allow the user to select the nano testnet in the network settings screen

### Security Checklist

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -14,6 +14,7 @@ import {
   _DEFAULT_TOKEN as DEFAULT_TOKEN,
   _PRIMARY_COLOR as PRIMARY_COLOR,
   _SENTRY_DSN as SENTRY_DSN,
+  DEFAULT_EVENT_TOKEN,
 } from './config';
 
 /**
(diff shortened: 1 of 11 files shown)


alexruzenhack (approved):

src/constants.js
```diff
@@ -190,6 +190,23 @@ export const EXPLORER_TESTNET_URL = 'https://explorer.testnet.hathor.network/';
...
+export const NETWORK_NANO_TESTNET = 'testnet';
+export const NODE_SERVER_NANO_TESTNET_URL = 'https://node1.nano-testnet.hathor.network/v1a/';
+export const EXPLORER_NANO_TESTNET_URL = 'https://explorer.alpha.nano-testnet.hathor.network/';
+export const TX_MINING_SERVICE_NANO_TESTNET_URL = 'https://txmining.nano-testnet.hathor.network/';
+
+export const PRE_SETTINGS_NANO_TESTNET = {
+  stage: STAGE_TESTNET,
+  network: NETWORK_NANO_TESTNET,
+  walletServiceUrl: WALLET_SERVICE_TESTNET_BASE_URL,
+  walletServiceWsUrl: WALLET_SERVICE_TESTNET_BASE_WS_URL,
```
pedroferreira1, line 203: Is it the same as our public testnet? I know that we won't allow Hathor Play to connect to wallet service but is it a required key?

pedroferreira1 (approved):

src/constants.js
andreabadesso: They're not required, I agree we should remove them, done

PR #543, opened 2024-08-22
refactor: skip directly to the ChoosePinScreen

andreabadesso:
### Acceptance Criteria
- We should skip the seed confirmation depending on a constant file in config

### Future possibilities

I think this could go to `master` (changing the default to `true`), as it is a nice option to have on whitelabel apps, but we would have to add a mechanism for display the seed after it's loaded

### Security Checklist

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -46,7 +46,7 @@ export const _DEFAULT_TOKEN = {
  * Whether we should skip the words confirmation screen during
  * new wallet creation
  */
-export const SKIP_SEED_CONFIRMATION = false;
+export const SKIP_SEED_CONFIRMATION = true;
 
 /**
  * App's primary color (Hathor purple)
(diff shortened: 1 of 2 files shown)


alexruzenhack (approved):

src/config.js
```diff
@@ -42,6 +42,12 @@ export const _DEFAULT_TOKEN = {
   ...hathorLib.constants.DEFAULT_NATIVE_TOKEN_CONFIG,
 };
 
+/**
+ * Whether we should skip the words confirmation screen during
+ * new wallet creation
+ */
+export const SKIP_SEED_CONFIRMATION = false;
```
pedroferreira1, line 49: This should be `true`, right?

andreabadesso: It should.

When I opened this PR was thinking on opening it to `master`, but I guess we can cherry-pick it in the future and set it to false again

andreabadesso: Fixed

pedroferreira1 (approved):

PR #546, opened 2024-08-23
refactor: remove nc status from tx details

alexruzenhack:
### Acceptance Criteria
- Remove Nano Contract status from Tx Details Modal

### Security Checklist

### DIFF
--- a/src/components/TxDetailsModal.js
+++ b/src/components/TxDetailsModal.js
@@ -16,6 +16,7 @@ import SlideIndicatorBar from './SlideIndicatorBar';
 import CopyClipboard from './CopyClipboard';
 import { PublicExplorerListButton } from './PublicExplorerListButton';
 import { COLORS } from '../styles/themes';
+import { TransactionStatusLabel } from './TransactionStatusLabel';
 
 class TxDetailsModal extends Component {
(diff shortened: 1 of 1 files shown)


andreabadesso (approved):

src/components/TxDetailsModal.js
```diff
@@ -105,18 +103,6 @@ class TxDetailsModal extends Component {
...
-                      <ListItem
-                        title={t`Nano Contract Status`}
-                        text={(
-                          <TransactionStatusLabel
-                            hasFirstBlock={hasFirstBlock}
-                            isVoided={isVoided}
-                          />
-                        )}
-                      />
-                    )}
```
pedroferreira1, line 119: I know I said to remove the status because it won't matter if the tx is confirmed by a block for Hathor Play but I think we should keep a status if the tx is voided, it seems important for the wallet user to know if the tx was invalidated somehow

alexruzenhack: Ok. I've added the status label only for voided tx: 37c14d0c29141a94c9959900587235ccd0b1c68c

pedroferreira1 (approved):

Issue #510, opened 2024-07-17
Request nano contract transaction state by other endpoint then `nano_contract/state`


alexruzenhack:
## Motivation

The endpoint `/nano_contract/state` returns data only after the transaction was executed, which makes its utilization in the registration process undesired.

## Resolution

We should use the `/transaction` endpoint when the transaction was not executed yet, and `/state` after it has been executed.

## Usage

- It is used in the effect `registerNanoContract` at `sagas/nanoContract.js`

## Issues

- [x] It is lacking `blueprint_name` in the transaction; without it, we are forced to make another request to get blueprint info.
  - We are going to call the blueprint info endpoint during registration and set it on store
  - Also, we are going to call the blueprint info endpoint every time we receive a new nano contract transaction, and set it on store
  - In both cases we first check if the blueprint info is already on store

## Tasks

- [x] Assess where the `/state` is used in the wallet-mobile

PR #549, opened 2024-08-27
refactor: replace nano contract state call for history call

alexruzenhack:
### Acceptance Criteria
- Replace call to nano contract state by other call to get nano contract information

This is important because the "/state" endpoint is not consistent as it can return "not found" while the new nano contract was not confirmed by a block.

Closes: https://github.com/HathorNetwork/hathor-wallet-mobile/issues/510

### Security Checklist

### DIFF
--- a/locale/pt-br/texts.po
+++ b/locale/pt-br/texts.po
@@ -57,7 +57,7 @@ msgid "[Last] dddd [•] HH:mm"
 msgstr "[Última] dddd [•] HH:mm"
 
 #.  See https://momentjs.com/docs/#/displaying/calendar-time/
-#: src/models.js:107 src/utils.js:427
+#: src/models.js:107 src/utils.js:426
 msgid "DD MMM YYYY [•] HH:mm"
 msgstr "DD MMM YYYY [•] HH:mm"
 
@@ -1058,7 +1058,7 @@ msgid "Nano Contract not found."
 msgstr "Nano Contract não encontrado"
 
 #: src/sagas/nanoContract.js:49
-msgid "Error while trying register Nano Contract."
+msgid "Error while trying to register Nano Contract."
 msgstr "Erro ao registrar Nano Contract."
 
 #: src/sagas/nanoContract.js:50
--- a/locale/pt-br/texts.po
+++ b/locale/pt-br/texts.po
@@ -57,7 +57,7 @@ msgid "[Last] dddd [•] HH:mm"
 msgstr "[Última] dddd [•] HH:mm"
 
 #.  See https://momentjs.com/docs/#/displaying/calendar-time/
-#: src/models.js:107 src/utils.js:426
+#: src/models.js:107 src/utils.js:427
 msgid "DD MMM YYYY [•] HH:mm"
 msgstr "DD MMM YYYY [•] HH:mm"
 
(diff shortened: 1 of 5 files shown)


src/sagas/nanoContract.js
```diff
@@ -111,9 +111,9 @@ export function* registerNanoContract({ payload }) {
     return;
   }
 
-  let ncState = null;
+  let ncHistory = null;
   try {
-    ncState = yield call([ncApi, ncApi.getNanoContractState], ncId);
+    ncHistory = yield call([ncApi, ncApi.getNanoContractHistory], ncId, 1);
```
pedroferreira1, line 116: I understand that you are using this to check if the nano contract exists but semantically I prefer doing this like the desktop wallet.

```
  try {
    const response = yield call([wallet, wallet.getFullTxById], ncId);
    tx = response.tx;
  } catch (error) {
    console.error('Error while registering Nano Contract.', error);
    yield put(nanoContractRegisterError(t`Error while registering the nano contract.`));
    return;
  }

  if (!nanoUtils.isNanoContractCreate(tx)) {
    console.debug('Transaction is not a nano contract creation transaction.');
    yield put(nanoContractRegisterError(t`Invalid nano contract creation transaction.`));
    return;
  }
```
https://github.com/HathorNetwork/hathor-wallet/blob/master/src/sagas/nanoContract.js

What do you think?

alexruzenhack: Replaced: 4b81d31958257d4050a6af348bf85023fc5ed77c

locale/pt-br/texts.po
```diff
@@ -57,7 +57,7 @@ msgid "[Last] dddd [•] HH:mm"
 msgstr "[Última] dddd [•] HH:mm"
 
 #.  See https://momentjs.com/docs/#/displaying/calendar-time/
-#: src/models.js:107 src/utils.js:426
+#: src/models.js:107 src/utils.js:427
```
andreabadesso, line 60: Why this line changed in `src/utils`?

locale/texts.pot
```diff
@@ -48,7 +48,7 @@ msgid "[Last] dddd [•] HH:mm"
 msgstr ""
 
 #: src/models.js:107
-#: src/utils.js:426
+#: src/utils.js:427
```
andreabadesso, line 51: Why this changed? You didn't change `src/utils`

src/sagas/nanoContract.js
```diff
@@ -45,7 +46,8 @@ export const failureMessage = {
   walletNotReadyError: t`Wallet is not ready yet to register a Nano Contract.`,
   addressNotMine: t`The informed address does not belong to the wallet.`,
   nanoContractStateNotFound: t`Nano Contract not found.`,
-  nanoContractStateFailure: t`Error while trying to get Nano Contract state.`,
+  nanoContractFailure: t`Error while trying register Nano Contract.`,
```
andreabadesso, line 49:
```suggestion
  nanoContractFailure: t`Error while trying to register Nano Contract.`,
```

andreabadesso (changes requested):

pedroferreira1 (approved):

locale/pt-br/texts.po
alexruzenhack: I ran an update, probably it was lacking an update.

locale/texts.pot
alexruzenhack: I ran an update, probably it was lacking an update.

src/sagas/nanoContract.js
alexruzenhack: Fixed: 003a3f8af07e1f5cb783427c7378624a816a9567

andreabadesso (approved):

PR #564, opened 2024-09-12
fix: wrong success navigation by forcing ready state on navigation

alexruzenhack:
### Acceptance Criteria
- Change state to ready when navigating to success screen

### Security Checklist


andreabadesso (approved):

tuliomir (approved):
question(non-blocking): I agree with the change, but what will be the effect of having the Redux store state updated if the user has already navigated away from the screen?

Wouldn't resetting the state be more helpful?
Maybe an explanation about the motivation on the PR description would help understanding that.

alexruzenhack:
> question(non-blocking): I agree with the change, but what will be the effect of having the Redux store state updated if the user has already navigated away from the screen?
> 
> Wouldn't resetting the state be more helpful? Maybe an explanation about the motivation on the PR description would help understanding that.

This is a kind of "reset". It reset the state to "ready" which is the starting state.

Issue #531, opened 2024-08-14
Refactor fetchUnregisteredTokens effect to use concurrent calls


alexruzenhack:
@andreabadesso has proposed the an implementation, see [this comment](https://github.com/HathorNetwork/hathor-wallet-mobile/pull/495#discussion_r1715712576). I have to test it before make some adjustments. One point of attention is the rate-limit on the API.

PR #565, opened 2024-09-12
feat: make requestUnregisteredTokens concurrent

alexruzenhack:
### Motivation

The user needs the data as quick as possible in a new nano contract request. To achieve this requirement we can make the http request concurrent, however we should respect the rate-limit rules set in public node endpoints.

### Acceptance Criteria
- Make requestUnregisteredTokens to call token details in "parallel"
- Add an action to end the request unregistered tokens process
- Add an utility function to split a list in groups

### Future possibility

We can implement a client object that implements a more sophisticated limit-rate strategy that adapts to back pressure from server. This control is interesting to pursuit in order to better rationalize the resources at app disposal and at the same time helps the server to heal in case of spike usage.

>[!NOTE]
>It closes https://github.com/HathorNetwork/hathor-wallet-mobile/issues/531.

>[!WARNING]
>To work properly it depends on the fix on promise leak, see https://github.com/HathorNetwork/hathor-wallet-lib/pull/772.

### Security Checklist

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -188,8 +188,13 @@ export const types = {
   FIRSTADDRESS_FAILURE: 'FIRSTADDRESS_FAILURE',
   /* It updates the redux state of new nano contract transaction status on wallet connect register. */
   WALLETCONNECT_NEW_NANOCONTRACT_STATUS: 'WALLETCONNECT_NEW_NANOCONTRACT_STATUS',
+  /* It triggers a process to fetch token details for a list of unregistered tokens. */
   UNREGISTEREDTOKENS_REQUEST: 'UNREGISTEREDTOKENS_REQUEST',
-  UNREGISTEREDTOKENS_UPDATE: 'UNREGISTEREDTOKENS_UPDATE',
+  /* It signals the process has loaded at least one token details with success. */
+  UNREGISTEREDTOKENS_SUCCESS: 'UNREGISTEREDTOKENS_SUCCESS',
+  /* It signals the process has failed to load at least one token details. */
+  UNREGISTEREDTOKENS_FAILURE: 'UNREGISTEREDTOKENS_FAILURE',
+  /* It signals the end of the process. */
   UNREGISTEREDTOKENS_END: 'UNREGISTEREDTOKENS_END',
   WALLETCONNECT_NEW_NANOCONTRACT_RETRY: 'WALLETCONNECT_NEW_NANOCONTRACT_RETRY',
   WALLETCONNECT_NEW_NANOCONTRACT_RETRY_DISMISS: 'WALLETCONNECT_NEW_NANOCONTRACT_RETRY_DISMISS',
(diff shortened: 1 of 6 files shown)


src/actions.js
```diff
@@ -188,8 +188,14 @@ export const types = {
   FIRSTADDRESS_FAILURE: 'FIRSTADDRESS_FAILURE',
   /* It updates the redux state of new nano contract transaction status on wallet connect register. */
   WALLETCONNECT_NEW_NANOCONTRACT_STATUS: 'WALLETCONNECT_NEW_NANOCONTRACT_STATUS',
+  /* It triggers a process to fetch token details for a list of unregistered tokens. */
   UNREGISTEREDTOKENS_REQUEST: 'UNREGISTEREDTOKENS_REQUEST',
-  UNREGISTEREDTOKENS_UPDATE: 'UNREGISTEREDTOKENS_UPDATE',
+  /* It signals the process has loaded at least one token details with success. */
+  UNREGISTEREDTOKENS_SUCCESS: 'UNREGISTEREDTOKENS_SUCCESS',
```
andreabadesso, line 194: What do you think about calling this `UNREGISTEREDTOKENS_REQUEST_SUCCESS` (and the same for other actions)

I think this would be better to debug using react dev tools

src/sagas/tokens.js
```diff
@@ -312,29 +348,29 @@ export function* requestUnregisteredTokens(action) {
...
+  const burstDelay = NODE_RATE_LIMIT_CONF.thin_wallet_token.delay;
+  const uidGroups = splitInGroups(uids, perBurst);
+  for (const group of uidGroups) {
+    // Fork is a non-blocking effect, it doesn't cause the caller suspension.
+    const tasks = yield all(group.map((uid) => fork(getTokenDetails, wallet, uid)));
+    // Awaits a group to finish before burst the next group
+    yield join(tasks);
+    // Skip delay if there is only one group or is the last group
+    if (uidGroups.length === 1 || group === uidGroups.at(-1)) {
+      break;
```
andreabadesso, line 367: I know that technically `break` will work here, as this is the only element in the array, but I think `continue` makes more sense here and will have the same effect

src/utils.js
```diff
@@ -527,3 +527,27 @@ export const parseScriptData = (scriptData, network) => {
...
+  if (groupSize === 0) {
+    return list;
+  }
+
+  const groups = [];
+  for (let i = 0; i < list.length; i += groupSize) {
+    groups.push(list.slice(i, i + groupSize));
+  }
+  return groups;
+}
```
andreabadesso, line 553: We have `lodash` in the project, you can use [chunk](https://lodash.com/docs/4.17.15#chunk)

andreabadesso (changes requested):

src/constants.js
```diff
@@ -298,3 +298,11 @@ export const NANO_CONTRACT_ACTION = {
   withdrawal: 'withdrawal',
   deposit: 'deposit',
 };
+
+export const NODE_RATE_LIMIT_CONF = {
+  thin_wallet_token: {
+    perSecond: 3,
```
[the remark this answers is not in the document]
alexruzenhack: No, it is not being used. Yes, we are planning this, despite it doesn't have an issue yet.

src/actions.js
alexruzenhack: I will call it `UNREGISTEREDTOKENS_DOWNLOAD_SUCCESS` and add `DOWNLOAD` to the other actions for consistency.

alexruzenhack: Done at: f58331b54963d2e26ea2976474046cc273875344

src/sagas/tokens.js
alexruzenhack: Ok. Done.

src/utils.js
alexruzenhack: I believe we should decrease our dependency over it. This split group is easy, why not use it?

r4mmer (approved):

Issue #514, opened 2024-08-05
Refactor fetch history to allow concurrency while mounting hydrating the transaction list


alexruzenhack:
The `fetchHistory` on `sagas/nanoContract` has a loop with async calls to `isAddressMine` and `getTxBalanceFn`, we should find a solution to call the loop allowing concurrency but maintaining the transactions order in the end.

Origin: https://github.com/HathorNetwork/hathor-wallet-mobile/pull/512

Consulting resources:
- [Faster async functions and promises](https://v8.dev/blog/fast-async)
- [Node Asynchronous Work](https://nodejs.org/en/learn/asynchronous-work/asynchronous-flow-control)

PR #569, opened 2024-09-17
refactor: fetchHistory on sagas/nanoContract to avoid iteration suspension

alexruzenhack:
### Motivation

This PR closes https://github.com/HathorNetwork/hathor-wallet-mobile/issues/514 to add concurrency on isAddressMine calls without mess with history ordering.

### Acceptance Criteria
- Refactor fetchHistory to avoid iteration suspension while hydrating the NcTxHistory item with isMine flag
- As the isAddressMine is an O(1) call and don't require a http request, we can use a plain Promise.all to collect the tasks

#### fetchHistory still working properly

https://github.com/user-attachments/assets/fdd10982-f498-45d6-b4d5-f7c0215a526a

### Security Checklist

### DIFF
--- a/src/sagas/nanoContract.js
+++ b/src/sagas/nanoContract.js
@@ -249,20 +249,16 @@ export async function fetchHistory(req) {
   }
 
   const network = wallet.getNetworkObject();
-  // As isAddressMine call is async, we should collect the tasks to avoid
-  // suspending the iteration with an await.
-  const isMineTasks = [];
-  // Translate rawNcTxHistory to NcTxHistory and collect isAddressMine tasks
-  const historyNewestToOldest = rawHistory.map((rawTx) => {
+  // Translate rawNcTxHistory to NcTxHistory
+  // Prouce a list ordered from newest to oldest
+  const historyTransformer = rawHistory.map(async (rawTx) => {
     const caller = addressUtils.getAddressFromPubkey(rawTx.nc_pubkey, network).base58;
     const actions = rawTx.nc_context.actions.map((each) => ({
       type: each.type, // 'deposit' or 'withdrawal'
       uid: each.token_uid,
       amount: each.amount,
     }));
-    // As this is a promise, collect as task to hydrate later.
-    // This strategy avoids an await suspension.
-    isMineTasks.push(isAddressMine(wallet, caller, useWalletService));
+    const isMine = await isAddressMine(wallet, caller, useWalletService);
 
     const tx = {
       txId: rawTx.hash,
@@ -275,17 +271,12 @@ export async function fetchHistory(req) {
       firstBlock: rawTx.first_block,
       caller,
       actions,
+      isMine,
     };
     return tx;
   });
 
-  // Hydrate ncTxHistory with isMine flag
-  const isMineResults = await Promise.all(isMineTasks);
-  isMineResults.forEach((isMine, idx) => {
-    historyNewestToOldest[idx].isMine = isMine;
-  });
-
-  return { history: historyNewestToOldest };
+  return { history: await Promise.all(historyTransformer) };
 }
 
 /**
--- a/src/sagas/nanoContract.js
+++ b/src/sagas/nanoContract.js
@@ -37,7 +37,7 @@ import {
 } from '../actions';
 import { logger } from '../logger';
 import { NANO_CONTRACT_TX_HISTORY_SIZE } from '../constants';
-import { getNanoContractFeatureToggle, isAddressMine } from '../utils';
+import { getNanoContractFeatureToggle } from '../utils';
 
 const log = logger('nano-contract-saga');
 
@@ -52,6 +52,25 @@ export const failureMessage = {
   blueprintInfoFailure: t`Couldn't get Blueprint info.`,
   notRegistered: t`Nano Contract not registered.`,
   nanoContractHistoryFailure: t`Error while trying to download Nano Contract transactions history.`,
+};
+
+/**
+ * Call the async wallet method `isAddressMine` considering the type of wallet.
+ *
+ * @param {Object} wallet A wallet instance
+ * @param {string} address A wallet address to check
+ * @param {boolean} useWalletService A flag that determines if wallet service is in use
+ */
+export const isAddressMine = async (wallet, address, useWalletService) => {
+  // XXX: Wallet Service doesn't implement isAddressMine.
+  // See issue: https://github.com/HathorNetwork/hathor-wallet-lib/issues/732
+  // Default to `false` if using Wallet Service.
+  if (useWalletService) {
+    return false;
+  }
+
+  const isMine = await wallet.isAddressMine(address);
+  return isMine;
 };
 
 export function* init() {
--- a/src/sagas/nanoContract.js
+++ b/src/sagas/nanoContract.js
@@ -251,7 +251,7 @@ export async function fetchHistory(req) {
   const network = wallet.getNetworkObject();
   // Translate rawNcTxHistory to NcTxHistory
   // Prouce a list ordered from newest to oldest
-  const historyTransformer = rawHistory.map(async (rawTx) => {
+  const transformedTxHistory = rawHistory.map(async (rawTx) => {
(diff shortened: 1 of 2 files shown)


src/sagas/nanoContract.js
```diff
@@ -215,16 +215,18 @@ function* registerNanoContractOnError(error) {
...
- *   history: {};
+ *   history: NcTxHistory;
  * }}
  *
  * @throws {Error} when request code is greater then 399 or when response's success is false
  */
 export async function fetchHistory(req) {
   const {
     wallet,
+    useWalletService,
```
alexruzenhack, line 229: [note] Receive it from the saga effect is more efficient. More efficient than that would be a flag in the `wallet` instance to indicate the kind of wallet, wither `fullnode` or `walletService`. Do you think should add this flag in the wallet interface?

src/sagas/nanoContract.js
```diff
@@ -245,28 +247,21 @@ export async function fetchHistory(req) {
...
-  /* TODO: Make it run concurrently while guaranting the order.
-  /* see https://github.com/HathorNetwork/hathor-wallet-mobile/issues/514
-   */
-  const historyNewestToOldest = new Array(rawHistory.length)
-  for (let idx = 0; idx < rawHistory.length; idx += 1) {
-    const rawTx = rawHistory[idx];
-    const network = wallet.getNetworkObject();
+  const network = wallet.getNetworkObject();
+  // As isAddressMine call is async we collect the tasks to avoid
+  // suspend the iteration with an await.
```
andreabadesso, line 252:
```suggestion
  // As isAddressMine call is async, we collect the tasks to avoid
  // suspending the iteration with an await.
```

src/sagas/nanoContract.js
```diff
@@ -245,28 +247,21 @@ export async function fetchHistory(req) {
...
+  const isMineTasks = [];
+  // Translate rawNcTxHistory to NcTxHistory and collect isAddressMine tasks
+  const historyNewestToOldest = rawHistory.map((rawTx) => {
```
andreabadesso, line 255: Since every iteration of this loop will create a "task" to be awaited later, why don't you just return promises in this map and await them? something like:

```
const historyTransformPromises = rawHistory.map(async (rawTx) => {
  const caller = addressUtils.getAddressFromPubkey(rawTx.nc_pubkey, network).base58;
  const actions = rawTx.nc_context.actions.map((each) => ({
    type: each.type, // 'deposit' or 'withdrawal'
    uid: each.token_uid,
    amount: each.amount,
  }));

  const isMine = await utils.isAddressMine(wallet, caller, useWalletService);

  const tx = {
    txId: rawTx.hash,
    timestamp: rawTx.timestamp,
    tokens: rawTx.tokens,
    isVoided: rawTx.is_voided,
    ncId: rawTx.nc_id,
    ncMethod: rawTx.nc_method,
    blueprintId: rawTx.nc_blueprint_id,
    firstBlock: rawTx.first_block,
    caller,
    actions,
    isMine,
  };
  return tx;
});

const historyNewestToOldest = await Promise.all(historyTransformPromises);
```

This would also avoid iterating the loop twice

andreabadesso (changes requested):

src/sagas/nanoContract.js
andreabadesso: Maybe this comment should also die if you accept my suggestion below

src/sagas/nanoContract.js
alexruzenhack: I believe we should start to measure performance to really answer this kind of question. But in the lack of data there is a catch in your proposal, the `await` provokes the block code to be suspended and add other operations as a cost. In my proposal the iteration runs from start to end without any suspension in both iterations despite to be two of them, there is only one suspension in the script `Promise.all` and of course the suspensions that happens in the `utils.isAddressMine` calls, but they happen out of the iteration context, which probably make it quicker. But his mental mode needs a performance test to be proof. Maybe we didn't deliver any benefit at all, but doesn't it makes sense?

andreabadesso: I don't believe there will be any noticeable difference in performance, and even if there is, it will most likely be irrelevant.

Regarding the 'suspensions' you mentioned, I'm not sure I fully understand what you mean. JavaScript execution isn't blocked at all when awaiting promises. When the utils.isAddressMine promise is awaited, the promise returned by the lambda passed to map is paused until it resolves, but execution continues elsewhere in the code, including other iterations of the map. Once the promise resolves, everything after the await is placed back into the event loop and executed.

The JavaScript event loop itself isn't suspended by awaiting promises. The JavaScript engine efficiently manages asynchronous operations.

Given that, I think the main focus here should be on code clarity. In my view, the version I suggested is clearer, as it handles everything in one go without the need for additional mutations in a separate loop

src/sagas/nanoContract.js
```diff
@@ -35,9 +37,7 @@ import {
 } from '../actions';
 import { logger } from '../logger';
 import { NANO_CONTRACT_TX_HISTORY_SIZE } from '../constants';
-import { consumeGenerator, getNanoContractFeatureToggle } from '../utils';
-import { getRegisteredNanoContracts, safeEffect } from './helpers';
-import { isWalletServiceEnabled } from './wallet';
+import * as utils from '../utils';
```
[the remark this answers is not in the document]
alexruzenhack: You right!

src/sagas/nanoContract.js
alexruzenhack: >I don't believe there will be any noticeable difference in performance, and even if there is, it will most likely be irrelevant.

I've tested it and in fact there aren't a significant difference, despite mine proposition be slightly better, but almost not noticeable.

```log
❯ node benchmark.js
Function 1 { avg: 602.44054917, ops: 1.6599148270775086 }
Function 2 { avg: 602.5024887400015, ops: 1.6597441814576985 }
```

I will not post here the benchmark ugly code, but if you want I share with you.

I'm adopting your proposition because of the clarity gain.

src/sagas/nanoContract.js
```diff
@@ -249,20 +249,16 @@ export async function fetchHistory(req) {
...
 
   const network = wallet.getNetworkObject();
-  // As isAddressMine call is async, we should collect the tasks to avoid
-  // suspending the iteration with an await.
-  const isMineTasks = [];
-  // Translate rawNcTxHistory to NcTxHistory and collect isAddressMine tasks
-  const historyNewestToOldest = rawHistory.map((rawTx) => {
+  // Translate rawNcTxHistory to NcTxHistory
+  // Prouce a list ordered from newest to oldest
+  const historyTransformer = rawHistory.map(async (rawTx) => {
```
andreabadesso, line 254:
```suggestion
  const transformedTxHistory = rawHistory.map(async (rawTx) => {
```

andreabadesso (approved):

src/utils.js
```diff
@@ -527,3 +515,22 @@ export const parseScriptData = (scriptData, network) => {
...
+ * @param {string} address A wallet address to check
+ * @param {boolean} useWalletService A flag that determines if wallet service is in use
+ */
+export const isAddressMine = async (wallet, address, useWalletService) => {
+  // XXX: Wallet Service doesn't implement isAddressMine.
+  // See issue: https://github.com/HathorNetwork/hathor-wallet-lib/issues/732
+  // Default to `false` if using Wallet Service.
+  if (useWalletService) {
+    return false;
+  }
```
pedroferreira1, line 532: I prefer to throw an error here instead of returning the wrong value/information to the user. We should never use this method with the wallet service until we implement it there.

What do you think?

alexruzenhack: I think we should not do that. The `isMine` flag in transaction history is not a required property, it is a convenience. We are not using the wallet service with nano contract right now, but in case we do, the only consequence will be a list of transactions without a signal if any transaction is our or not. Set the value of `isMine` to `false` for all transactions is not wrong, we can think about it as muting a signaling that is not possible.

pedroferreira1 (approved):

PR #570, opened 2024-09-18
refactor: remove address requirement from nc registering process

alexruzenhack:
### Motivation

Liberate a Nano Contract registration while using Wallet Service. It is not possible now because the address is checked as belonging to the wallet, but Wallet Service's wallet doesn't implement this check.

### Acceptance Criteria

- Remove address requirement to register a Nano Contract

>[!WARNING]
>It is lacking:
>- [ ] Load wallet's `firstAddress` on wallet initialization
>- [ ] Adapt NanoContractDetailsScreen to consume the `firstAddress` as the default address of signature in case of an operation
>- [ ] Adapt NanoContractDetailsScreen to show a feedback of first address request failure in case it happens

### Security Checklist


alexruzenhack:
We decided to not go ahead with this refactoring because there is another refactoring that would do this changes unecessary.

PR #587, opened 2024-10-25
migration for new improved pin security mode

r4mmer:
### Acceptance Criteria
- Create a migration between old and new biometry modes.
- When the feature changes, the storage state should change so the biometry can still work.

### Security Checklist


src/utils.js
```diff
@@ -362,6 +400,26 @@ export const changePin = async (wallet, oldPin, newPin) => {
   return true;
 };
 
+/**
+ * @param {string} oldPin
+ * @param {string} newPin
+ *
+ * @return {Promise<boolean>} Wether the change password was successful
+ */
+export async function changePinOnAccessData(oldPin, newPin) {
```
pedroferreira1, line 409: Why do you need this new method? Can't you use the method `changePin` above? The code seems the same, you just need the wallet object.

src/reducers/reducer.js
```diff
@@ -912,15 +916,22 @@ const onSetUseWalletService = (state, action) => ({
...
+const onSetUseSafeBiometryMode = (state, action) => ({
+  ...state,
+  safeBiometryEnabled: action.payload,
+});
+
 const onResetWalletSuccess = (state) => {
   const oldUnleashClient = state.unleashClient;
   const oldFeatureTogglesInitialized = state.featureTogglesInitialized;
   const oldFeatureToggles = state.featureToggles;
+  const oldSafeBiometryEnabled = state.safeBiometryEnabled;
```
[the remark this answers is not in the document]
r4mmer: Yes, but this is the state of the feature flag, so that we can catch toggles.

From what I understand the `state.featureToggles` is also here because the feature sagas do not fetch and set the state on the restart so we need to keep it unchanged when resetting the state.

src/sagas/featureToggle.js
```diff
@@ -123,6 +125,7 @@ export function* monitorFeatureFlags(currentRetry = 0) {
     const featureToggles = mapFeatureToggles(unleashClient.getToggles());
 
     yield put(setFeatureToggles(featureToggles));
+    yield put(setUseSafeBiometryMode(featureToggles[SAFE_BIOMETRY_MODE_FEATURE_TOGGLE]));
```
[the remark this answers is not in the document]
r4mmer: There are 3 state keys that are created from feature flags that require a special handling, the ones controlling wallet-service, push-notification and safe-biometry-mode.

The first 2 can safely be recreated during the startWallet saga ([reference](https://github.com/HathorNetwork/hathor-wallet-mobile/blob/feat/improved-pin-security-migration/src/sagas/wallet.js#L194-L195)) after the reset, but since the safe biometry mode is required to render the first screen after the reset we need to port its state in the reset.

src/utils.js
r4mmer: Yes, but to create the wallet object we need to have a started wallet and this is not the case when the user unlocks the wallet and we need to migrate the access data before starting the wallet.

Also, migration after the wallet is started can be done but may be bad to the UX because of the order of operations:
- Input pin to unlock wallet
- Load history and show list of transactions
- Migrate access data from the started wallet
  - CPU intensive with no UI indicator.
  - Wallet may be unresponsive.

src/reducers/reducer.js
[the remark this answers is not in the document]
pedroferreira1: Why don't you keep this new one together with the other feature toggles?

src/utils.js
pedroferreira1: Why don't you change the `changePin` method to use this one? You can get the `storage` from the `wallet` object and call this new method, which would receive the `storage` object.

Does it make sense? I'm trying to reduce the duplicated code here.

src/reducers/reducer.js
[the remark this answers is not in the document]
r4mmer: It is kept together, but to identify that it has changed we need to keep a separate copy so that after the `featureToggles` are updated we can compare the 2.

src/utils.js
r4mmer: I didn't because of the `checkPinAndPassword` method of the wallet that is called.
But I can use the storage methods for this

r4mmer: Done

src/utils.js
```diff
@@ -91,6 +91,46 @@ export const getAmountParsed = (text) => {
...
 
+/**
+ * Migrate the biometry configuration state if needed.
+ *
+ * @param {string} currentPassword
+ * @param {bool} safeBiometryEnabled
+ */
+export async function biometricsMigration(currentPassword, safeBiometryEnabled) {
+  const oldBiometry = STORE.getItem(IS_OLD_BIOMETRY_ENABLED_KEY);
+  const safeBiometry = STORE.getItem(IS_BIOMETRY_ENABLED_KEY);
```
andreabadesso, line 102: This is confusing...

Shouldn't this be `const biometry ...`?

andreabadesso (changes requested):

andreabadesso (approved):

pedroferreira1 (approved):

PR #590, opened 2024-10-28
Docs: improved pin security

r4mmer:
### Acceptance Criteria
- Update and QA steps on biometry to also test the new biometry mode.

### Security Checklist

### DIFF
--- a/QA.md
+++ b/QA.md
@@ -54,7 +54,11 @@
     1. Open About and go back.
     1. Click on Register a Token and check whether the camera loads correctly.
     1. Go to Manual Info, type anything and get an Invalid Configuration string.
-    1. Test Biometry (Only close the app on the steps required, this is important)
+
+1. **Biometry tests (Only close the app on the steps required, this is important)**
+    1. Test old Biometry mode
+        1. Go to unleash and disable the `safe-biometry-mode.rollout` for this wallet.
+        1. Close the app and open, navigate to Settings > Security.
         1. Enable biometry (either Fingerprint, FaceID or TouchID depending on the device)
         1. Send 1 HTR to your own address (it will ask for the biometry instead of pin)
         1. Change the pin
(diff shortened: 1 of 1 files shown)


QA.md
```diff
@@ -62,6 +66,16 @@
         1. Send 1 HTR to your own address (it will ask for the biometry again)
         1. Disable biometry
         1. Lock wallet and it must require you to type your PIN Code.
+    1. Test safe Biometry mode
+        1. Go to unleash and enable the `safe-biometry-mode.rollout` for this wallet.
+        1. Enable biometry (either Fingerprint, FaceID or TouchID depending on the device, it will require your pin)
```
andreabadesso, line 71: Biometry will already be enabled at this point, maybe make the last step from the previous section disable the biometry

andreabadesso (changes requested):

QA.md
r4mmer: The second last step of the last section is to disable the biometry, so the biometry is not enabled here.
And the first step of the last section is to disable the feature flag, so we need to start with the feature flag enable.

andreabadesso: You're right

andreabadesso (approved):

QA.md
```diff
@@ -62,6 +66,16 @@
...
+        1. Check that the "Change PIN" option is no longer on the list.
+        1. Send 1 HTR to your own address (it will ask for the biometry instead of pin)
+        1. Close the app and open, it should ask for biometry instead of pin when opening the wallet.
+        1. Send 1 HTR to your own address (it will ask for the biometry again)
+        1. Disable biometry (it will ask for the biometry)
+        1. Check that the "Change PIN" option appears on the list.
+        1. Lock wallet and it must require you to type your PIN Code.
```
pedroferreira1, line 78: Does it make sense to test the migration? In this case, we would need to add one more section here with old biometry enabled, then activate in the Unleash and validate if it works fine to load the wallet and send a tx.

What do you think?

pedroferreira1 (approved):

PR #593, opened 2024-11-01
feat: extends RCTHTTPRequestHandler to handle authentication challenge for hardening

alexruzenhack:
### Motivation

The default network security provided by ATS on iOS is broad, and this PR narrows it down to pinned certificates.

### Description

React Native makes use of the iOS network framework in its own way. A possible entry point to hook network events and intercept a connection lay on the RCTHTTPRequestHandler class, which is a category of NSURLSessionDataDelegate, therefore responsible to intercept request events and add logic over it. The React Native implementation of the data delegate doesn't implmenet the authentication challenge and relies on the default behavior provided by ATS. This fact is turns our job easier because we don't need to worry about likelly side effects. 

### Acceptance Criteria
- Extends RCTHTTPRequestHandler to handle the authentication challenge event
- Filter connections to allow only pinned domains using the ATS settings
- Conduct certificate evaluation for the allowed domains using ATS
- Cancel the connection if the host domain is not allowed
- Cancel the connection if the certificate is not trust worthy
- Crashes the application if the ATS settings is not valid

This work needs the ATS settings to be completed.

### Security Checklist


ios/RCTHTTPRequestHandler+AuthenticationChallenge.m
```diff
@@ -0,0 +1,189 @@
...
+     */
+    NSError *nsError = (__bridge_transfer NSError *)error;
+    NSLog(@"Trust evaluation failed: %@", nsError);
+  }
+  
+  return isValid;
+}
+
+- (NSSet *)getAllowedDomains {
+  NSDictionary *infoDictionary = [[NSBundle mainBundle] infoDictionary];
```
andreabadesso, line 98: Should we also throw an error if `NSAppTransportSecurity` or `NSPinnedDomains` are not found in `Info.plist`?

ios/RCTHTTPRequestHandler+AuthenticationChallenge.m
```diff
@@ -0,0 +1,189 @@
...
+  for (NSString *allowedDomain in allowedDomains) {
+    if ([allowedDomain hasPrefix:@"*."]) {
+      /* A domain is only valid for the immediate subdomain.
+       * Take for example the "hathor.network" domain, it can only
+       * validate a domain like "mainnet.hathor.network", and must
+       * invalidate a domain like "node1.mainnet.hathor.network".
+       *
+       * This validation only plays by the rules of ATS certification pinning.
+       * See: https://developer.apple.com/news/?id=g9ejcf8y
+       */
```
andreabadesso, line 156: Ok, from what I understand, this implementation enforces domain-level pinning (e.g. mainnet.hathor.network) and strict subdomain validation. Are we certain we want this approach rather than CA certificate pinning? 

With the current implementation:

- We'll need to pin certificates for each domain (or parent domain)
- Certificate rotation would require app updates (this one is a no-go for me)

If we want to pin the CA certificate instead, we'd need to modify this implementation

andreabadesso (changes requested):

ios/RCTHTTPRequestHandler+AuthenticationChallenge.m
alexruzenhack: Oh! I think so. I'm going to add this.

ios/RCTHTTPRequestHandler+AuthenticationChallenge.m
alexruzenhack: >We'll need to pin certificates for each domain (or parent domain)
I don't see any problem with this right now, specially due to our launch date. We can go with this approach now and improve it later.

>Certificate rotation would require app updates (this one is a no-go for me)
Not necessarily. There are techniques we can use to avoid this.

ios/RCTHTTPRequestHandler+AuthenticationChallenge.m
```diff
@@ -0,0 +1,189 @@
...
+ * iOS network framework. It plays the role of a NSURLSessionDelegate of the kind NSURLSessionDataDelegate.
+ * A delegate is a callback injected in the NSURLSession, and aims to intercept each phase of a connection.
+ * In React Native, when a session is created, this delegate is injected in.
+ */
+- (void)authenticationChallenge_URLSession:(NSURLSession *)session
+                                      task:(NSURLSessionTask *)task
+                       didReceiveChallenge:(NSURLAuthenticationChallenge *)challenge
+                         completionHandler:(void (^)(NSURLSessionAuthChallengeDisposition disposition, NSURLCredential *credential))completionHandler {
+  // Set of allowed domains
+  NSSet *allowedDomains = [self getAllowedDomains];
```
msbrogli, line 50: Why are we running this for every request? Isn't it immutable? I guess it might have some performance impact. Caching should work fine.

msbrogli (changes requested):

ios/RCTHTTPRequestHandler+AuthenticationChallenge.m
alexruzenhack: Done at: e447699556130341b2c10bd3d9b32c90416f34c3

ios/RCTHTTPRequestHandler+AuthenticationChallenge.m
alexruzenhack: You right, done at: e447699556130341b2c10bd3d9b32c90416f34c3

ios/RCTHTTPRequestHandler+AuthenticationChallenge.m
alexruzenhack: Oh, just to follow this up, I saw that a certificate can have an extension of _alternative hostnames_ or something like that, which can let us to add many other hostnames in just one leaf certificate.

ios/RCTHTTPRequestHandler+AuthenticationChallenge.m
```diff
@@ -0,0 +1,319 @@
...
+    } else {
+      NSLog(@"Certificate not set.");
+      // Cancel the authentication challenge
+      completionHandler(NSURLSessionAuthChallengeCancelAuthenticationChallenge, nil);
+    }
+  } else {
+    // Trust is not of type ServerTrust, proceeding with default challenge, if any
+    NSLog(@"Not ServerTrust challenge. Calling default authentication challenge.");
+    // For other authentication methods, call the original implementation
+    [self authenticationChallenge_URLSession:session didReceiveChallenge:challenge completionHandler:completionHandler];
```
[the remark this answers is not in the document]
alexruzenhack: Well, I didn't observed it. I don't have familiarity with objective-C either, but I will double check.

ios/RCTHTTPRequestHandler+AuthenticationChallenge.m
```diff
@@ -0,0 +1,319 @@
...
+          completionHandler(NSURLSessionAuthChallengeCancelAuthenticationChallenge, nil);
+        }
+      } else {
+        // Host is not allowed, cancelling authentication
+        NSLog(@"Host domain %@ is not allowed. Cancelling authentication.", challenge.protectionSpace.host);
+        // Cancel the authentication challenge
+        completionHandler(NSURLSessionAuthChallengeCancelAuthenticationChallenge, nil);
+      }
+    } else {
+      NSLog(@"Certificate not set.");
```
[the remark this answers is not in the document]
alexruzenhack: Done.

ios/RCTHTTPRequestHandler+AuthenticationChallenge.m
```diff
@@ -0,0 +1,319 @@
...
+    if (cachedCert) {
+      return (__bridge SecCertificateRef)cachedCert;
+    }
+    
+    NSString *certPath = [[NSBundle mainBundle] pathForResource:certFilename ofType:@"der"];
+    NSData *certData = [NSData dataWithContentsOfFile:certPath];
+    
+    if (!certData) {
+      @throw [NSException exceptionWithName:NSInternalInconsistencyException
+                                     reason:[NSString stringWithFormat:@"%@ not found", certFilename]
```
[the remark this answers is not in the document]
alexruzenhack: Fixed.

ios/RCTHTTPRequestHandler+AuthenticationChallenge.m
```diff
@@ -0,0 +1,319 @@
...
+  }
+}
+
+/**
+ * Fetch anchor certificates either from the resources or the cache.
+ */
+- (NSArray *)getAnchorCertificates {
+  // Load root certificates from resources in the first run and from the cache afterwards
+  SecCertificateRef hathor_root_ca_1 = [self getCert:@"hathor_network_root_ca_1"];
+  SecCertificateRef hathor_root_ca_2 = [self getCert:@"hathor_network_root_ca_2"];
```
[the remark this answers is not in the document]
alexruzenhack: I don't see a benefit by doing this. We will only have a production and a backup root CA.

ios/RCTHTTPRequestHandler+AuthenticationChallenge.m
```diff
@@ -0,0 +1,319 @@
...
+}
+
+/**
+ * As this class has only one instance and it will probably live until the user quits the app,
+ * it will probably never be called. However, we shouldn't worry much about it because
+ * the app memory will be released anyway after quit.
+ */
+- (void)swizzled_dealloc {
+    [self clearCertificateCache];
+    [self swizzled_dealloc]; // This will call the original dealloc method
```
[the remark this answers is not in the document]
alexruzenhack: It is a thin from swizzle, to execute the "original" method, we must call the custom one, the converse is true as well.

ios/RCTHTTPRequestHandler+AuthenticationChallenge.m
```diff
@@ -0,0 +1,319 @@
+#import <objc/runtime.h>
+#import <Foundation/Foundation.h>
+#import <React/RCTHTTPRequestHandler.h>
+#import <Security/Security.h>
+
+/**
+ * There will be only one instance of this class on runtime accordinly the React Native code.
+ * This provides a good opportunity to cache data on the class itself.
+ */
```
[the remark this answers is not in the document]
alexruzenhack: Fixed.

ios/RCTHTTPRequestHandler+AuthenticationChallenge.m
andreabadesso: > Not necessarily. There are techniques we can use to avoid this.

Which techniques?

andreabadesso (approved):

ios/RCTHTTPRequestHandler+AuthenticationChallenge.m
alexruzenhack: Oh, you right about "certificate rotation requires app updates". In my answer I was thinking in the scope of a leaf certificate, which by pinning the intermediate certificate we don't need an app update when the leaf certificate changes, but we do need an app update when the intermediate certificate changes, indeed. There is no other technique possible here to avoid app update.

andreabadesso: Ok, understood that if the intermediate certificate change or is invalidated this will require an app update.

src/constants.js
```diff
@@ -221,7 +221,7 @@ export const PRE_SETTINGS_TESTNET = {
   txMiningServiceUrl: TX_MINING_SERVICE_TESTNET_URL,
 };
 
-export const NODE_SERVER_MAINNET_URL = 'https://mobile.wallet.hathor.network/v1a/';
+export const NODE_SERVER_MAINNET_URL = 'https://mobile-wallet.trust.hathor.network/v1a/';
 export const EXPLORER_MAINNET_URL = 'https://explorer.hathor.network/';
```
andreabadesso, line 225: What about the explorer URLs?

andreabadesso: And tx mining service

luislhl:
Closing this as part of a cleanup of the **mobile wallet network hardening / private PKI** project (Nov 2024 – Jan 2025).

That project was never finished — all activity across every related PR and issue stops on **2025-01-09**, and no decision to pause or cancel it was ever recorded. Nothing shipped to users: the certificate pinning was never merged into the wallet, and org-wide code search finds no `NSPinnedDomains` / `network_security_config` / `trusted_roots` on any default branch.

The pilot infrastructure that *was* hand-applied to production is still there and **both pilot PKIs are now expired at every level — root, intermediate and leaf** — so those endpoints have been serving broken TLS for ~19 months.

I have inventoried everything the project left behind (Route53, Kubernetes, ACM, API Gateway, GCP, repos) and written it up, with a costed teardown plan and an explicit do-not-delete list, here:

👉 **https://github.com/HathorNetwork/ops-tools/issues/1760**

That issue is now the single parking place for this work. **Nothing here is lost by closing it** — the branch and its diff remain reachable, and #1760 links back to this item. If the project is ever reprioritised, reopen from there.

---

**Specific to this PR:** this is the client half of the project — the iOS ATS pinning via `RCTHTTPRequestHandler`, plus the Android `network_security_config.xml` and `res/raw/trusted_roots`.

It was approved by @andreabadesso on 2024-12-05 and then simply never merged. That turns out to be the single most important fact in the whole cleanup: **because this never landed, no released wallet build ever trusted or pinned the Hathor private CA.** That is what makes the server-side teardown safe, and it is why #1760 can proceed rather than being blocked on a trust-chain hazard.

Note that the PR's own description flagged the ATS/Android config as placeholder values "used during the tests", to be replaced "in a further PR with the proper certificate configuration when the certificates are issued". Those proper certificates were issued, and then expired, without this ever being updated.

Closing preserves the branch `poc/network-hardening`. (There is also a sibling branch `poc/ci-certificate-check` that never had a PR opened at all.)

@andreabadesso — flagging you since you approved this. If you know of any TestFlight build, internal track or QA device that *was* configured against `*.trust.hathor.network`, please say so on #1760; that is the Phase 0 gate before anything gets deleted.

PR #596, opened 2024-11-06
fix: errors on ses

andreabadesso:
### Acceptance Criteria
- These patches fix situations where SES blocked libraries from mutating javascript native objects
- We also fix a race condition caused by `react-native-mmkv` blocking execution.

### Security Checklist


patches/react+18.2.0.patch
```diff
@@ -0,0 +1,26 @@
...
+index 3fe63ac..415fb06 100644
+--- a/node_modules/react/cjs/react-jsx-dev-runtime.development.js
++++ b/node_modules/react/cjs/react-jsx-dev-runtime.development.js
+@@ -369,7 +369,9 @@ function describeNativeComponentFrame(fn, construct) {
+   reentry = true;
+   var previousPrepareStackTrace = Error.prepareStackTrace; // $FlowFixMe It does accept undefined.
+ 
+-  Error.prepareStackTrace = undefined;
++  // XXX: SES blocks this, it shouldn't. This will cause us to have less information on dev logs.
++  // We should remove this comment after this is fixed on SES side.
```
tuliomir, line 11: question: Do we have an issue we can track for this on SES repos?

Just in case this is forgotten for a while, the developer that eventually picks it up can immediately check some resource to understand if it's safe to remove this patch already.

tuliomir (approved):

pedroferreira1 (approved):

patches/react+18.2.0.patch
andreabadesso: Apparently it was fixed:
https://github.com/endojs/endo/issues/251

But stills throws an error here.

I'll open a KTLO to investigate this deeper

But since this only fails in `dev`, I'll continue with the merge

PR #601, opened 2024-11-07
feat: reown on nano contracts

andreabadesso:
### Acceptance Criteria
- We should disable reown when the network does not support nano contracts
- We should add a missing key in allow-scripts

### Security Checklist

### DIFF
--- a/src/actions.js
+++ b/src/actions.js
@@ -307,9 +307,28 @@ export const setIsShowingPinScreen = (isShowingPinScreen) => ({
 export const setIsOnline = (status) => ({ type: types.SET_IS_ONLINE, payload: status });
 
 /**
- * version {str} version of the connected server (e.g., 0.26.0-beta)
- * network {str} network of the connected server (e.g., mainnet, testnet)
- * */
+ * payload {{
+ *   decimal_places {number} Number of decimal places for token amounts (e.g., 2 means 0.01 is minimum unit)
+ *   genesis_block_hash {string} hash of the network's genesis block
+ *   genesis_tx1_hash {string} hash of the first genesis transaction
+ *   genesis_tx2_hash {string} hash of the second genesis transaction
+ *   max_number_inputs {number} Maximum number of inputs allowed in a transaction (default 255)
+ *   max_number_outputs {number} Maximum number of outputs allowed in a transaction (default 255)
+ *   min_tx_weight {number} Minimum transaction weight allowed (8)
+ *   min_tx_weight_coefficient {number} Coefficient for transaction weight calculation (0)
+ *   min_tx_weight_k {number} K parameter for transaction weight calculation (0)
+ *   min_weight {number} Minimum weight allowed for any operation (8)
+ *   nano_contracts_enabled {boolean} Whether nano contracts feature is enabled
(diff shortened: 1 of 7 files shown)


src/sagas/reown.js
```diff
@@ -130,12 +130,15 @@ function* isReownEnabled() {
 }
 
 function* init() {
-  const walletStartState = yield select((state) => state.walletStartState);
-
-  if (walletStartState !== WALLET_STATUS.READY) {
```
tuliomir, line 135: chore: We can remove `WALLET_STATUS` from the imports (line 75) since we removed the only places where it was being read.

pedroferreira1, line 135: Linter warned that `WALLET_STATUS` is not used anymore

src/reducers/reducer.js
```diff
@@ -747,10 +747,7 @@ const onSetUnleashClient = (state, { payload }) => ({
 
 const onSetServerInfo = (state, action) => ({
   ...state,
-  serverInfo: {
```
pedroferreira1, line 750: We should update the default redux value for `serverInfo` and the docstring

src/actions.js
```diff
@@ -310,8 +310,8 @@ export const setIsOnline = (status) => ({ type: types.SET_IS_ONLINE, payload: st
  * version {str} version of the connected server (e.g., 0.26.0-beta)
  * network {str} network of the connected server (e.g., mainnet, testnet)
  * */
```
pedroferreira1, line 312: We should update the args docstring

src/sagas/wallet.js
```diff
@@ -254,10 +254,14 @@ export function* startWallet(action) {
   yield put(setUniqueDeviceId(uniqueDeviceId));
 
   try {
-    yield call(wallet.start.bind(wallet), {
+    // XXX: This comes as undefined when the facede is the wallet-service.
```
tuliomir, line 257:
```suggestion
    // XXX: This comes as undefined when the facade is the wallet-service.
```

src/sagas/reown.js
```diff
@@ -178,11 +181,15 @@ function* init() {
     // Pass extend = true so session expiration date get renewed
     yield call(refreshActiveSessions, true);
     yield fork(listenForAppStateChange);
-    yield fork(listenForNetworkChange);
+    // yield fork(listenForNetworkChange);
```
pedroferreira1, line 184: Remove

pedroferreira1 (approved):

tuliomir (approved):

src/reducers/reducer.js
andreabadesso: Done, thanks

src/sagas/reown.js
andreabadesso: Fixed, thanks

src/sagas/reown.js
andreabadesso: Fixed, thanks

src/sagas/wallet.js
andreabadesso: Fixed, thanks

PR #629, opened 2024-11-13
fix: wait for feature toggles on app startup

r4mmer:
### Acceptance Criteria
- Create a screen to wait for feature toggles to initialize.

### Security Checklist


tuliomir (approved):

src/store.js
```diff
@@ -18,6 +18,8 @@ export const PIN_BACKUP_KEY = 'asyncstorage:pinBackup';
 export const IS_OLD_BIOMETRY_ENABLED_KEY = 'mobile:isBiometryEnabled';
 export const IS_BIOMETRY_ENABLED_KEY = 'mobile:isSafeBiometryEnabled';
 export const SUPPORTED_BIOMETRY_KEY = 'mobile:supportedBiometry';
+export const SAFE_BIOMETRY_FEATURE_FLAG_KEY = 'asyncstorage:featureFlagSafeBiometryMode';
```
pedroferreira1, line 21: I didn't understand why you created this new flag. I think you could use `IS_BIOMETRY_ENABLED_KEY` to know if it's using the safe biometry, and the `FEATURE_TOGGLES_LAST_KNOWN_VALUES_KEY` to know if it should migrate.

Is there any reason to create this new one?

src/utils.js
```diff
@@ -94,17 +94,28 @@ export const getTokenLabel = (token) => `${token.name} (${token.symbol})`;
...
+  STORE.setItem(SAFE_BIOMETRY_FEATURE_FLAG_KEY, unleashSafeBiometryFeature);
+  // The safe biometry feature flag has changed, we need to check if a migration is required.
 
-  if (safeBiometryEnabled) {
-    // Safe biometry mode, need to migrate if old biometry is enabled.
-    if (oldBiometry) {
+  if (unleashSafeBiometryFeature) {
+    // Unleash flag is enabling safe biometry
+    // if we have the old mode active a migration is required.
+    // else we can ignore migration since the use is not using biometry.
```
tuliomir, line 116:
```suggestion
    // else we can ignore migration since the user is not using biometry.
```

src/store.js
tuliomir: If I understand correctly, the user may leave the "Pin Screen" window open for a while and `FEATURE_TOGGLES_LAST_KNOWN_VALUES_KEY` may have been updated multiple times. 

`SAFE_BIOMETRY_FEATURE_FLAG_KEY`, however, will have never been updated. It will only be updated once the user inserts a valid PIN or allows biometry to do so. This seems to be a way to ensure the cpu-heavy and potentially slow migration process will only be executed once, and only immediately after a deliberate user interaction.

Maybe we should add a comment making this safeguard measure explicit, so that future maintenance doesn't change this behavior by mistake.

tuliomir (approved):

src/store.js
r4mmer: There are 3 keys that operate together:
- last known value of the feature toggles, updated instantly when they change with no operation required.
- `SAFE_BIOMETRY_FEATURE_FLAG_KEY` which the app uses as the actual value of the feature flag.
  - This only changes during a migration.
  - Differences between the feature flag key and the last known value of the feature toggle are an indication that a migration is necessary.
- The "is biometry enabled" keys, which indicate if the user has manually enabled the biometry in the settings.

The `SAFE_BIOMETRY_FEATURE_FLAG_KEY` is used as a stable and persistent feature flag that also indicates the current operating mode (feature flag + store state), this stability is what solved the race condition since the feature toggles can change at any time and will not affect the store state.

pedroferreira1 (approved):

PR #644, opened 2025-01-10
feat: show full network name from serverInfo (or versionData)

andreabadesso:
image

### Acceptance Criteria
- We should show the full server name in both facades instead of just `testnet`, e.g. `testnet-nano-alpha`
- We should hide the network bar when changing networks

### Security Checklist

### DIFF
--- a/src/reducers/reducer.js
+++ b/src/reducers/reducer.js
@@ -522,7 +522,7 @@ const initialState = {
    * The full network name of the connected server (e.g. testnet-golf instead
    * of only testnet).
    *
-   * @type {null|string} null if uninitialized, string after user it's set in
+   * @type {null|string} null if uninitialized, string after it's set in
    * the wallet saga.
    */
   fullNodeNetworkName: null,


src/components/NetworkSettings/NetworkStatusBar.js
```diff
@@ -48,10 +48,12 @@ const style = {
 };
 
 export const NetworkStatusBar = () => {
-  const getStatusText = (networkSettings) => `${customNetworkText}: ${networkSettings.network}`;
+  const getStatusText = (networkName) => `${customNetworkText}: ${networkName}`;
   const networkSettings = useSelector((state) => state.networkSettings);
+  const networkName = useSelector((state) => state.networkName);
```
tuliomir, line 53: suggestion(non-blocking): Create the function only after declaring its dependency variables. Also, fetch all `store` data in a single selector fetch.

```js
const { networkSettings, networkName } = useSelector((state) => ({
  networkSettings: state.networkSettings,
  networkName: state.networkName,
});
const getStatusText = (networkName) => `${customNetworkText}: ${networkName}`;
```

tuliomir (approved):

src/sagas/wallet.js
```diff
@@ -260,6 +261,17 @@ export function* startWallet(action) {
...
+    let network = get(serverInfo, 'network');
+    if (useWalletService) {
+      // In the wallet-service facade, serverInfo is null, so we need to get
+      // version data
+      const versionData = yield call([wallet, wallet.getVersionData]);
+      network = versionData.network;
+    }
+
+    // Set the network name in redux
+    yield put(setNetworkName(network));
```
pedroferreira1, line 274: The `serverInfo` is the same return as in the `versionData`. Why don't you save the `versionData` in the `serverInfo` redux state that we already have instead of creating a new one to save this same information?

src/reducers/reducer.js
```diff
@@ -2131,3 +2132,8 @@ export const onUnregisteredTokensDownloadEnd = (state) => ({
     isLoading: false,
   },
 });
+
+export const onSetNetworkName = (state, action) => ({
+  ...state,
+  networkName: action.payload,
```
pedroferreira1, line 2138: Shouldn't we have the default value for `networkName`? Even if it's `null` or empty string, at least we could have it there with a docstring.

src/actions.js
```diff
@@ -1507,3 +1508,13 @@ export const appStateUpdate = (oldState, newState) => ({
...
   }
 });
+
+/**
+ * Network name, which is used in the network bar
+ */
+export const SET_NETWORK_NAME = 'SET_NETWORK_NAME';
+
+export const setNetworkName = (networkName) => ({
+  type: SET_NETWORK_NAME,
```
pedroferreira1, line 1518: Why don't you use `types.SET_NETWORK_NAME`?

src/actions.js
```diff
@@ -1507,3 +1508,13 @@ export const appStateUpdate = (oldState, newState) => ({
     newState,
   }
 });
+
+/**
+ * Network name, which is used in the network bar
+ */
+export const SET_NETWORK_NAME = 'SET_NETWORK_NAME';
+
+export const setNetworkName = (networkName) => ({
```
pedroferreira1, line 1517: Missing docstring

andreabadesso: Fixed, thanks

src/reducers/reducer.js
andreabadesso: Yes! Fixed, thanks

tuliomir (approved):

src/reducers/reducer.js
```diff
@@ -518,6 +518,14 @@ const initialState = {
     address: null,
     error: null,
   },
+  /**
+   * The full network name of the connected server (e.g. testnet-golf instead
+   * of only testnet).
+   *
+   * @type {null|string} null if uninitialized, string after user it's set in
```
pedroferreira1, line 525:
```suggestion
   * @type {null|string} null if uninitialized, string after it's set in
```

pedroferreira1 (approved):

src/reducers/reducer.js
andreabadesso: Fixed, thanks

PR #653, opened 2025-02-14
fix: Crash on invalid words

tuliomir:
There were two events that triggered the validation of the words on the "Import Wallet" screen:
- Clicking the "Next" button at the bottom of the screen
- Clicking the native "Submit" button from the TextArea.

Each of these events had their own validation, which led to inconsistent app behavior and crashes when using the "Submit" button.

### Acceptance Criteria
- In the "Import Wallet" screen, user must not receive an error by clicking "Submit" with an empty text
- Also must not receive an error with invalid words

### Security Checklist

### Example interaction

https://github.com/user-attachments/assets/f0cc8b56-ae82-4a48-acb0-a27196e6a239

### DIFF
--- a/src/screens/InitWallet.js
+++ b/src/screens/InitWallet.js
@@ -321,10 +321,10 @@ class LoadWordsScreen extends React.Component {
   loadClicked = () => {
     Keyboard.dismiss();
     const { rawWords } = this.state;
-    const { isValid, errorMessage } = handleValidWords(rawWords);
+    const { isValid, errorMessage, words } = handleValidWords(rawWords);
 
     if (isValid) {
-      this.props.navigation.navigate('ChoosePinScreen', { words: rawWords });
+      this.props.navigation.navigate('ChoosePinScreen', { words: words.join(' ') });
     } else {
       this.setState({ errorMessage });
     }
--- a/src/screens/InitWallet.js
+++ b/src/screens/InitWallet.js
@@ -235,9 +235,53 @@ class NewWordsScreen extends React.Component {
   }
 }
 
+/**
+ * Handles the validation of the raw seed words input
+ * @param {string} rawWords String containing all the words written by the user
+ * @returns {{words: string[], errorMessage: string, isValid: boolean}}
+ */
+function handleValidWords(rawWords) {
+  let trimmedWordsStr = rawWords.trim(/\s+/);
+  let errorMessage = '';
+  let isValid = false;
+
+  try {
+    const { valid, words } = walletUtils.wordsValid(trimmedWordsStr);
+    trimmedWordsStr = words;
+    isValid = valid;
+  } catch (e) {
+    // Handling unknown errors
+    if (!(e instanceof hathorErrors.InvalidWords)) {
+      throw e;
+    }
+
+    // Handling Invalid Words errors
+    isValid = false;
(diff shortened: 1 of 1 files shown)


src/screens/InitWallet.js
```diff
@@ -235,9 +235,53 @@ class NewWordsScreen extends React.Component {
...
+
 class LoadWordsScreen extends React.Component {
+  /**
+   * State for the import wallet words screen
+   * @type {{rawWords: '', words: string[], errorMessage: string, isValid: boolean}}
+   */
   state = {
-    words: '',
+    rawWords: '',
+    words: [],
```
pedroferreira1, line 284: Are you using this state?

src/screens/InitWallet.js
```diff
@@ -235,9 +235,53 @@ class NewWordsScreen extends React.Component {
...
+  } catch (e) {
+    // Handling unknown errors
+    if (!(e instanceof hathorErrors.InvalidWords)) {
+      throw e;
+    }
+
+    // Handling Invalid Words errors
+    isValid = false;
+    errorMessage = e.message;
+    const listOfInvalidWords = e.invalidWords.filter((word) => word.length !== 0);
```
pedroferreira1, line 261: Is this filter necessary? I think the method will return only the invalid words. Or it could return an empty string?

src/screens/InitWallet.js
```diff
@@ -265,44 +309,24 @@ class LoadWordsScreen extends React.Component {
...
-    const words = this.state.words.join(' ');
-    this.setState({ errorMessage: '' });
-    const result = walletUtils.wordsValid(words);
-    if (result.valid) {
-      this.props.navigation.navigate('ChoosePinScreen', { words });
+    const { rawWords } = this.state;
+    const { isValid, errorMessage } = handleValidWords(rawWords);
+
+    if (isValid) {
+      this.props.navigation.navigate('ChoosePinScreen', { words: rawWords });
```
pedroferreira1, line 327: In the old code the parameter was a string with `nonEmptyWordsLowerCase` and now you are using the raw words. Is this still working with multiple spaces between two words or with a mix of upper/lower case?

src/screens/InitWallet.js
tuliomir: Yes, it could! That would result in an error message like `List of invalid words: ` , with no words at all.

src/screens/InitWallet.js
tuliomir: Yes, the `handleValidWords()` method requires the raw text from the TextArea. For the `onChangeText()` event handler that's trivial, because it receives it at each change. For the `loadClicked()`, however, I would need to somehow send it.

The cleanest solution I found was to always update the `rawWords` at each change. The alternative would be to send an event on the `loadClicked()` events that were raised by the submit button only, and that felt cumbersome.

src/screens/InitWallet.js
tuliomir: You're right, that was a bug.
It was fixed in https://github.com/HathorNetwork/hathor-wallet-mobile/pull/653/commits/4ad8667946a1107b52f2c4139a8d63b26d3ee0a8 by passing the treated words array joined as a string, like it was before.

pedroferreira1 (approved):

PR #654, opened 2025-02-14
fix: duplicated HTR symbol

tuliomir:
### Acceptance Criteria
- Removes duplicated "HTR" Symbol from the create token screen
- Replaces hardcoded "HTR" symbol with the dynamic value for the native token
- Deprecates obsolete "Default Token" property from the wallet config file.
- Updates the readme instructions on how to update the translation files

Before and After

### Security Checklist

### DIFF
--- a/src/screens/CreateTokenAmount.js
+++ b/src/screens/CreateTokenAmount.js
@@ -139,7 +139,9 @@ class CreateTokenAmount extends React.Component {
                 <InfoBox
                   items={[
                     <Text>{t`Deposit:`} <Strong style={amountStyle}>
-                      {hathorLib.numberUtils.prettyValue(this.state.deposit)} {_DEFAULT_TOKEN.symbol}
+                      {hathorLib.numberUtils.prettyValue(
+                        this.state.deposit
+                      )} {_DEFAULT_TOKEN.symbol}
                     </Strong></Text>,
                     <Text>
                       {jt`You have ${amountAvailableText} available`}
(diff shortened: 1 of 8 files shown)


andreabadesso (approved):

src/screens/CreateTokenAmount.js
```diff
@@ -141,7 +141,7 @@ class CreateTokenAmount extends React.Component {
                       {hathorLib.numberUtils.prettyValue(this.state.deposit)} HTR
                     </Strong></Text>,
                     <Text>
-                      {jt`You have ${amountAvailableText} HTR available`}
+                      {jt`You have ${amountAvailableText} available`}
```
r4mmer, line 144: _Unrelated to line_

We should change the `amountAvailableText` to use the native token symbol we get from the fullnode.
This is because side-dags may use another symbol for the native token.

r4mmer (changes requested):

src/screens/CreateTokenAmount.js
tuliomir: Fixed on the `CreateTokenAmount` and `CreateTokenConfirm` screens. There are still other instances of hardcoded `HTR` messages in other places in the code, but those must be fixed in another PR.

Fixed native token

tuliomir: To reproduce this test, change inside the `src/sagas/wallet.js` -> `startWallet()`:
```js
    const serverInfo = yield call(wallet.start.bind(wallet), {
      pinCode: pin,
      password: pin,
    });

    serverInfo.native_token = {
      name: 'MyNative',
      symbol: 'NTV',
    };
    yield put(setServerInfo(serverInfo));
    console.log(`New serverInfo:`, serverInfo);
 ```

r4mmer (approved):

Issue #385, opened 2023-11-14
bug: Testnet tokens don't get removed when changing network to mainnet


andreabadesso:
Steps to reproduce:

1. Change network to `testnet`
2. Register some testnet tokens
3. Change network to `mainnet`

The tokens from testnet should be removed

tuliomir:
This behavior only happens when connected to the Wallet Service as of v0.30.0

PR #660, opened 2025-03-05
fix: clear tokens from other networks

tuliomir:
### Motivation
When changing networks, the registered tokens were not being cleaned properly. This resulted in situations like testnet tokens being exhibited while the user is on mainnet. This was raised by #385 .

### Acceptance Criteria
- All registered custom tokens and nano contracts should be cleaned when changing networks

### Alternative Solutions
A more user-friendly approach to this issue would be comparing the genesis block of the previous network ( if any ) and the current network, and only cleaning those tokens if these hashes changed.

This would ensure that changing the fullnode while staying inside the same network ( for example, changing from `https://node1.mainnet.hathor.network/v1a/` to `https://node2.mainnet.hathor.network/v1a/` ) would not remove all registered tokens unnecessarily.

This solution was not implemented here so that the bug is fixed immediately, while this greater refactor is left to #667 .

### Security Checklist


src/sagas/wallet.js
```diff
@@ -162,7 +162,10 @@ export function* startWallet(action) {
   // but the wallet may be closed unexpectedly
   const storage = STORE.getStorage();
   yield call([storage.store, storage.store.cleanMetadata]); // clean metadata on memory
-  yield call([storage, storage.cleanStorage], true); // clean transaction history
+  // clean transaction history, addresses, token and nano-contract indexes
+  // XXX: It would be better to keep registered tokens and nano contracts
```
andreabadesso, line 166: Do we have an issue created describing this?

andreabadesso (approved):

src/sagas/wallet.js
tuliomir: Yes, we have:
- #667

src/sagas/wallet.js
```diff
@@ -162,7 +162,10 @@ export function* startWallet(action) {
...
+  // if the network has same genesis
+  yield call([storage, storage.cleanStorage], true, true, true);
```
pedroferreira1, line 168: I have two questions:

1. When is this method called? If I open the app the wallet locked (but started) and put my PIN, would it call this method to start my loaded wallet? In this case it would clean the token data previously added, is that correct?

2. Are you cleaning the registered tokens and nano contracts from redux as well?

tuliomir: 1. You caught a nasty bug :bug: I reverted the changes on https://github.com/HathorNetwork/hathor-wallet-mobile/pull/660/commits/a0a0ea1babe261b7c4f806c1eee3ec4d7092048f and implemented the correct fix on https://github.com/HathorNetwork/hathor-wallet-mobile/pull/660/commits/dd01b71ce759076a458aa51023ec55070a4a6955 using the same method, but on the correct place: the `networkSettings` saga.

2. The registered tokens are being cleaned by the solution described above, but the nano contracts are cleaned on wallet reload. See [the reducer code here](https://github.com/HathorNetwork/hathor-wallet-mobile/blob/77626f8ad8cd3208cb1de24428f89cf43693cc3f/src/reducers/reducer.js#L1442-L1448).

pedroferreira1 (approved):

andreabadesso (approved):

PR #661, opened 2025-03-06
fix: we should confirm that the user wants to decline the nc transaction

andreabadesso:
### Acceptance Criteria

### Security Checklist

### DIFF
--- a/src/components/Reown/NanoContract/NewNanoContractTransactionRequest.js
+++ b/src/components/Reown/NanoContract/NewNanoContractTransactionRequest.js
@@ -16,7 +16,8 @@ import {
   ScrollView,
   TouchableWithoutFeedback,
   Image,
-  Text
+  Text,
+  BackHandler
 } from 'react-native';
 import { useDispatch, useSelector } from 'react-redux';
 import { useNavigation } from '@react-navigation/native';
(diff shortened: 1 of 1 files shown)


src/components/Reown/SignOracleDataRequest.js
```diff
@@ -47,43 +49,62 @@ export const SignOracleDataRequestData = ({ data }) => (
...
+  };
+
+  const { navigateBack } = useBackButtonHandler(
+    onDeclineTransaction,
+  );
 
   const onAcceptSignOracleDataRequest = () => {
     // Signal the user has accepted the current request and pass the accepted data.
     dispatch(reownAccept());
-    navigation.goBack();
```
andreabadesso, line 55: We don't need to go back as we will navigate automatically to the PinScreen after accepting the request.

This prevents a weird animation where it would navigate back and then forward

src/hooks/useBackButtonHandler.js
```diff
@@ -0,0 +1,107 @@
...
+   * @param {string} screenName - Name of the screen to navigate to
+   * @param {Object} params - Optional parameters to pass to the screen
+   * @param {Function} beforeNavigateCallback - Optional callback to execute before navigation
+   */
+  const navigateTo = (screenName, params = {}, beforeNavigateCallback) => {
+    setIsIntentionallyNavigatingBack(true);
+
+    // Execute callback if provided
+    if (beforeNavigateCallback) {
+      beforeNavigateCallback();
```
raul-oliveira, line 91: NIT: have you tried to deal with beforeNavigateCallback as an async function?
it will wait for the current js/event loop

src/hooks/useBackButtonHandler.js
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    // We need to give time for any modal to close
+    // before navigating back
+    setTimeout(() => {
```
raul-oliveira, line 75: NIT: have you tried to deal with beforeNavigateCallback as an async function?
it will wait for the current js/event loop

raul-oliveira (approved):
Code looks good

pedroferreira1 (approved):

src/hooks/useBackButtonHandler.js
andreabadesso: I guess it would, but it's an optional parameter, so we can't rely on it

PR #666, opened 2025-03-07
feat: i18n script automation

tuliomir:
### Motivation
The translation update process required many manual steps which could be simplified by automation. 

### Acceptance Criteria
- Simplifies the i18n translation update with a single command `make i18n`
- Simplifies the i18n validation with a single command `make check_i18n`
- Updates the CI with this simplified validation

### Note 1: Flake Nix
The `flake.nix` custom commands for i18n were changed to reflect those new commands available.

### Note 2: Strict `po` validation
Also, there was an instruction on the QA document for excuting `check_po_strict` during the QA process. When I tried implementing this strict validation it triggered hundreds of errors.

So, for now, I decided to leave the strict validation disabled and deal with this in a future PR. The only change needed in this current scope would be to add the `true` parameter on line 233:

https://github.com/HathorNetwork/hathor-wallet-mobile/blob/0f2f474edb01f8d73c3907cf0f5c0461c42142b6/scripts/update_translations.js#L233

### Note 3: Strict file validation on CI
I tried making the CI validation fail if any file changes were detected while validating the translation files during the CI. However, we were faced with line-break changes when the developer made the translation using different operational systems.

So this stricter validation was removed from CI, and can be re-enabled by simply using the `checkForChanges()` return value to decide if the script should pass or not. In the future we can implement different command line values to fine-tune the validation.

The motivation for this stricter validation is that some times the CI passes even when the translation files are not up-to-date. This results in a future PR having to deal with translation outside its scope when the CI would finally catch up with the pending adjustments.

### Security Checklist


locale/pt-br/texts.po
```diff
@@ -1754,8 +1755,9 @@ msgstr "Saldo insuficiente"
 msgid ""
 "Ensure your wallet balance covers the required amount to accept this "
 "transaction."
-msgstr "Certifique-se de que o saldo da sua carteira cubra o valor necessário"
-"para aceitar esta transação."
+msgstr ""
+"Certifique-se de que o saldo da sua carteira cubra o valor necessário para "
+"aceitar esta transação."
```
tuliomir, line 1760: This was the only manual change needed for the new and stricter `check_i18n` validation to pass.

scripts/update_translations.js
```diff
@@ -0,0 +1,245 @@
...
+  const invalidPot = checkPotOutdated();
+  const invalidPos = checkPoTranslations();
+  const translationFilesChanged = checkForChanges();
+  if (invalidPot || invalidPos || hasFuzzyTags || translationFilesChanged) {
+    console.log(`❌ Translations are not up-to-date. Please review the changes.`);
+    process.exit(1);
+  } else {
+    console.log(`✅ Translations validated successfully.`);
+    process.exit(0);
+  }
```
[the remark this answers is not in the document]
tuliomir: Thanks! Refactored on https://github.com/HathorNetwork/hathor-wallet-mobile/pull/666/commits/0882d112926e298656b4d40cadd0314b50f503df

scripts/update_translations.js
```diff
@@ -0,0 +1,245 @@
...
+    execSync(`npx ttag extract -o ${myTmpDir}/pot ./src/`, { stdio: 'inherit' });
+    // Generate the `.po` files
+    execSync(`msgen -o ${myTmpDir}/po ${myTmpDir}/pot`, { stdio: 'inherit' });
+    // Check if the `.pot` file is outdated
+    execSync(`msgcmp ${myTmpDir}/po ./locale/texts.pot`, { stdio: 'inherit' });
+  } catch (error) {
+    isPotOutdated = true;
+    console.error('❌ Error checking pot file, or pot file outdated:', error);
+  } finally {
+    fs.rmSync(myTmpDir, { recursive: true, force: true });
```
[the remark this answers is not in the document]
tuliomir: Great idea! Added cleanup function on https://github.com/HathorNetwork/hathor-wallet-mobile/pull/666/commits/96ffa0318b39e79ad86b09305d5b152df21599e0

scripts/update_translations.js
```diff
@@ -0,0 +1,258 @@
...
+ * Checks for filesystem changes using `git status`. If any changes are found, it means the
+ * translation files have changed, and this may trigger a CI fail when running inside the CI.
+ * @returns {boolean} True if changes were found, or if an error happened
+ */
+function checkForChanges() {
+  console.log(`⏳ Checking for filesystem changes after i18n execution...`);
+  let statusResult = '';
+  let hasChanges = false;
+  try {
+    statusResult = execSync('git status --porcelain', { encoding: 'utf-8' });
```
pedroferreira1, line 179: Is it checking changes only in translation files?

pedroferreira1 (approved):
> Also, there was an instruction on the QA document for excuting check_po_strict during the QA process. When I tried implementing this strict validation it triggered hundreds of errors.

We were running the strict validation only for pt-br, which is the translation file that we will always need the translation (because we can do it ourselves)

scripts/update_translations.js
tuliomir: This checks for changes in the whole repository. However, the results were not being consistent across operational systems.

So I made this validation not fail the validation on https://github.com/HathorNetwork/hathor-wallet-mobile/pull/666/commits/78fbdb2331ed86b21032d4119d3a1b2be8a6efe0 . More information on this relaxation of rules is available on the updated PR description.

andreabadesso (approved):

PR #671, opened 2025-03-20
refactor: decimal places

andreabadesso:
### Acceptance Criteria
- We should use decimal places from the serverInfo instead of using the constant one
- We should refactor AmountTextInput to be a functional component

### Security Checklist

### DIFF
--- a/src/components/AmountTextInput.js
+++ b/src/components/AmountTextInput.js
@@ -7,7 +7,6 @@
 
 import React, { useState, useRef, useEffect, forwardRef, useImperativeHandle } from 'react';
 import { StyleSheet, TextInput } from 'react-native';
-import { constants } from '@hathor/wallet-lib';
 import { getAmountParsed, getIntegerAmount } from '../utils';
 import { COLORS } from '../styles/themes';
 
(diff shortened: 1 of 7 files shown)


src/components/NewPaymentRequest.js
```diff
@@ -30,6 +30,7 @@ const mapStateToProps = (state) => ({
   selectedToken: state.selectedToken,
   wallet: state.wallet,
   tokenMetadata: state.tokenMetadata,
+  decimalPlaces: state.serverInfo.decimal_places,
```
pedroferreira1, line 33:
```suggestion
  decimalPlaces: state.serverInfo?.decimal_places,
```

src/components/AmountTextInput.js
```diff
@@ -5,64 +5,82 @@
...
+ * @param {boolean} [props.autoFocus] - Whether the input should be focused on mount
+ * @param {number} [props.decimalPlaces] - Number of decimal places to use (optional)
+ * @param {React.Ref} ref - Forwarded ref, exposes the focus() method
+ * @returns {React.ReactElement} A formatted amount input component
+ */
+const AmountTextInput = forwardRef((props, ref) => {
+  const inputRef = useRef(null);
+  const [text, setText] = useState(props.value || '');
+  const decimalPlaces = props.decimalPlaces == null
+    ? constants.DECIMAL_PLACES
```
pedroferreira1, line 34: Why don't you get it from the redux as well? The default shouldn't be the lib constant, in my opinion.

andreabadesso: I don't like adding redux to components, I think they should be as "pure" as possible

But I made this a required parameter, can you please check dc72286e06195051f186828cfcf78c46bd5092ac?

src/components/NewPaymentRequest.js
andreabadesso: Done

pedroferreira1 (approved):

raul-oliveira (approved):

PR #682, opened 2025-04-29
chore: Cleanup scripts

tuliomir:
### Motivation
When upgrading React Native versions, it's imperative to clean not only the local development folder but also the global caches for both Android and iOS at the operational system level. Failing to do so causes bugs with misleading messages that are hard to solve otherwise.

These scripts aim to facilitate this cleanup, providing the developer with a more reliable state upon which to start working.

Also, it's helpful to have this script run at pre-release, as it ensures the environment is as close as possible to that of the developer.

### Acceptance Criteria
- Adds new cleanup scripts that also wipe global caches
- Incorporates those new scripts into the `pre_release.sh` script

### Security Checklist

### DIFF
--- a/scripts/cleanup_ios.sh
+++ b/scripts/cleanup_ios.sh
@@ -27,6 +27,8 @@ echo "🧹 Cleaning Xcode DerivedData..."
 rm -rf ~/Library/Developer/Xcode/DerivedData/*
 rm -rf ~/Library/Developer/Xcode/Archives/*
 
+# Deintegrating CocoaPods. This will cause all pods to be re-downloaded upon the next install
+# which will take a lot of bandwidth and processing time
 echo "🧹 Deintegrating and cleaning CocoaPods..."
 pod deintegrate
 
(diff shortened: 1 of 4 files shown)


scripts/cleanup_ios.sh
```diff
@@ -0,0 +1,36 @@
...
+xcodebuild clean
+
+echo "🧹 Cleaning CocoaPods caches..."
+rm -rf Pods
+rm -rf ~/Library/Caches/CocoaPods
+
+
+echo "🧹 Cleaning Xcode DerivedData..."
+rm -rf ~/Library/Developer/Xcode/DerivedData/*
+rm -rf ~/Library/Developer/Xcode/Archives/*
```
andreabadesso, line 28: This will delete all data from all projects the user have in his computer, right? Maybe do something like:

```bash
PROJECT_NAME=$(ls *.xcodeproj | sed 's/.xcodeproj//')
rm -rf ~/Library/Developer/Xcode/DerivedData/"$PROJECT_NAME"-*
```

andreabadesso (changes requested):

pedroferreira1 (approved):

PR #690, opened 2025-07-01
fix: Minimizes modal flickering issues

tuliomir:
### Summary
After the React Navigation upgrade on #688, the modals started flickering during the show and hide animations. This is a known issue on the dependency we use, and is being tracked by https://github.com/react-native-modal/react-native-modal/issues/92 .

This PR aims to minimize the flickering impact on the user through the use of workarounds that are discussed in the issue above.

### Acceptance Criteria
- Flickering is reduced during modal exhibition

### Before the fix
https://github.com/user-attachments/assets/85c13594-8843-4594-9994-354bb6079406

### After the fix
https://github.com/user-attachments/assets/d4357fef-ad13-409f-a253-7f39ad9807f3

### Security Checklist

### DIFF
--- a/src/components/HathorModal.js
+++ b/src/components/HathorModal.js
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React from 'react';
+import React, { useState } from 'react';
 import {
   StyleSheet, View,
 } from 'react-native';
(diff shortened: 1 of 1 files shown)


tuliomir:
Superseded by #691

PR #699, opened 2025-07-08
fix: multiple modal nc address seletor

andreabadesso:
https://github.com/user-attachments/assets/2cca18db-8760-4c12-9131-a2f7be45b722

### Acceptance Criteria
- We should remove the delay before showing children in the BackdropModal as it was not benefiting us in any way and was causing the `SelectAddressModal` to display a small modal before rendering the full content.
- We should change the `SelectAddressModal` modal in NanoContractDetail screen to only show one modal at a time, preventing nested modals which were causing react-native to break
- We should increase the SelectAddressModal size so it's the same size of all other modals

### Security Checklist

### DIFF
--- a/src/components/NanoContract/NanoContractDetailsHeader.js
+++ b/src/components/NanoContract/NanoContractDetailsHeader.js
@@ -26,6 +26,7 @@ import { TextValue } from '../TextValue';
 import { TextLabel } from '../TextLabel';
 import { EditInfoContainer } from '../EditInfoContainer';
 import { SelectAddressModal } from './SelectAddressModal';
+import { EditAddressModal } from './EditAddressModal';
 import { UnregisterNanoContractModal } from './UnregisterNanoContractModal';
 
 /**
@@ -44,16 +45,24 @@ export const NanoContractDetailsHeader = ({ nc, address, onAddressChange }) => {
   const [isShrank, toggleShrank] = useState(true);
   const [selectedAddress, setSelectedAddress] = useState(address);
   const [showSelectAddressModal, setShowSelectAddressModal] = useState(false);
+  const [showEditAddressModal, setShowEditAddressModal] = useState(false);
+  const [selectedItem, setSelectedItem] = useState({ address });
   const [showUnregisterNanoContractModal, setShowUnregisterNanoContractModal] = useState(false);
+  const [modalStep, setModalStep] = useState('select');
 
   const isExpanded = () => !isShrank;
 
   const onEditAddress = () => {
+    setModalStep('select');
     setShowSelectAddressModal(true);
   };
 
   const toggleSelectAddressModal = () => {
     setShowSelectAddressModal(!showSelectAddressModal);
+  };
+
+  const toggleEditAddressModal = () => {
+    setShowEditAddressModal(!showEditAddressModal);
   };
 
   const onUnregisterNanoContract = () => {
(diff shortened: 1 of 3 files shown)


src/components/NanoContract/SelectAddressModal.js
```diff
@@ -115,7 +118,6 @@ export const SelectAddressModal = ({ address, show, onSelectAddress, onDismiss }
                   <Text style={styles.infoText}>{t`To change, select other address on the list below.`}</Text>
                   <Text>
                     <TextValue bold>{t`Address`}</TextValue>
-                    {/* the unicode character u00A0 means no-break space. */}
```
tuliomir, line 118: suggestion: I think it's best to keep this comment, as it's very self explanatory.

src/components/BackdropModal.js
```diff
@@ -58,25 +58,6 @@ const BackdropModal = ({
...
-      }, 100);
-
-      return () => clearTimeout(timer);
-    }
-
-    // Hide children immediately when modal is dismissed
-    setShowChildren(false);
-    return undefined;
-  }, [visible]);
-
```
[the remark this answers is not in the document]
tuliomir: chore: By removing this effect we have a regression error on the "Backup Words" screen where the "Wrong Words" modal content is rendered before the modal completes its display animation.

Since this error is very specific to this Backup screen, my first suggestion would be adding a `delayContentRendering = false` parameter to this `BackdropModal`. If the parameter is true, only then the rendering is delayed. And this parameter would only be set on the `BackupWords.js` screen.

src/components/NanoContract/SelectAddressModal.js
```diff
@@ -35,41 +35,33 @@ import { selectAddressAddressesRequest } from '../../actions';
  * Callback function called when an address is selected
  * @param {() => {}} props.onDismiss
  * Callback function called to dismiss the modal
+ * @param {(item: Object) => {}} props.onEditAddress
+ * Callback function called when an address item is selected for editing
```
tuliomir, line 39: chore: We have to update these docstrings. Some params were removed and others added

src/components/NanoContract/NanoContractDetailsHeader.js
```diff
@@ -64,12 +67,28 @@ export const NanoContractDetailsHeader = ({ nc, address, onAddressChange }) => {
     setShowUnregisterNanoContractModal(!showUnregisterNanoContractModal);
   };
 
+  const handleEditAddress = (item) => {
+    setSelectedItem(item);
+    setModalStep('edit');
+  };
+
+  const onDismissEditAddressModal = () => {
+    setSelectedItem({ address });
```
tuliomir, line 76: question(non-blocking): This reverts to the original address, right? Maybe add a comment here reinforcing this?

All other address handlers receive their properties from the parameter, which may confuse future maintainers.

andreabadesso: Yes, this behavior is intentional. When the user dismisses the edit modal (by clicking "Go back"), we want to reset the selectedItem back to the original address prop so that if they open the edit flow again, it starts from the current address rather than whatever address they had previously selected but didn't confirm.

tuliomir (approved):

pedroferreira1 (approved):

PR #729, opened 2025-08-06
feat: update feature flags on network change

pedroferreira1:
### Motivation

We were only fetching feature flags for mainnet stage (which was hardcoded from constants) because we were never updating the unleash client after a network settings change.

### Acceptance Criteria
- Update the unleash client on network settings change
- Use network settings stage for the unleash client
- Send to unleash a full network, so we can also create feature strategies for specific networks, e.g., `nano-bravo-testnet`, not only `testnet`.

### Security Checklist

### DIFF
--- a/src/sagas/networkSettings.js
+++ b/src/sagas/networkSettings.js
@@ -31,6 +31,7 @@ import {
 } from './helpers';
 import { STORE } from '../store';
 import { isWalletServiceEnabled } from './wallet';
+import { monitorFeatureFlags } from './featureToggle';
 import { logger } from '../logger';
 
 const log = logger('network-settings-saga');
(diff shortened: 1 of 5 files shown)


src/sagas/networkSettings.js
```diff
@@ -287,6 +289,10 @@ export function* persistNetworkSettings(action) {
   // Reset network name when changing networks
   yield put(setFullNodeNetworkName(''));
 
+  // After we persist the network settings, we must create a new unleash
+  // client because the network/stage might have changed
+  yield call(monitorFeatureFlags, 0, false);
```
andreabadesso, line 294: Hmmm, maybe instead of calling the `monitorFeatureFlags` saga directly from this saga we should just listen to the `networkChanged` action in the featureToggle saga and react to it...

andreabadesso (approved):

PR #730, opened 2025-08-06
feat: redirect to Dashboard after network settings success

pedroferreira1:
### Acceptance Criteria
- Redirect to Dashboard after network settings update success

https://github.com/user-attachments/assets/d99fdfb1-dea5-4b36-9433-ac269118190c

### Security Checklist

### DIFF
--- a/src/screens/NetworkSettings/CustomNetworkSettingsScreen.js
+++ b/src/screens/NetworkSettings/CustomNetworkSettingsScreen.js
@@ -154,8 +154,11 @@ export const CustomNetworkSettingsScreen = ({ navigation }) => {
     setInvalidModel(validate(form));
   };
 
-  const handleFeedbackModalDismiss = () => {
+  const handleFeedbackModalDismiss = (navigate) => {
     dispatch(networkSettingsUpdateReady());
+    if (navigate) {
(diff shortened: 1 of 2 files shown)


src/screens/NetworkSettings/CustomNetworkSettingsScreen.js
```diff
@@ -154,8 +154,11 @@ export const CustomNetworkSettingsScreen = ({ navigation }) => {
     setInvalidModel(validate(form));
   };
 
-  const handleFeedbackModalDismiss = () => {
+  const handleFeedbackModalDismiss = (navigate) => {
     dispatch(networkSettingsUpdateReady());
+    if (navigate) {
+      navigation.navigate('Dashboard');
```
tuliomir, line 160: suggestion: Use `NavigationService.resetToMain();` instead, so that the "Back" button from OS does not return to the Custom Network screen.

See its usage on App.js for a reference.

https://github.com/HathorNetwork/hathor-wallet-mobile/blob/e0f165c58b428498622751b2f54d7215f4312150/src/App.js#L779-L783

tuliomir (approved):

PR #741, opened 2025-08-13
refactor: using @lavamoat/react-native-lockdown instead of vendored SES

andreabadesso:
### Acceptance Criteria
- We should remove the vendored SES lockdown file and use the new @lavamoat/react-native-lockdown lib
- We should re-enable Hermes on our iOS builds
- We should pass the radix to all `Number.toString()` calls as SES causes them to be flagged as unsafe and to throw

### Security Checklist

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -84,8 +84,8 @@ android {
         applicationId "network.hathor.wallet"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 99
-        versionName "0.32.2"
+        versionCode 101
+        versionName "0.33.0"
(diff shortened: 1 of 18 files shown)


src/sagas/ses.js
```diff
@@ -5,77 +5,30 @@
...
-    disableSes();
-  }
-
-  if (unleashEnabled && !storageEnabled) {
-    log.debug('Unleash is enabled and storage is not, enabling SES!');
-    enableSes();
-  }
 }
 
 export function* isSESEnabled() {
```
pedroferreira1, line 34: Do you still use this method?

pedroferreira1 (approved):

src/sagas/ses.js
andreabadesso: We don't, removed, thanks

PR #751, opened 2025-10-02
refactor: using the library method to get the blueprint id

andreabadesso:
### Acceptance Criteria

### Security Checklist

### DIFF
--- a/src/sagas/reown.js
+++ b/src/sagas/reown.js
@@ -64,7 +64,6 @@ import { eventChannel } from 'redux-saga';
 import { get, values } from 'lodash';
 import { Core } from '@walletconnect/core';
 import { WalletKit } from '@reown/walletkit';
-import { ncApi } from '@hathor/wallet-lib';
 import {
   TriggerTypes,
   TriggerResponseTypes,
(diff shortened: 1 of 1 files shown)


tuliomir (approved):

src/sagas/reown.js
```diff
@@ -473,11 +473,27 @@ export function* processRequest(action) {
...
+        let shouldIgnoreRetry = false;
+        for (let i = 0; i < dontRetryErrors.length; i += 1) {
+          if (e.message.indexOf(dontRetryErrors[i]) > -1) {
+            shouldIgnoreRetry = true;
+            break;
+          }
+        }
+
+        let retry = false;
+        if (!shouldIgnoreRetry) {
```
tuliomir, line 490: suggestion(non-blocking): This took me a while to understand, as `retry` and `ignore` are already words that carry negative meanings, and the `!` made this more confusing.

I suggest renaming `shouldIgnoreRetry` to `shouldRetry`.

```js
let shouldRetry = true;
... if (e.message.indexOf() > -1) {
  shouldRetry = false;
  break;
} ...

...

let retry = false;
if (shouldRetry) {
  retry = yield call(...)
}
```

Maybe it would be even better to rename `retry` to `retryHandlerResponse` to better inform what this variable carries.

tuliomir (approved):

tuliomir (approved):

PR #752, opened 2025-10-06
feat: token swap screens

r4mmer:
### Acceptance Criteria

- Create initial screens for the token swap protocol

### Security Checklist

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -104,6 +104,9 @@ import UnifiedQRScanner from './screens/UnifiedQRScanner';
 import RegisterOptionsScreen from './screens/RegisterOptionsScreen';
 import { NavigationSerializingProvider } from './hooks/navigation';
 import { SendTransactionRequestScreen } from './screens/Reown/SendTransactionRequestScreen';
+import TokenSwap from './screens/TokenSwap';
+import TokenSwapLoadingScreen from './screens/TokenSwapLoadingScreen';
+import TokenSwapTokenList from './screens/TokenSwapTokenList';
 
 /**
  * This Stack Navigator is exhibited when there is no wallet initialized on the local storage.
@@ -159,6 +162,38 @@ const DashboardStack = () => {
  * the permission is defined.
  */
 const CameraPermissionScreen = () => null;
+
+/**
+ * Stack of screens dedicated to the token swap process
+ */
+const SwapStack = ({ navigation }) => {
+  const Stack = createStackNavigator();
+  const [initialRoute, setInitialRoute] = useState('TokenSwapLoadingScreen');
+  /** @type {null|'loading'|'loaded'|'error'} */
+  const allowedTokensStatus = useSelector((state) => state.tokenSwap.loadAllowedTokensStatus);
+
+  useEffect(() => {
+    // When we load the tokens, move to the token swap main screen
+    if (allowedTokensStatus === 'loaded') {
+      setInitialRoute('TokenSwap');
+      navigation.replace('TokenSwap');
+    }
+  }, [allowedTokensStatus]);
+
+  return (
+    <Stack.Navigator
+      initialRouteName={initialRoute}
+      screenOptions={{
+        headerShown: false,
+      }}
+    >
+      <Stack.Screen name='TokenSwapLoadingScreen' component={TokenSwapLoadingScreen} />
+      <Stack.Screen name='TokenSwap' component={TokenSwap} />
+      <Stack.Screen name='TokenSwapListInputToken' component={TokenSwapTokenList('input')} />
+      <Stack.Screen name='TokenSwapListOutputToken' component={TokenSwapTokenList('output')} />
+    </Stack.Navigator>
+  );
+};
 
 /**
  * Stack of screens dedicated to the token sending process
@@ -369,6 +404,7 @@ const RegisterNanoContractStack = ({ navigation }) => {
 const tabBarIconMap = {
   Home: 'icDashboard',
   Send: 'icSend',
+  Swap: 'icSettings',
   Receive: 'icReceive',
   Settings: 'icSettings',
 };
@@ -407,6 +443,7 @@ const TabNavigator = () => {
         name='Home'
         component={IS_MULTI_TOKEN ? DashboardStack : MainScreen}
       />
+      <Tab.Screen name='Swap' component={SwapStack} />
       <Tab.Screen name='Send' component={SendStack} />
       <Tab.Screen name='Receive' component={ReceiveScreen} />
       <Tab.Screen name='Settings' component={Settings} />
--- a/src/reducers/reducer.js
+++ b/src/reducers/reducer.js
@@ -538,6 +538,21 @@ const initialState = {
    * the wallet saga.
    */
   fullNodeNetworkName: null,
+  /**
+   * 
+   */
+  tokenSwap: {
+    allowedTokens: null,
+    // null | 'loading' | 'loaded' | 'error'
+    loadAllowedTokensStatus: null,
+    inputToken: null,
+    inputTokenAmount: 0,
+    outputToken: null,
+    outputTokenAmount: 0,
+    swapPathQuote: null,
+    // null | 'loading' | 'loaded' | 'error'
+    loadSwapPathQuoteStatus: null,
+  },
 };
 
 export const reducer = (state = initialState, action) => {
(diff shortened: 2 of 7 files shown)


src/screens/TokenSwap.js
```diff
@@ -0,0 +1,394 @@
...
+  } = useSelector((state) => state.tokenSwap);
+  const { decimalPlaces } = useSelector((state) => ({
+    decimalPlaces: state.serverInfo?.decimal_places
+  }));
+
+  const navigation = useNavigation();
+
+  useEffect(() => {
+    if (!inputToken) {
+      dispatch(tokenSwapSetInputToken(allowedTokens[0]));
```
tuliomir: issue: For now this is not a problem as the api responses are mocked, but in the future could the `allowedTokens` be an empty array for some edge case reason?

If so, we should have at least a simple protection to avoid crashing the app here.

src/screens/TokenSwap.js
```diff
@@ -0,0 +1,394 @@
...
+      setInputTokenAmount(0n);
+
+      setSwapDirection('input');
+    }
+
+    // switch the tokens being swapped
+    dispatch(tokenSwapSwitchTokens());
+  };
+
+  const isButtonDisabled = () => {
```
tuliomir: chore: Specify which button for clarity
```suggestion
  const isReviewButtonDisabled = () => {
```

src/screens/TokenSwap.js
r4mmer: This is an interesting observation, since this is a "swap" service it makes no sense to deploy a list with less than 2 tokens so I didn't even consider this possibility because having 1 token on the list makes the entire feature unavailable.

Maybe we should check for this length on the saga that requests the allowed tokens list and just put an error in this case, what do you think?

tuliomir: I agree that's the best approach! Rejecting the received allowedTokens lists if it has less than two tokens in the saga.

What I'm trying to avoid here is a crash for something so small such as an array out of bounds.

src/screens/TokenSwapReview.js
```diff
@@ -98,15 +93,15 @@ const TokenSwapReview = () => {
...
+  const onSwapButtonPress = () => {
     const pinParams = {
       cb: executeSend,
       canCancel: true,
       screenText: t`Enter your 6-digit pin to authorize operation`,
       biometryText: t`Authorize operation`,
       biometryLoadingText: t`Building transaction`,
     };
-    navigation.navigate('PinScreen', pinParams);
+    NavigationService.navigate('PinScreen', pinParams);
```
tuliomir: chore: I'd recommend agains using the NavigationService for common navigations: for that it's better that we use the `navigation` obtained from `useNavigation()` as it was before.

The only use for it is to ask the app itself to reset to main and other navigations that the screen needs to ask for the global Navigation object to do in its behalf.

src/reducers/reducer.js
```diff
@@ -2259,3 +2292,96 @@ export const onSetCreateNanoContractCreateTokenTxStatus = (state, { payload }) =
...
+export const onTokenSwapSetAllowedTokens = (state, { payload }) => {
+  return {
+    ...state,
+    tokenSwap: {
+      ...state.tokenSwap,
+      allowedTokens: payload.allowedTokens,
+      contractId: payload.contractId,
+      loadAllowedTokensStatus: 'loaded',
+      inputToken: payload.allowedTokens[0],
+      outputToken: payload.allowedTokens[1],
```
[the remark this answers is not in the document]
tuliomir: No problem, now that the full context is clear I think this name is fine.

src/screens/TokenSwap.js
r4mmer: Done

tuliomir (approved):

src/components/TokenSelect.js
```diff
@@ -47,7 +47,7 @@ const TokenSelect = (props) => {
     }
 
     const balance = get(props.tokensBalance, `${item.uid}.data.available`, 0);
-    const tokenState = get(props.tokensBalance, `${item.uid}.status`, 'loading');
+    const tokenState = get(props.tokensBalance, `${item.uid}.status`, props.ignoreLoading ? 'ready' : 'loading');
```
pedroferreira1: Is this a new `props`? We should add it to the docstring

src/screens/TokenSwapTokenList.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+export default function TokenSwapTokenList(direction) {
+  return () => {
+    const dispatch = useDispatch();
+    const navigation = useNavigation();
+    const params = useParams();
+    const token = params.token ?? null;
+
+    const tokenMetadata = useSelector(state => state.tokenMetadata);
+    const tokensBalance = useSelector(state => state.tokensBalance);
```
pedroferreira1: Do we have the balance of unregistered tokens here?

src/App.js
```diff
@@ -160,6 +165,39 @@ const DashboardStack = () => {
...
+  const [initialRoute, setInitialRoute] = useState('TokenSwapLoadingScreen');
+  /** @type {null|'loading'|'loaded'|'error'} */
+  const allowedTokensStatus = useSelector((state) => state.tokenSwap.loadAllowedTokensStatus);
+
+  useEffect(() => {
+    // When we load the tokens, move to the token swap main screen
+    if (allowedTokensStatus === 'loaded') {
+      setInitialRoute('TokenSwap');
+      navigation.replace('TokenSwap');
+    }
```
pedroferreira1: I'm not fully against this but we usually execute the loading on the screen itself, so we can control the components loading/error/success on the same place depending on the redux state.

What do you think about it?

src/sagas/tokenSwap.js
```diff
@@ -0,0 +1,107 @@
...
+} from '../actions';
+
+const ALLOWED_TOKENS_URL = 'https://httpbin.org/json'
+
+export function* handleFetchAllowedTokensRequest() {
+  yield put(tokenSwapSetAllowedTokens(
+    'fake-contract-id',
+    [
+      { symbol: 'HTR', name: 'Hathor', uid: '00' },
+      { symbol: 'CTHOR', name: 'Cathor', uid: '00000000f76262bb1cca969d952ac2f0e85f88ec34c31f26a13eb3c31e29d4ed' },
```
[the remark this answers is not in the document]
r4mmer: No, I have placed the icon as optional on the token schema because the figma was not prepared for token icons.

src/sagas/tokenSwap.js
```diff
@@ -0,0 +1,107 @@
...
+    const swapQuote = yield call(getSwapQuote);
+    // put swap quote and calculate amount?
+    yield put(tokenSwapFetchSwapQuoteSuccess(swapQuote));
+  } catch (err) {
+    // Some error happened
+    console.error(err);
+    yield put(tokenSwapFetchSwapDataError());
+  } finally {
+    if (yield cancelled()) {
+      // This task has been cancelled, we should just ignore changes?
```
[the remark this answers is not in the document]
r4mmer: The `takeLatest` at the end of the file ensures that only 1 execution of the saga (the latest one) is running at any time.
When a new request arrives it cancels the old one and start another.

This is to ensure that we do not have a race condition with 2 requests for quotes with different values.

src/App.js
r4mmer: I was copying the behavior of the `SendStack` but I also think that the replace is not doing much here.

src/reducers/reducer.js
```diff
@@ -2259,3 +2299,96 @@ export const onSetCreateNanoContractCreateTokenTxStatus = (state, { payload }) =
...
+    },
+  };
+};
+
+export const onTokenSwapFetchAllowedTokensError = (state) => ({
+  ...state,
+  tokenSwap: {
+    ...state.tokenSwap,
+    allowedTokens: null,
```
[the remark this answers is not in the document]
r4mmer: Done

src/reducers/reducer.js
```diff
@@ -538,6 +538,27 @@ const initialState = {
...
+   * @property {null|'loading'|'loaded'|'error'} loadAllowedTokensStatus Status of the allowed tokens request
+   * @property {null|TokenData} inputToken Token being traded in
+   * @property {null|TokenData} outputToken Token being swapped out
+   * @property {null|TokenSwapQuote} swapPathQuote Latest quote from the pool manager for the swap proposed
+   * @property {null|'loading'|'loaded'|'error'} loadSwapPathQuoteStatus Status of the quote request
+   */
+  tokenSwap: {
+    allowedTokens: null,
+    contractId: null,
+    // null | 'loading' | 'loaded' | 'error'
```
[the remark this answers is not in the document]
r4mmer: Done

PR #765, opened 2025-10-30
token swap: connect screens with calls

r4mmer:
### Acceptance Criteria

- Remove all mocks and connect the screens with the method utils

### Security Checklist

### DIFF
--- a/src/sagas/featureToggle.js
+++ b/src/sagas/featureToggle.js
@@ -95,8 +95,8 @@ export function* handleToggleUpdate() {
 }
 
 export function* monitorFeatureFlags(currentRetry = 0, createRoutine = true) {
-  let stage; let
-    fullNetwork;
+  let stage;
+  let fullNetwork;
   // We try to get the network settings from the storage (which is updated as soon
   // as we persist the networkSettings update) and if it's empty, we get the default state
   const networkSettingsStorage = STORE.getItem(networkSettingsKeyMap.networkSettings);
@@ -113,7 +113,7 @@ export function* monitorFeatureFlags(currentRetry = 0, createRoutine = true) {
     properties: {
       platform: Platform.OS,
       stage,
-      fullNetwork,
+      network: fullNetwork,
       appVersion,
     },
   };


src/screens/TokenSwap.js
```diff
@@ -289,24 +316,32 @@ const TokenSwap = () => {
                 </View>
               </View>
 
-              { showQuote && (
+              { showQuote && quote && (
```
r4mmer: There is a split second screen change when cleaning up the state where `quote` is `null` but `showQuote` is still `true`.
This makes an error since we use properties of the quote in this block.

src/screens/TokenSwap.js
```diff
@@ -204,13 +218,26 @@ const TokenSwap = () => {
...
+
+  const getAmountString = (amount, token) => {
+    return `${renderValue(amount, false)} ${token.symbol}`;
+  };
+
+  const getConversionRate = (swapQuote) => {
+    if (!swapQuote) {
+      return null;
+    }
+    return `${getAmountString(100*Number(swapQuote.amount_out)/Number(swapQuote.amount_in), outputToken)} = ${getAmountString(100, inputToken)}`;
```
tuliomir: question(non-blocking): Why multiply by 100? Won't `renderValue()` deal with the decimal cases already?

src/screens/TokenSwap.js
```diff
@@ -204,13 +218,26 @@ const TokenSwap = () => {
     return t`Balance: ${amount}`;
   };
 
-  const onReviewButtonPress = () => {
+  const getAmountWithSlippage = (amount, direction) => {
+    return calcAmountWithSlippage(direction, amount, TOKEN_SWAP_SLIPPAGE);
+  };
+
+  const getAmountString = (amount, token) => {
+    return `${renderValue(amount, false)} ${token.symbol}`;
```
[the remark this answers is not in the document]
r4mmer: The prootocol may permit using an NFT as token in the swap but I don't think we will be having NFT liquidity pools, so I assume only non-NFT tokens will be used.

PR #770, opened 2025-11-03
fix: show error modal if it fails to register nano

pedroferreira1:
### Acceptance Criteria
- Show error modal if it fails to register nano in the Reown modal screen

https://github.com/user-attachments/assets/bec37ae3-9237-4bc3-a7f5-d4560c308e40

Previously (with the bug)

https://github.com/user-attachments/assets/bd6feacd-8d52-408a-a97e-68ab3a659fe4

### Security Checklist

### DIFF
--- a/src/components/Reown/NanoContract/BaseNanoContractRequest.js
+++ b/src/components/Reown/NanoContract/BaseNanoContractRequest.js
@@ -424,14 +424,12 @@ export const BaseNanoContractRequest = ({
 
   if (notRegistered && hasNcRegisterFailed) {
     return (
-      <>
-        <FeedbackModal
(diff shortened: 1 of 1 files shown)


src/components/Reown/NanoContract/BaseNanoContractRequest.js
```diff
@@ -418,6 +422,19 @@ export const BaseNanoContractRequest = ({
     );
   }
 
+  if (notRegistered && hasNcRegisterFailed) {
```
pedroferreira1: I had to move up the modal because it wasn't showing up. It was always getting to the condition below

andreabadesso (approved):

PR #798, opened 2025-11-14
fix: edge cases rendering issues

r4mmer:
### Acceptance Criteria

- Rendering should not break when the contract returns a strange response (like swapping tokens for 0 tokens on the input/output)
- Fix issue with IOS url params serialization
- When editing an input on the swap, lock the other input.
- Do not allow token swaps with 0 amount.
- Use runFromMining method when sending nano contract transactions.
- Small changes to improve UX 

### Security Checklist

### DIFF
--- a/src/screens/TokenSwapReview.js
+++ b/src/screens/TokenSwapReview.js
@@ -5,7 +5,7 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-import React, { useState, useEffect } from 'react';
+import React, { useState } from 'react';
 import {
   Keyboard,
   KeyboardAvoidingView,
@@ -112,7 +112,7 @@ const TokenSwapReview = () => {
         sendTransaction,
         promise,
       });
-    } catch(err) {
+    } catch (err) {
       console.error(err);
       this.exitOnError();
     } finally {
--- a/src/styles/themes.js
+++ b/src/styles/themes.js
@@ -48,7 +48,7 @@ export const COLORS = {
   /**
    * Light shadow for adding details on the UI
    */
-   lightShadow: '#ababab',
+  lightShadow: '#ababab',
   /**
    * @type {string} More washed down version of the text
    */


src/screens/TokenSwap.js
```diff
@@ -190,6 +205,21 @@ const TokenSwap = () => {
     dispatch(tokenSwapSwitchTokens());
   };
 
+  const checkQuotedAmount = () => {
```
tuliomir: suggestion: Add docstrings to this method, explaining what this return boolean means.

src/screens/TokenSwap.js
```diff
@@ -363,11 +383,12 @@ const styles = StyleSheet.create({
   },
   screenContent: {
     flex: 1,
-    backgroundColor: COLORS.lowContrastDetail,
+    backgroundColor: '#fff',
```
tuliomir: issue: We must use the `COLORS` object to centralize all colors in the app. This is a necessary step towards having dark mode in the future.

There are two more hardcoded colors in this file that should also centralize their values in `COLORS`.

src/screens/TokenSwapReview.js
```diff
@@ -67,13 +71,26 @@ const TokenSwapReview = () => {
...
+      quote.direction === 'input' ? quote.amount_in : quote.amount_out,
+      tokenIn.uid,
+      tokenOut.uid
+    ));
     navigation.goBack();
   };
 
+  useEffect(() => {
+    setLoading(false);
+  }, [modal]);
```
tuliomir: chore: Add an explanation for this effect, since it changes the `loading` behavior based on Modals, but there is a feedback modal that has a spinner loader below. It's not clear in the code what are the expected behaviors.

pedroferreira1 (approved):

src/screens/TokenSwapReview.js
r4mmer: This was to avoid leaving the loading open, but changed to a try..finally on the only place that uses setLoading(true)

src/screens/TokenSwap.js
r4mmer: Done

PR #802, opened 2025-11-18
upgrade packages page size 16

r4mmer:
### Acceptance Criteria

### Security Checklist

### DIFF
--- a/src/screens/Security.js
+++ b/src/screens/Security.js
@@ -17,7 +17,12 @@ import { t } from 'ttag';
 import HathorHeader from '../components/HathorHeader';
 import baseStyle from '../styles/init';
 import {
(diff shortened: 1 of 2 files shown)


andreabadesso (approved):

src/screens/Security.js
```diff
@@ -74,11 +74,16 @@ export class Security extends React.Component {
         this.onSafeBiometryDisabled();
       }
     } else {
-      this.setState({ biometryEnabled: value });
-      setBiometryEnabled(value);
+      this.onOldBiometryEnabled();
```
pedroferreira1: Shouldn't we call it only `if value === true`?

And on disable shouldn't we call 

```
this.setState({ biometryEnabled: false });
    setBiometryEnabled(false);
```

pedroferreira1 (approved):

PR #805, opened 2025-11-27
feat: improve UI for token swap screens

pedroferreira1:
### Acceptance Criteria
- Improve UI for token swap screens
  - Align amount text input to the left
  - Make divider icon a circle 
  - Reduce shadow
  - Add button margin
  - Update details text
  - Review back button goes back
  - Review screen arrow down icon purple

https://github.com/user-attachments/assets/5da3d330-c742-43d3-9e83-a50830e914df

### Security Checklist

### DIFF
--- a/src/screens/TokenSwap.js
+++ b/src/screens/TokenSwap.js
@@ -237,10 +237,6 @@ const TokenSwap = () => {
       || !checkQuotedAmount()
   );
 
-  const renderGhostElement = () => (
-    <View style={{ width: 80, height: 40 }} />
-  );
-
   const getAvailableString = (token) => {
     if (!token) {
       return '';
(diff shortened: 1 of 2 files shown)


src/screens/TokenSwap.js
```diff
@@ -425,10 +426,10 @@ const styles = StyleSheet.create({
     // For IOS
     shadowColor: COLORS.lightShadow,
     shadowOffset: { width: 0, height: 2 },
-    shadowOpacity: 0.5,
+    shadowOpacity: 0.1,
```
tuliomir: question(non-blocking): Would it be better to use `textColorShadowLighter` instead of applying a color and an opacity? It was designed only with text color in mind, but maybe it's useful here as well.

https://github.com/HathorNetwork/hathor-wallet-mobile/blob/87b5884eac4acb8abc8cd192055323f11c47f03f/src/styles/themes.js#L42-L47

r4mmer (approved):

PR #846, opened 2026-03-15
feat: persist registered tokens across network changes

andreabadesso:
https://github.com/user-attachments/assets/04a11873-b4ce-43b4-b2c0-7436ea0bffc6

### Acceptance Criteria
- Registered tokens are saved per network, keyed by genesis block hash
- When switching networks, tokens from the previous network are preserved in AsyncStorage
- When switching back to a previously used network, tokens are automatically restored
- Adding or removing a token updates the persisted tokens for the current network
- Two different testnets are correctly distinguished by their genesis block hash

### Security Checklist

## Summary by CodeRabbit

* **New Features**
  * Per-network token persistence: tokens are saved per network and restored automatically on wallet start or network switch.
  * Token snapshots are updated whenever tokens change to keep persisted per-network state current.

* **Behavioral Change**
  * Network persist now waits for tokens to be saved before proceeding with cleanup, improving reliability during network switches.

* **Tests**
  * Comprehensive tests added for token save/restore, per-network isolation, and snapshot/update flows.


src/sagas/tokens.js
```diff
@@ -458,12 +467,116 @@ export function* fetchTokenData(tokenId, force = false) {
...
+    if (!genesisHash) return;
+
+    const wallet = yield select((state) => state.wallet);
+    if (!wallet || !wallet.isReady()) return;
+
+    const tokens = yield call(getRegisteredTokens, wallet, true);
+    yield call([STORE, STORE.saveTokensForNetwork], genesisHash, tokens);
+  } catch (e) {
+    log.error('Error updating network token snapshot:', e);
+  }
```
pedroferreira1: This part of the method is basically the same as `saveNetworkTokens`. Do we need both?

andreabadesso: Refactored, can you please check?

pedroferreira1 (approved):

tuliomir (approved):

PR #857, opened 2026-04-02
feat: add batch reject for pending Reown dApp requests

andreabadesso:
https://github.com/user-attachments/assets/2bcaabf9-d10d-466c-8103-2ae56b19b42a

### Acceptance Criteria
- [ ] When a dApp sends multiple requests, a dark floating banner appears at the top showing the pending count and a "REJECT ALL" button
- [ ] Tapping the banner expands a bottom-sheet carousel showing each pending request (dApp icon, name, method type)
- [ ] Tapping "REJECT ALL" on the banner or carousel rejects all pending requests, closes the active request screen, and returns to the dashboard
- [ ] The banner only appears when there are 2+ pending requests
- [ ] The banner updates in real-time as new requests arrive (polled every 2s)
- [ ] After rejecting all, no further request modals pop up
- [ ] Single-request flow is unchanged (no banner when only 1 request)
- [ ] Reown Sessions screen shows a red pending requests banner with count and "Review" link to a dedicated Pending Requests management screen
- [ ] Pending Requests screen allows per-request reject and bulk reject

### Follow-up
- #858 — Allow selecting a specific request from the carousel to review

### Security Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a pending session requests screen to view and manage incoming Reown requests with dApp details and method information.
  * Added a floating notification banner alerting users when pending requests exist.
  * Added ability to reject individual requests or all pending requests in bulk.
  * Added automatic periodic refresh of the pending requests list to keep data current.


src/components/Reown/NanoContract/BaseNanoContractRequest.js
```diff
@@ -476,6 +477,7 @@ export const BaseNanoContractRequest = ({
                 secondary
                 danger
               />
+              <RejectAllButton />
```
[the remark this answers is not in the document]
andreabadesso: The `RejectAllButton` component was removed entirely in a subsequent commit — the global overlay approach replaces it. This comment is no longer applicable.

src/components/Reown/SendTransactionRequest.js
```diff
@@ -531,6 +532,7 @@ export const SendTransactionRequest = ({ sendTransactionRequest, onAccept, onRej
                       secondary
                       danger
                     />
+                    <RejectAllButton />
```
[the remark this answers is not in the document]
andreabadesso: Same as above — `RejectAllButton` was removed. The global overlay handles rejection now.

src/screens/Reown/PendingRequests.js
```diff
@@ -0,0 +1,476 @@
...
+  useEffect(() => {
+    dispatch(reownFetchPendingRequests());
+  }, [dispatch]);
+
+  const handleRejectAll = useCallback(() => {
+    setRejecting(true);
+    dispatch(reownRejectAllPendingRequests());
+    // The saga will update the pending requests list
+    setTimeout(() => setRejecting(false), 500);
+  }, [dispatch]);
```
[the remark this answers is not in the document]
andreabadesso: Fixed in 1531da38 — timeout is now stored in a ref and cleared on unmount.

src/components/Reown/ReownPendingOverlay.js
```diff
@@ -0,0 +1,570 @@
...
+                </Text>
+              </View>
+            )}
+            <View style={styles.bannerTextWrap}>
+              <Text style={styles.bannerCount}>{count}</Text>
+              <Text style={styles.bannerLabel}>
+                {count === 1
+                  ? t`pending request`
+                  : t`pending requests`}
+              </Text>
```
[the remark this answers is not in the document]
andreabadesso: Fixed in 1531da38 — removed the dead `count === 1` branch since `visible` requires `count > 1`.

andreabadesso:
**Re: CodeRabbit review (latest round)**

1. **`pendingPollTask` leak on early returns** (Major) — Fixed in b6a51805. Added `yield cancel(pendingPollTask)` before the early return when `requestSession` is missing, and before all 5 recursive retry paths. The normal exit path already had the cancel.

2. **Double-response on batch reject** (Nitpick) — Acknowledged. WalletKit handles duplicate responses gracefully (second call errors but is caught and logged). Adding a batch-reject flag would add complexity for a harmless edge case. Leaving as-is.

src/components/Reown/ReownPendingOverlay.js
```diff
@@ -0,0 +1,580 @@
...
+      return {
+        color: COLORS.feedbackWarning300,
+        bg: COLORS.feedbackWarning100,
+      };
+    case 'sign':
+      return { color: COLORS.primary, bg: COLORS.primaryOpacity10 };
+    default:
+      return {
+        color: COLORS.positiveBalanceColor,
+        bg: 'hsla(180, 80%, 95%, 1)',
```
tuliomir: suggestion(non-blocking): I'd suggest adding this as a constant somehow, like the colors we already have.

src/components/Reown/ReownPendingOverlay.js
```diff
@@ -0,0 +1,580 @@
...
+        style={[
+          styles.banner,
+          { transform: [{ translateY: bannerAnim }] },
+        ]}
+        pointerEvents='box-none'
+      >
+        <TouchableOpacity
+          style={styles.bannerTouchable}
+          onPress={() => setExpanded(true)}
+          activeOpacity={0.85}
```
tuliomir: suggestion(non-blocking): Another hardcoded value that would benefit from a constant.

src/components/Reown/ReownPendingOverlay.js
```diff
@@ -0,0 +1,580 @@
...
+            />
+
+            {/* Page dots */}
+            <PageDots count={count} activeIndex={activeIndex} />
+
+            {/* Actions */}
+            <TouchableOpacity
+              style={styles.rejectAllBtn}
+              onPress={onRejectAll}
+              activeOpacity={0.8}
```
tuliomir: suggestion(non-blocking): Same here. Note that the opacity of these elements is not the same. If we had a descriptive constant name it would be trivial to know why they differ.

There are other such elements in this screen I could comment the same idea, but this is not blocking so I'll just mention those two.

src/components/Reown/ReownPendingOverlay.js
```diff
@@ -0,0 +1,580 @@
...
+  },
+  bannerTouchable: {
+    flexDirection: 'row',
+    alignItems: 'center',
+    justifyContent: 'space-between',
+    backgroundColor: COLORS.textColor,
+    borderRadius: 14,
+    paddingVertical: 10,
+    paddingHorizontal: 14,
+    shadowColor: '#000',
```
tuliomir: chore: Here we should really use the color constants. Some elements below also need adjusting.

src/screens/Reown/PendingRequests.js
```diff
@@ -0,0 +1,481 @@
...
+    case 'read': return COLORS.positiveBalanceColor;
+    default: return COLORS.midContrastDetail;
+  }
+};
+
+const getCategoryBgColor = (category) => {
+  switch (category) {
+    case 'write': return COLORS.feedbackWarning100;
+    case 'sign': return COLORS.primaryOpacity10;
+    case 'read': return 'hsla(180, 80%, 95%, 1)';
```
tuliomir: question(non-blocking): Why use a hardcoded value instead of a color constant?

src/components/Reown/ReownPendingOverlay.js
andreabadesso: Good call. Promoted this to a shared constant `COLORS.positiveBalanceBgColor` in `src/styles/themes.js` (paired with the existing `positiveBalanceColor`), and used it here and in `PendingRequests.js`. Addressed in 740bc2b0.

src/components/Reown/ReownPendingOverlay.js
andreabadesso: Extracted as the `ACTIVE_OPACITY_BANNER` constant at the top of the file (with a comment explaining the rationale). Addressed in 740bc2b0.

src/components/Reown/ReownPendingOverlay.js
andreabadesso: Agreed — the differing opacities now live as named constants at the top of the file: `ACTIVE_OPACITY_BANNER` (0.85), `ACTIVE_OPACITY_PRIMARY_BUTTON` (0.8), and `ACTIVE_OPACITY_SECONDARY_BUTTON` (0.7), each with a short comment on when to use which. Also applied to the other activeOpacity usages in this file. Addressed in 740bc2b0.

src/components/Reown/ReownPendingOverlay.js
andreabadesso: Done — `shadowColor` now uses `COLORS.black`, and the rgba literals for the banner icon background, banner label text, and overlay backdrop were extracted to named constants (`WHITE_OPACITY_15`, `WHITE_OPACITY_75`, `BACKDROP_SCRIM`). Addressed in 740bc2b0.

src/screens/Reown/PendingRequests.js
andreabadesso: No good reason — replaced with `COLORS.positiveBalanceBgColor` (the new shared constant in `themes.js`). Addressed in 740bc2b0.

tuliomir (approved):

src/components/Reown/ReownPendingOverlay.js
```diff
@@ -0,0 +1,598 @@
...
+      toValue: expanded ? 1 : 0,
+      duration: 250,
+      useNativeDriver: true,
+    }).start();
+  }, [expanded, overlayAnim]);
+
+  const onRejectAll = useCallback(() => {
+    dispatch(reownRejectAllPendingRequests());
+    // Unblock the currently-processing request saga
+    dispatch(reownReject());
```
pedroferreira1: The reown `rejectAllPendingRequests` method already triggers `reownReject`. No need to call it here as well

src/sagas/reown.js
```diff
@@ -1583,6 +1646,84 @@ export function* onGetUtxosRequest({ payload }) {
...
+      log.error('[rejectAllPendingRequests] Error rejecting request', e);
+    }
+  }
+
+  // Unblock the currently-processing request in the saga queue
+  // (it's waiting on REOWN_ACCEPT/REOWN_REJECT in a race)
+  yield put(reownReject());
+
+  // Clear the pending requests list
+  yield put(setReownPendingRequests([]));
```
pedroferreira1: Is it possible that a new request arrives while you were rejecting the previous ones? In that case, we could have here `getPendingSessionRequests` before, instead of just setting an empty array

pedroferreira1 (approved):

src/components/Reown/ReownPendingOverlay.js
andreabadesso: Good catch! Removed the redundant `dispatch(reownReject())` from the component — also dropped the now-unused import.

src/sagas/reown.js
andreabadesso: Yes, good point. Replaced `setReownPendingRequests([])` with a fresh `walletKit.getPendingSessionRequests()` call so any request that arrived during the loop is picked up correctly.

PR #859, opened 2026-04-06
feat: easy tokens import

raul-oliveira:
### Acceptance Criteria
- Detect unregistered tokens the wallet has interacted with (via transaction history)
- Display a list of unregistered tokens with name, symbol, balance, and short UID
- Allow the user to select one or more tokens for import
- Show a confirmation screen before registering selected tokens
- Provide a modal with loading/success/error states during import
- Show a banner on the Dashboard when new unregistered tokens are detected
- Allow opening token details in the explorer from the token list
- Add an "Import tokens" entry in Settings for manual access
- Detect new unregistered tokens from incoming transactions in real-time

### Security Checklist

#### Token Banner
image

#### Import token screen
image

##### Confirm state
image

##### Success state
image

##### Error state
image

#### Empty state
image

#### Settings screen
image

## Summary by CodeRabbit

* **New Features**
  * Token import flow: discover unregistered tokens, select tokens, confirm import, and view import progress, success, or error states (modal + progress UI).
  * Dashboard banner, Settings entry, and dedicated screens to surface and manage token import.
  * Token list UI, token row component with selection and “open in explorer” action, and a lightweight token list card/row layout.

* **Localization**
  * Added Danish, Brazilian Portuguese, Russian, and base translation entries for the token import UI and related strings.

* **Style**
  * New theme colors for info banners, controls, muted text, and tag surfaces.


src/components/ImportTokensModal.js
```diff
@@ -0,0 +1,114 @@
...
+} from 'react-native';
+import { t } from 'ttag';
+import NewHathorButton from './NewHathorButton';
+import HathorModal from './HathorModal';
+import { COLORS } from '../styles/themes';
+import checkIcon from '../assets/images/icCheckBig.png';
+import errorIcon from '../assets/images/icErrorBig.png';
+
+/**
+ * A modal overlay with 3 states for the token import flow:
```
raul-oliveira: remove this description

src/components/ImportTokensModal.js
```diff
@@ -0,0 +1,114 @@
...
+ * - Loading (importing): spinner + "Adding tokens..."
+ * - Success: teal checkmark circle + "New tokens added!" + "SEE ALL TOKENS" button
+ * - Error: red X circle + "Add tokens failed" + "TRY AGAIN" button
+ *
+ * @param {Object} props
```
raul-oliveira, line 21: keep only this  part of the description, the standard jsdocs.

src/components/ImportTokensModal.js
```diff
@@ -0,0 +1,114 @@
...
+ * @param {'importing'|'success'|'error'} props.status - Current import status
+ * @param {Function} props.onDismiss - Called when "SEE ALL TOKENS" is pressed (success)
+ * @param {Function} props.onRetry - Called when "TRY AGAIN" is pressed (error)
+ */
+const ImportTokensModal = ({ status, onDismiss, onRetry }) => {
+  const renderContent = () => {
+    if (status === 'importing') {
```
raul-oliveira: move this state string to a object ImportTokensModalState

src/sagas/index.js
```diff
@@ -37,6 +38,7 @@ const sagas = {
   reownSagas: { saga: reownSagas, retryCount: 0, critical: false },
   nanoContractSagas: { saga: nanoContractSagas, retryCount: 0, critical: true },
   sesSagas: { saga: sesSagas, retryCount: 0, critical: true },
+  tokenImportSagas: { saga: tokenImportSagas, retryCount: 0, critical: true },
```
raul-oliveira: set this saga as no critical

src/sagas/tokenImport.js
```diff
@@ -0,0 +1,237 @@
...
+ * Fetch details (name, symbol) for a single token.
+ */
+function* fetchTokenDetails(wallet, uid) {
+  try {
+    const storedToken = yield call(() => wallet.storage.getToken(uid));
+    if (storedToken) {
+      return storedToken;
+    }
+  } catch (_e) {
+    // Fallback to API
```
raul-oliveira: at least log the error

src/sagas/tokenImport.js
```diff
@@ -0,0 +1,237 @@
...
+    if (unregisteredUids.length === 0) {
+      yield put(tokenImportFetchSuccess({}));
+      return;
+    }
+
+    const unregisteredTokens = {};
+    for (const uid of unregisteredUids) {
+      const details = yield call(fetchTokenDetails, wallet, uid);
+      const balance = details.balance
+        ? { available: details.balance.tokens.unlocked, locked: details.balance.tokens.locked }
```
raul-oliveira: remove this ternary if

src/sagas/tokenImport.js
```diff
@@ -0,0 +1,237 @@
...
+
+  if (!wallet || !wallet.isReady()) {
+    return;
+  }
+
+  const newTokens = {};
+  for (const uid of tokenUids) {
+    const details = yield call(fetchTokenDetails, wallet, uid);
+    const balance = details.balance
+      ? { available: details.balance.tokens.unlocked, locked: details.balance.tokens.locked }
```
raul-oliveira: remove this ternary if

src/screens/ImportTokensScreen.js
```diff
@@ -0,0 +1,354 @@
...
+import { CircleInfoIcon } from '../components/Icons/CircleInfo.icon';
+import { COLORS } from '../styles/themes';
+import { getShortHash } from '../utils';
+import { OpenInNewIcon } from '../components/Icons/OpenInNew.icon';
+
+/**
+ * ImportTokensScreen displays a list of unregistered tokens found in the wallet.
+ * The user can select one or more tokens and proceed to the confirmation screen.
+ *
+ * Redux state used:
```
raul-oliveira: remove this comment

src/screens/ImportTokensScreen.js
```diff
@@ -0,0 +1,354 @@
...
+  const explorerUrl = useSelector((state) => state.networkSettings.explorerUrl);
+
+  const [selectedUids, setSelectedUids] = useState(new Set());
+
+  const tokenList = Object.values(unregisteredTokens);
+  const hasTokens = tokenList.length > 0;
+  const hasSelection = selectedUids.size > 0;
+
+  const toggleToken = (uid) => {
+    setSelectedUids((prev) => {
```
raul-oliveira, line 74: simplify the selected state removing this set

src/screens/Dashboard.js
```diff
@@ -148,7 +149,12 @@ export const Dashboard = () => {
         isTokensSelected(currList)
         && (
           <TokenSelect
-            header={<TokensHeader isNanoEnabled={isNanoEnabled} />}
+            header={(
+              <>
+                <TokensHeader isNanoEnabled={isNanoEnabled} />
+                <TokenImportBanner />
+              </>
+            )}
```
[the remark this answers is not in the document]
raul-oliveira: This is intentional. `TokenImportBanner` exists only to surface the "import unregistered tokens" flow, which operates exclusively on the Tokens collection — Nano Contracts have a separate registration path. Mounting the banner above the Nano Contracts tab would offer an action that doesn't apply to that view, so the banner is scoped to the Tokens tab on purpose.

locale/pt-br/texts.po
```diff
@@ -457,16 +480,55 @@ msgid "FEE TOKEN"
...
+"fonte."
+
+#: src/screens/ImportTokensScreen.js:156
+msgid "Tokens found"
+msgstr "Tokens encontrados"
+
+#: src/screens/ImportTokensScreen.js:182
+msgid "Continue"
+msgstr "Continue"
+
```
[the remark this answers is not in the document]
raul-oliveira: Done — `pt-br` `msgstr` is now `Continuar` (`locale/pt-br/texts.po:530`). The `da` and `ru-ru` locales remain empty, consistent with how those files are maintained today.

src/components/ImportTokensModal.js
```diff
@@ -0,0 +1,117 @@
...
+        <View style={styles.actionContainer}>
+          <NewHathorButton
+            discrete
+            title={t`Try again`}
+            onPress={onRetry}
+          />
+        </View>
+      </>
+    );
+  };
```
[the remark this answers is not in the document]
raul-oliveira: Done — `renderContent` now routes both `IMPORTING` and `IDLE` through the spinner branch, removing the one-frame flash of error UI on first mount.

src/components/ImportTokensModal.js
```diff
@@ -0,0 +1,117 @@
...
+
+  const handleDismiss = () => {
+    if (status === ImportTokensModalState.SUCCESS && onDismiss) {
+      onDismiss();
+    } else if (status === ImportTokensModalState.ERROR && onRetry) {
+      onRetry();
+    }
```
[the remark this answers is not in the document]
raul-oliveira: Done — `handleDismiss` only calls `onDismiss`. Retry is reachable solely via the explicit "Try again" button, so dismissing the modal cannot trigger a destructive retry.

src/sagas/tokenImport.js
```diff
@@ -0,0 +1,233 @@
...
+    const details = yield call(fetchTokenDetails, wallet, uid);
+    const balance = yield call(fetchTokenBalance, wallet, uid);
+    newTokens[uid] = {
+      uid: details.uid,
+      name: details.name,
+      symbol: details.symbol,
+      balance,
+    };
+    yield delay(THROTTLE_MS);
+  }
```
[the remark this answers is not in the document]
raul-oliveira: Partial fix:

- **Null-guard applied:** `const tokenUids = action.payload ?? [];` so an undefined payload no longer throws inside the `for...of` loop.
- **Native/registered filter intentionally skipped here.** The only producer (`src/sagas/wallet.js:580-590`) already filters `NATIVE_TOKEN_UID`, `registeredUids`, and `currentUnregistered` before dispatching `TOKEN_IMPORT_NEW_TX_TOKENS`. Re-filtering inside the consumer would duplicate logic enforced at the boundary, against the project convention of trusting internal call sites.

src/sagas/tokenImport.js
```diff
@@ -0,0 +1,233 @@
...
+ * When any token gets registered (from any flow), remove it from tokenImport state.
+ */
+function* onTokenRegistered(action) {
+  const token = action.payload;
+  if (!token || !token.uid) return;
+
+  const { unregisteredTokens } = yield select((state) => state.tokenImport);
+  if (unregisteredTokens[token.uid]) {
+    yield put(tokenImportSuccess([token.uid]));
+  }
```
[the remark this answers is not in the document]
raul-oliveira: Done — `onTokenRegistered` now dispatches `tokenImportRemoveFromList([token.uid])` instead of `tokenImportSuccess`. Cleanup is decoupled from `importStatus` so unrelated registrations no longer flip the modal to SUCCESS.

src/sagas/tokenImport.js
```diff
@@ -0,0 +1,233 @@
...
+}
+
+export function* saga() {
+  yield takeLatest(types.TOKEN_IMPORT_FETCH_REQUESTED, fetchUnregisteredTokens);
+  yield takeLatest(types.TOKEN_IMPORT_NEW_TX_TOKENS, handleNewTokenDetection);
```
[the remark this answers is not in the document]
raul-oliveira: Done — switched to `takeEvery(types.TOKEN_IMPORT_NEW_TX_TOKENS, handleNewTokenDetection)` (`src/sagas/tokenImport.js:221`). Bursts of incoming transactions can no longer cancel an in-flight handler and silently drop accumulated UIDs.

src/screens/ConfirmImportScreen.js
```diff
@@ -0,0 +1,228 @@
...
+ *
+ * Route params:
+ * @param {Array} tokens - Array of token objects to import
+ *   Each token: { uid, name, symbol, balance }
+ */
+const ConfirmImportScreen = () => {
+  const dispatch = useDispatch();
+  const navigation = useNavigation();
+  const { tokens } = useParams();
+  const importStatus = useSelector((state) => state.tokenImport.importStatus);
```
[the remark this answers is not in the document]
raul-oliveira: Skipping. This screen is reachable only from `ImportTokensScreen.handleContinue` (line 87), which always supplies the `tokens` array. Per project convention we trust internal call sites and validate at boundaries (user input, external APIs).

src/screens/ImportTokensScreen.js
```diff
@@ -0,0 +1,348 @@
...
+  const navigation = useNavigation();
+  const unregisteredTokens = useSelector((state) => state.tokenImport.unregisteredTokens);
+  const loading = useSelector((state) => state.tokenImport.loading);
+  const explorerUrl = useSelector((state) => state.networkSettings.explorerUrl);
+
+  const [tokenList, setTokenList] = useState(
+    () => Object.values(unregisteredTokens).map(
+      (token) => ({ ...token, selected: false })
+    )
+  );
```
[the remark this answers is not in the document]
raul-oliveira: Done — `tokenList` is now derived directly from the `state.tokenImport.unregisteredTokens` selector on every render. Selection state lives in a separate `selectedUids` `Set`, with a dedicated effect that drops uids no longer present in the unregistered map.

src/reducers/reducer.js
```diff
@@ -2370,3 +2398,107 @@ const onTokenIconUpdated = (state, { payload }) => ({
...
+});
+
+const onTokenImportFetchSuccess = (state, { payload }) => ({
+  ...state,
+  tokenImport: {
+    ...state.tokenImport,
+    loading: false,
+    unregisteredTokens: payload,
+  },
+});
```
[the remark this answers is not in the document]
raul-oliveira: Done — `onTokenImportFetchSuccess` now merges instead of replacing:

```js
unregisteredTokens: {
  ...payload,
  ...state.tokenImport.unregisteredTokens,
}
```

Live detections from `TOKEN_IMPORT_NEW_DETECTED` dispatched during the fetch window are preserved (state takes precedence). Same merge-precedence pattern is already used in `onTokenIconsLoaded` (`src/reducers/reducer.js:2387-2392`).

src/screens/ImportTokensScreen.js
```diff
@@ -0,0 +1,392 @@
...
+    try {
+      const supported = await Linking.canOpenURL(url);
+      if (!supported) {
+        log.error(`Cannot open URL: ${url}`);
+        return;
+      }
+      await Linking.openURL(url);
+    } catch (e) {
+      log.error(`Failed to open explorer URL ${url}:`, e);
+    }
```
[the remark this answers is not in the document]
raul-oliveira: Skipping. The default `explorerUrl` values (`src/constants.js:218,236`) include the trailing slash, so the constructed link is well-formed under default configuration. Custom URLs entered via `CustomNetworkSettingsScreen` could omit it, but normalization is better handled at the network-settings boundary if it ever proves necessary.

src/components/ImportTokensModal.js
```diff
@@ -0,0 +1,113 @@
...
+            title={t`Try again`}
+            onPress={onRetry}
+          />
+        </View>
+      </>
+    );
+  };
+
+  // Outside-tap should always dismiss; "Try again" is reachable only via the
+  // explicit button so we never trigger a destructive retry by accident.
```
tuliomir: suggestion(non-blocking): Remove this comment

src/reducers/reducer.js
```diff
@@ -2370,3 +2398,112 @@ const onTokenIconUpdated = (state, { payload }) => ({
...
+
+const removeUidsFromUnregistered = (unregistered, uids) => {
+  const next = { ...unregistered };
+  for (const uid of uids) {
+    delete next[uid];
+  }
+  return next;
+};
```
tuliomir, line 2470: thought: I'm not sure we should have this kind of calculation in the `reducers` file. The business logic of what should be removed should live in the component, not the reducer.

I believe a Redux `slice` would serve you better here, and would also serve as a first step toward modernizing our Redux codebase as a whole.

src/reducers/reducer.js
```diff
@@ -2370,3 +2398,112 @@ const onTokenIconUpdated = (state, { payload }) => ({
...
+      tokenUids,
+    ),
+  },
+});
+
+// Used when a token is registered through an unrelated flow (e.g. the manual
+// "Register token" screen). We want the uid removed from the unregistered list
+// but importStatus must NOT transition to SUCCESS, otherwise the next time the
+// user opens the import modal it would briefly flash "New tokens added!".
+const onTokenImportRemoveFromList = (state, { payload: tokenUids }) => ({
```
tuliomir, line 2488: suggestion: consider event-style actions over setter-style here

Having both `onTokenImportSuccess` and `onTokenImportRemoveFromList` exist because of the comment's caveat (*"importStatus must NOT transition to SUCCESS..."*) is a sign that the dispatcher is being asked to know about the coupling between `importStatus` and `unregisteredTokens`. That's the "actions as setters, not events" smell from the [Redux style guide](https://redux.js.org/style-guide/#model-actions-as-events-not-setters)

A single event-style action like `TOKEN_REGISTERED({ uid, source })` would let the reducer own the rule:

```js
importStatus: source === 'import-modal'
  ? TOKEN_IMPORT_MODAL_STATE.SUCCESS
  : state.tokenImport.importStatus,
```

One action, one reducer path, no comment needed to explain the divergence.

Stepping further back: `importStatus` is ephemeral modal UI state, while `unregisteredTokens` is domain state. They might belong in separate slices entirely — at which point the "don't flash SUCCESS on unrelated flows" problem disappears by construction, since the modal slice would never hear about non-modal registrations.

One more question: would it be better to move the `tokenImport success` entirely to the component state instead of Redux?

src/sagas/tokenImport.js
```diff
@@ -0,0 +1,225 @@
...
+  const tokens = action.payload;
+  const wallet = yield select((state) => state.wallet);
+
+  if (!wallet || !wallet.isReady()) {
+    yield put(tokenImportFailed());
+    return;
+  }
+
+  try {
+    // eslint-disable-next-line no-unreachable
```
tuliomir: question: Why is this unreachable?

src/sagas/wallet.js
```diff
@@ -571,6 +573,23 @@ export function* handleTx(action) {
     return acc;
   }, [{}, new Set([])],);
 
+  // Detect unregistered tokens in this transaction
+  const allTxTokenUids = [...new Set(data.map((io) => io.token).filter(Boolean))];
+  const currentUnregistered = yield select((state) => state.tokenImport.unregisteredTokens);
```
tuliomir, line 655: question(non-blocking): Would it be better to store only the list of registered tokens, and deduce the amount of unregistered from it?

Keeping both lists had led to stale data in the past

src/screens/ImportTokensScreen.js
```diff
@@ -0,0 +1,392 @@
...
+      </Text>
+      <TouchableOpacity onPress={handleRegisterToken}>
+        <Text style={styles.registerLink}>
+          {t`Register a token`}
+        </Text>
+      </TouchableOpacity>
+    </View>
+  );
+
+  const renderContent = () => {
```
tuliomir, line 150: question(non-blocking): Could we reuse more code between the ImportTokensModal and this ImportTokensScreen? I feel they are pretty much duplicated.

src/components/ImportTokensModal.js
raul-oliveira: Done

src/reducers/reducer.js
raul-oliveira: Could you clarify what you mean by "the business logic of what should be removed should live in the component, not the reducer"? I'm reading `removeUidsFromUnregistered` as a pure shape transform — "remove these keys from this dict" — rather than domain logic, since the saga already decides which uids to remove and passes them in the payload. If you mean moving the helper itself out of `reducer.js` into a `utils` module (with the reducer still calling it), I'm happy to do that. If you mean having the component compute the next `unregisteredTokens` and pass the full updated dict in the action payload, I'd push back — that pushes state-shape concerns into a presentation component, which is the opposite of what reducers are for.

On the broader point, agreed — migrating `tokenImport` to a `createSlice` is a sensible first step toward modernizing the Redux layer, and the state is well-scoped and decoupled enough to be a good candidate. I'd prefer to do it in a dedicated follow-up PR rather than expand the scope of this one. Want me to open an issue to track it?

src/reducers/reducer.js
raul-oliveira: I'd lean toward leaving this as-is. The two actions (`TOKEN_IMPORT_SUCCESS` and `TOKEN_IMPORT_REMOVE_FROM_LIST`) are two distinct entry points to the same state effect (drop uids from the unregistered list): one comes from the modal flow and also signals "I'm done successfully", the other comes from any registration that happens through another flow. They're semantically different events at the dispatch site, even if they look setter-shaped in the reducer.

On the single `TOKEN_REGISTERED({ uid, source })` action: it doesn't actually remove the coupling, just relocates it — the reducer would still have to know that `source === 'import-modal'` means flip `importStatus` to SUCCESS. Same conditional, different file.

On moving `importStatus` to component state: I prototyped it, but the cost (new middleware + a hook to listen for dispatched actions) ends up being disproportionate to the gain (mostly removing a comment). I'd rather keep `importStatus` in Redux.

src/sagas/tokenImport.js
raul-oliveira: It was an old comment and it's removed now. Tks

src/sagas/wallet.js
raul-oliveira: The staleness concern is fair, but I don't think we can derive the unregistered set from the registered one alone here. Registered tokens live in `wallet.storage` with a persisted shape (uid/name/symbol). Unregistered tokens, on the other hand, are *discovered* — either by scanning `wallet.getTokens()` against the registered set when the wallet is ready, or by detecting unknown uids in incoming transactions — and since they don't exist as a queryable list anywhere, we have to assemble one ourselves. Beyond the uid, we also cache `name` and `symbol` per unregistered token (fetched via `fetchTokenDetails`) so the import screen can render meaningfully without re-fetching on every pass.

So the second list isn't redundant — it carries data the first one doesn't. The staleness risk is real, but it's constrained to two invalidation paths: `NEW_TOKEN` → `onTokenRegistered` removes the uid from `unregisteredTokens` (covering both the manual register and the import flow), and `TOKEN_METADATA_REMOVED` → re-runs `fetchUnregisteredTokens`, so a previously-registered token reappears in the list if it's still known to the wallet. If you've spotted a path that bypasses either of those, happy to take another look — but as far as I can tell, the invariant holds.

src/screens/ImportTokensScreen.js
raul-oliveira: I did a refactor, extracting the components. See if you like it.

src/components/Icons/OpenInNew.icon.js
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Material Design "open_in_new" icon (weight 300, optical size 24).
+ *
+ * @param {object} props
+ * @property {'default'|'outline'|'fill'} props.type
+ * @property {number} props.size
+ * @property {string} props.color
+ */
+export const OpenInNewIcon = ({ type = 'default', size = DEFAULT_ICON_SIZE, color = COLORS.primary }) => (
```
pedroferreira1, line 22: I think we already have screens (I remember the txdetail modal) that use an icon exactly as this. It's better to have an svg as it's done here, so I believe you could open a refactor PR after this one to make the other screens use this svg and remove the icon that is currently in use. Just for consistency, otherwise we may have same icons but with differences.

src/sagas/wallet.js
```diff
@@ -571,6 +573,23 @@ export function* handleTx(action) {
     return acc;
   }, [{}, new Set([])],);
 
+  // Detect unregistered tokens in this transaction
+  const allTxTokenUids = [...new Set(data.map((io) => io.token).filter(Boolean))];
```
pedroferreira1: I may be wrong but this seems to have a bug. You get the full list of inputs/outputs, so you might get a token that is sent for an address that does not belong to this wallet

src/components/Icons/OpenInNew.icon.js
raul-oliveira: Issue created! https://github.com/HathorNetwork/hathor-wallet-mobile/issues/868

src/sagas/wallet.js
raul-oliveira: Nice catch! It's fixed!

src/sagas/wallet.js
```diff
@@ -571,6 +573,27 @@ export function* handleTx(action) {
     return acc;
   }, [{}, new Set([])],);
 
+  // Detect unregistered tokens in this transaction.
+  // Use wallet.getTxBalance so we only consider tokens whose inputs/outputs touch
+  // an address that belongs to this wallet — a multi-token tx can carry unrelated
+  // tokens on third-party addresses and those must not surface in the import banner.
+  const txBalance = yield call([wallet, wallet.getTxBalance], tx);
+  const myTxTokenUids = Object.keys(txBalance);
+  const currentUnregistered = yield select((state) => state.tokenImport.unregisteredTokens);
```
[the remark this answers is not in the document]
raul-oliveira: Will it happen even with this commit?? https://github.com/HathorNetwork/hathor-wallet-mobile/pull/859/changes/c8cba11534d5213a24d7c0ad35c44aae80396e29

pedroferreira1 (approved):

src/sagas/wallet.js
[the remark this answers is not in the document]
raul-oliveira: Add this comment as a learn. Check the pr logic flow before suggesting fixes. It will improve the quality of the feedbacks

src/sagas/wallet.js
tuliomir: We actually have more than one place containing the "tokens that were discovered by scanning/detecting new txs".

The fact that we do and that a single review pass didn't easily show them tells me that this is a broader issue with token management in the application as a whole, instead of an issue with this specific PR. And that even if we made a good solution locally here, it wouldn't be used as a default on other relevant places as well.

I can look into this as a KTLO in the future, but for now I consider that this point LGTM in this PR. ✅

tuliomir (approved):

tuliomir (approved):

PR #861, opened 2026-04-16
feat: add token icons

obiyankenobi:
### Acceptance Criteria

### Security Checklist

## Summary by CodeRabbit

* **New Features**
  * Token icons and symbol-based avatars now appear in token lists and selectors.
  * Token icon cache persists locally and updates automatically when token metadata arrives.

* **Refactor**
  * Network name removed from the balance view.
  * Token list visuals simplified (arrow removed; streamlined row layout).

* **Documentation**
  * Translation source references and text wrapping adjusted (no visible text changes).


src/components/TokenAvatar.js
```diff
@@ -0,0 +1,85 @@
...
+// Bundled icons shipped with the app. Keyed by mainnet/testnet uid so the same
+// physical asset is reused on both networks.
+const BUNDLED_ICONS = {
+  // Native HTR (same uid on every network).
+  '00': htrIcon,
+  // hUSDC mainnet
+  '00003b17e8d656e4612926d5d2c5a4d5b3e4536e6bebc61c76cb71a65b81986f': husdcIcon,
+  // hUSDC testnet
+  '00000000901832d8f4987030eeed4bb578448e0c05e67dbd121dc917d3989477': husdcIcon,
+};
```
obiyankenobi, line 22: Change this so even hUSDC fetches uid from the json file, or at least the testnet one.

src/components/TokenAvatar.js
```diff
@@ -0,0 +1,85 @@
+/**
```
obiyankenobi, line 1: I prefer larger icons:
Image

Ana prefers smaller ones:
Image

src/components/TokenAvatar.js
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  // Fallback: gray circle with the symbol (first 4 chars, so long symbols still fit).
+  const label = (symbol || '').slice(0, 4);
+  return (
+    <View style={[styles.fallback, dim]}>
+      <Text style={[styles.fallbackText, { fontSize: Math.max(10, size * 0.3) }]}>
+        {label}
+      </Text>
+    </View>
```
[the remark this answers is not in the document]
obiyankenobi: Symbol will never be empty or undefined.

src/components/TokenAvatar.js
```diff
@@ -0,0 +1,80 @@
...
+import htrIcon from '../assets/tokens/htr.png';
+import husdcIcon from '../assets/tokens/husdc.png';
+
+// Bundled icons shipped with the app.
+const BUNDLED_TOKEN_ICONS = {
+  // Native HTR (same uid on every network)
+  '00': htrIcon,
+  // hUSDC mainnet
+  '00003b17e8d656e4612926d5d2c5a4d5b3e4536e6bebc61c76cb71a65b81986f': husdcIcon,
+};
```
tuliomir, line 22: suggestion(non-blocking): I recommend a structure like the token swap allowed tokens JSON ( [link](https://wallet.swap.allowed-tokens.hathor.network/) )

Consider even reusing this existing endpoint instead of creating a hardcoded list here and injecting PNG assets into the source code directly.

```json
{
  "networks": {
    "mainnet": {
      "00003b17e8d656e4612926d5d2c5a4d5b3e4536e6bebc61c76cb71a65b81986f": huscdIcon,
      ...
    },
    "testnet": {
      "cafecafecafe": huscdIcon,
      ...
    }
}
```

src/screens/MainScreen.js
```diff
@@ -561,9 +547,6 @@ class BalanceView extends React.Component {
           {`${lockedStr} ${token.symbol}`}
         </Text>
         <Text style={style.text1}>{t`Locked`}</Text>
-        <View style={style.networkView}>
-          <Text style={style.networkText}>{network}</Text>
-        </View>
```
tuliomir, line 566: chore(non-blocking): Add screenshots of this change , so that we can better understand its before/after

tuliomir (approved):

src/screens/MainScreen.js
obiyankenobi: Sorry, forgot to mention that. It was something I discussed with Ana and we agreed it does not make sense to keep it since we already have the yellow bar at the top. It basically removes the purple network indicator on the token screen. This is the original: 

Screenshot 2026-04-28 at 18 00 05

src/components/TokenAvatar.js
obiyankenobi: We use the explorer service metadata API that already exists for the icons. Only for HTR and mainnet USDC we have this as they are important tokens on the network and many wallets have them. This way, we reduce the amount of requests. I initially thought of doing just for HTR, but it seemed to make sense for USDC as well.

src/actions.js
```diff
@@ -254,8 +254,22 @@ export const types = {
...
+export const tokenIconsLoaded = (urls) => ({
+  type: types.TOKEN_ICONS_LOADED,
+  payload: urls,
+});
+
+export const tokenIconUpdated = (uid, url) => ({
+  type: types.TOKEN_ICON_UPDATED,
+  payload: { uid, url },
+});
+
```
pedroferreira1: Missing docstring.

What's this `urls` parameter? From the param name, it looks like an array of URLs of the icons but seen the reducer you do

```
const onTokenIconsLoaded = (state, { payload }) => ({
  ...state,
  tokenIcons: payload,
});
```

which appears to be an object, so it's important to have this docstring.

src/sagas/tokenIcons.js
```diff
@@ -0,0 +1,75 @@
...
+function* loadCachedIcons() {
+  try {
+    const cached = STORE.getItem(TOKEN_ICONS_CACHE_KEY);
+    if (cached && typeof cached === 'object') {
+      yield put(tokenIconsLoaded(cached));
+    }
+  } catch (e) {
+    log.error('Failed to read token icons cache', e);
+  }
+}
```
pedroferreira1, line 32: Is it possible that this method is called async and the first metadata update was called right before it and we overwrite the redux with the cached value? Maybe we could have in the redux a way to have `{ ...tokenIcons, ...cached }`

src/actions.js
obiyankenobi: Done

src/sagas/tokenIcons.js
obiyankenobi: Done

pedroferreira1 (approved):

PR #870, opened 2026-05-15
feat: fee based token in token swap

raul-oliveira:
### Acceptance Criteria
- Show the HTR network fee on the Token Swap selection and review screens when a fee-based token is involved
- For input fee-based tokens we may have change outputs, for these cases we should charge 0.02 HTR
- For output fee-based tokens 0.01 HTR, since we only have one output
- Disable REVIEW when the user does not have enough HTR to cover the fee

### AsyncStorage BigInt serialization fix (regression from v3 bump)
- **Bug:** `wallet-lib@3.x` exposes `bigint` balance fields on the registered token shape returned by `wallet.storage.getRegisteredTokens()`. The `tokens` saga persists that shape under `NETWORK_TOKENS_KEY` via `STORE.setItem`, which used native `JSON.stringify` — that throws `TypeError: Do not know how to serialize a BigInt`, surfacing as `Error updating network token snapshot:` in the logs and silently breaking the per-network token snapshot.
- **Fix:** `src/store.js` now uses `bigIntUtils.JSONBigInt.stringify` (in `AsyncStorageStore.setItem`) and `bigIntUtils.JSONBigInt.parse` (in the preload loop) — both already exported by the lib for this exact purpose. The token sanitization that was briefly added to `persistTokensForCurrentNetwork` was reverted: the storage layer now round-trips BigInt transparently, so callers persist the full shape.
- **Why it lands here:** the regression only appeared after the v3 bump that this PR brings in, and is reproducible whenever a token registers/unregisters while the wallet is ready. 

### Security Checklist

### Screenshots

https://github.com/user-attachments/assets/78b285e3-de8d-4d66-8b3b-0d3d7b9f8ace

Loading screen before review:
image

Review screen with error in the build tx:
image

Review screen:
image

Review screen with swap between two fee-based tokens:
image

## Summary by CodeRabbit

* **New Features**
  * Network Fee shown in token swap review; token-swap review now builds and displays transaction state (building/error/ready).
  * New insufficient-balance error shown on token swap screen before review.

* **Bug Fixes**
  * BigInt persistence fixed so balances and related data persist correctly.

* **Localization**
  * Updated translation entries and source references for token swap, token registration, QR flows, and send confirmations (multiple locales and template).

[Review Change Stack](https://app.coderabbit.ai/change-stack/HathorNetwork/hathor-wallet-mobile/pull/870?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/screens/TokenSwap.js
+++ b/src/screens/TokenSwap.js
@@ -20,15 +20,12 @@ import { useDispatch, useSelector } from 'react-redux';
 import { getStatusBarHeight } from 'react-native-status-bar-height';
 import { t } from 'ttag';
 import { get } from 'lodash';
-import hathorLib from '@hathor/wallet-lib';
 import { SwapIcon } from '../components/Icons/Swap.icon';
 import AmountInputAccessory from '../components/AmountInputAccessory';
 
 import { renderValue, formatAmountToInput } from '../utils';
 import {
   calcAmountWithSlippage,
-  calculateSwapNetworkFee,
-  isSwappingFeeBasedTokens,
   renderAmountAndSymbolWithSlippage,
   renderConversionRate,
   selectTokenSwapAllowedTokens,
@@ -91,9 +88,6 @@ const TokenSwap = () => {
   // below for the rationale.
   const [keyboardHeight, setKeyboardHeight] = useState(0);
 
-  // we use null value in this prop to track loading state
-  const [networkFee, setNetworkFee] = useState(0n);
-
   // Ref to track editing timeout so we can cancel it on new focus
   const editingTimeoutRef = useRef(null);
 
@@ -103,7 +97,6 @@ const TokenSwap = () => {
     outputToken,
     swapPathQuote: quote,
   } = useSelector((state) => state.tokenSwap);
-  const wallet = useSelector((state) => state.wallet);
   const { decimalPlaces } = useSelector((state) => ({
     decimalPlaces: state.serverInfo?.decimal_places
   }));
@@ -180,40 +173,6 @@ const TokenSwap = () => {
       setOutputTokenAmount(quote.amount_out);
     }
   }, [quote]);
-
-  useEffect(() => {
-    let cancelled = false;
-    refreshNetworkFee(() => cancelled);
-    return () => { cancelled = true; };
-  }, [quote, inputToken, outputToken, wallet, tokensBalance]);
-
-  const refreshNetworkFee = async (isCancelled) => {
-    if (!quote || !inputToken || !outputToken || !wallet
-        || !quote.path || quote.path.length === 0
-        || !isSwappingFeeBasedTokens(inputToken, outputToken)) {
-      setNetworkFee(0n);
-      return;
-    }
-    if (getAvailableAmount(inputToken, tokensBalance) < BigInt(quote.amount_in)) {
-      setNetworkFee(null);
-      return;
-    }
-
-    setNetworkFee(null);
-
-    try {
-      const fee = await calculateSwapNetworkFee(wallet, quote, inputToken, outputToken);
-      if (isCancelled()) return;
-      setNetworkFee(fee);
-    } catch (err) {
-      if (isCancelled()) return;
-      console.error('Failed to calculate network fee', err);
-      // Fall back to a single output's worth so the button doesn't stay
-      // disabled forever on a transient selector error — same posture as
-      // SendAmountInput.
-      setNetworkFee(hathorLib.constants.FEE_PER_OUTPUT);
-    }
-  };
 
   function onInputAmountChange(text, value) {
     setInputTokenAmountStr(text);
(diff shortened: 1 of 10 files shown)


src/screens/TokenSwap.js
```diff
@@ -292,6 +333,20 @@ const TokenSwap = () => {
     // Quote will be fetched when keyboard dismisses via onEndEditing handlers
   };
 
+  // True when the user has enough HTR to cover the network fee on top of any HTR
+  // they are also sending as the swap input.
+  const hasEnoughHTRForFee = () => {
+    if (!networkFee || networkFee === 0n) return true;
+    const htrBalance = get(tokensBalance, hathorLib.constants.NATIVE_TOKEN_UID);
+    if (!htrBalance) return false;
```
tuliomir: question(non-blocking): This is more a question about the protocol and TokenSwap NC Code than the mobile interface itself.

Suppose I have 1 KOIN and 0 HTR in my wallet balance. I want to sell 0.1 KOIN for 100 HTR. In the current ruleset, the transaction would be denied entirely because I don't have any HTR balance beforehand.

However, the contract could pay the fees for me, and we already have plenty of tests for those situations. In that case I would receive `100HTR - fees` as my payment, and would conclude the transaction successfully.

Should we consider this possibility, or is is outside the scope of this PR?

tuliomir (approved):

src/screens/TokenSwapReview.js
```diff
@@ -245,6 +296,14 @@ const TokenSwapReview = () => {
                     <Text style={styles.quoteValue}>{renderAmountAndSymbolWithSlippage('output', quote.amount_in, tokenIn, TOKEN_SWAP_SLIPPAGE)}</Text>
                   </View>
                 )}
+                <View style={styles.quoteRow}>
+                  <Text style={styles.quoteHeader}>{t`Network Fee`}</Text>
+                  <Text style={styles.quoteValue}>
+                    {networkFee != null && networkFee > 0n
+                      ? `${renderValue(networkFee, false)} ${hathorLib.constants.DEFAULT_NATIVE_TOKEN_CONFIG.symbol}`
+                      : t`No fee`}
```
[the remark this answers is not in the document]
raul-oliveira: We don't use in any place the serveInfo.native_token.symbol. Maybe in a future refactor.

src/screens/TokenSwapReview.js
```diff
@@ -100,57 +165,34 @@ const TokenSwapReview = () => {
...
+  // Closes the send-progress modal and navigates back. The beforeRemove
+  // listener releases reserved UTXOs along the way.
   const exitOnError = () => {
     setModal(null);
-    dispatch(tokenSwapFetchSwapQuote(
-      quote.direction,
-      quote.direction === 'input' ? quote.amount_in : quote.amount_out,
-      tokenIn.uid,
-      tokenOut.uid
-    ));
```
pedroferreira1, line 113: Why did you remove this?

src/screens/TokenSwapReview.js
```diff
@@ -51,17 +59,73 @@ const TokenSwapReview = () => {
...
+        const address = await wallet.getAddressAtIndex(0);
+        const [method, data] = buildTokenSwap(
+          contractId,
+          address,
+          quote,
+          tokenIn.uid,
+          tokenOut.uid,
+          TOKEN_SWAP_SLIPPAGE,
+        );
+        const tx = await wallet.createNanoContractTransaction(
```
pedroferreira1: This is not a `tx`, this is a `sendTransaction` object

src/screens/TokenSwapReview.js
```diff
@@ -100,57 +165,34 @@ const TokenSwapReview = () => {
...
-        data,
-        { pinCode: pin },
-      );
-      const promise = sendTransaction.runFromMining();
-
-      setLoading(false);
-
-      // show loading modal
+      await wallet.signTx(sendTx.transaction, { pinCode: pin });
+      const promise = sendTx.runFromMining();
```
pedroferreira1, line 203: This works but the flow we've been using in this case is to get the send transaction object and call the methods from the class

```
await sendTx.signTx(pin);
const promise = sendTx.runFromMining();
```

src/screens/TokenSwap.js
raul-oliveira: As we discussed with @pedroferreira1, this should be aligned with product in a future pr for us to address it, or not. 

In any case, it's a valid point

src/screens/TokenSwapReview.js
raul-oliveira: Restored. I deleted it by mistake

src/screens/TokenSwapReview.js
raul-oliveira: Renamed to `sendTransaction`.

src/screens/TokenSwapReview.js
raul-oliveira: I tried. It isn't working as expected. To track this I opened an issue in the wallet-lib project. https://github.com/HathorNetwork/hathor-wallet-lib/issues/1100

src/screens/TokenSwap.js
```diff
@@ -451,6 +451,12 @@ const TokenSwap = () => {
             </View>
 
             <View style={styles.buttonContainer}>
+              {inputToken && inputTokenAmount > 0n
+                && getAvailableAmount(inputToken, tokensBalance) < inputTokenAmount && (
+                <Text style={styles.error}>
+                  {t`Insufficient balance of ${inputToken.symbol}.`}
```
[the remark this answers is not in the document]
raul-oliveira: we are allowing the user to go to the review page to avoid guessing the fee in our side. it belongs to the wallet-lib responsability in this case.

pedroferreira1 (approved):

tuliomir (approved):

PR #876, opened 2026-05-27
fix: rc errors on importing and sending fbt

raul-oliveira:
### Acceptance Criteria
- Fix the import tokens version (merge issue)
- Should be able to build and send the transaction in the send tokens flow (HTR, Deposit and fee based tokens)
- Should show a loading state before the review screen shows
- The review screen should use the built tx when showing data

### Security Checklist

## Release Notes

* **New Features**
  * Transaction confirmation screen now displays real-time loading indicators with enhanced user feedback during transaction building and preparation.

* **Bug Fixes**
  * Improved error handling for transaction build failures, providing specific and actionable error messages for fee-related and balance shortage issues.

* **Localization**
  * Updated and refreshed translation strings for send operations, confirmations, and token swap workflows across supported languages.

[Review Change Stack](https://app.coderabbit.ai/change-stack/HathorNetwork/hathor-wallet-mobile/pull/876?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/sagas/tokenImport.js
+++ b/src/sagas/tokenImport.js
@@ -52,6 +52,7 @@ function* fetchTokenDetails(wallet, uid) {
         uid,
         name: details.tokenInfo.name,
         symbol: details.tokenInfo.symbol,
+        version: details.tokenInfo.version,
       };
     }
   } catch (e) {
@@ -101,6 +102,7 @@ export function* fetchUnregisteredTokens() {
         uid: details.uid,
         name: details.name,
         symbol: details.symbol,
+        version: details.version,
       };
       yield put(tokenFetchBalanceRequested(uid, true));
     }
@@ -135,6 +137,7 @@ export function* handleNewTokenDetection(action) {
       uid: details.uid,
       name: details.name,
       symbol: details.symbol,
+      version: details.version,
     };
     yield put(tokenFetchBalanceRequested(uid, true));
   }
@@ -168,11 +171,17 @@ export function* importSelectedTokens(action) {
           continue;
         }
 
+        const persistedPayload = {
+          uid: token.uid,
+          name: token.name,
+          symbol: token.symbol,
+          version: token.version,
+        };
         yield call(
           [wallet.storage, wallet.storage.registerToken],
-          { uid: token.uid, name: token.name, symbol: token.symbol },
+          persistedPayload,
         );
-        yield put(newToken({ uid: token.uid, name: token.name, symbol: token.symbol }));
+        yield put(newToken(persistedPayload));
         yield put(tokenFetchBalanceRequested(token.uid, true));
 
         alreadyRegistered.add(token.uid);
--- a/src/screens/SendConfirmScreen.js
+++ b/src/screens/SendConfirmScreen.js
@@ -70,6 +70,7 @@ const SendConfirmScreen = () => {
   const navigation = useNavigation();
   const params = useParams();
 
+  // Parse and store navigation params
   const { amount, address, token } = params;
   const isNFT = isTokenNFT(token.uid, tokenMetadata);
   const amountAndToken = `${renderValue(amount, isNFT)} ${token.symbol}`;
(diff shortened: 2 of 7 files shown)


src/sagas/tokenImport.js
```diff
@@ -52,6 +52,7 @@ function* fetchTokenDetails(wallet, uid) {
         uid,
         name: details.tokenInfo.name,
         symbol: details.tokenInfo.symbol,
+        version: details.tokenInfo.version,
```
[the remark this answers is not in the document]
raul-oliveira: it's not needed. everytime the wallet starts we fetch the tokens and the current api is already returning the version field.

src/screens/SendConfirmScreen.js
```diff
@@ -46,56 +70,91 @@ const SendConfirmScreen = () => {
...
+      if (sentSuccessfullyRef.current || releasedRef.current || !sendTx) {
+        return;
+      }
+      e.preventDefault();
+      releasedRef.current = true;
+      try { await sendTx.releaseUtxos(); } catch (err) { console.error(err); }
+      navigation.dispatch(e.data.action);
     });
+    return unsubscribe;
+  }, [sendTx, navigation]);
```
pedroferreira1: Thinking about this release here. After the user types the pin and the tx is sending, is not possible for the user to go back. But if the user taps the physical android back button, maybe this would release the utxos AND send the tx.

I don't think it's critical to change anything in this PR and @tuliomir has a task to review this physical back button handling, so maybe we should just mention this there.

pedroferreira1 (approved):

src/screens/SendConfirmScreen.js
tuliomir: Referencing the #791 here for better visibility

src/screens/SendConfirmScreen.js
```diff
@@ -46,56 +70,91 @@ const SendConfirmScreen = () => {
...
+      console.error(err);
+      setBuildError({ message: mapBuildError(err) });
+      setPhase(PHASE.ERROR);
+    }
   };
 
-  /**
-   * Executed when user clicks to send the tx and opens PIN screen
-   */
   const onSendPress = () => {
```
tuliomir, line 160: issue: We should not remove docstrings from methods

src/screens/SendConfirmScreen.js
```diff
@@ -119,7 +182,7 @@ const SendConfirmScreen = () => {
     let initialRoute = 'CameraPermissionScreen';
     if (isCameraAvailable) {
       initialRoute = 'SendScanQRCode';
-    } else if (isCameraAvailable === false) { // might be null
+    } else if (isCameraAvailable === false) {
```
tuliomir: chore: This comment, too, serves a purpose and should not be removed right now. Maybe after we add the unit/component/saga tests we can remove those, as the functionality will be safer from regressions already.

src/screens/SendConfirmScreen.js
```diff
@@ -46,56 +70,91 @@ const SendConfirmScreen = () => {
...
+    return () => { cancelled = true; };
+  }, []);
+
+  // Release reserved UTXOs if the user navigates away before sending.
+  useEffect(() => {
+    const unsubscribe = navigation.addListener('beforeRemove', async (e) => {
+      if (sentSuccessfullyRef.current || releasedRef.current || !sendTx) {
+        return;
+      }
+      e.preventDefault();
```
tuliomir: question(non-blocking): Shouldn't we prevent default also when the conditions above early return?

Another question: Shouldn't we try to release utxos every time the user navigates away? If I recall correctly, `sendTx.releaseUtxos()` is a no-op if the utxo was already spent by the tx, so we can simplify our code here by trusting it.

tuliomir (approved):

src/screens/SendConfirmScreen.js
raul-oliveira: Good point. I've refactored it to be simpler. tks

src/screens/SendConfirmScreen.js
raul-oliveira: Solved. comments restored

src/screens/SendConfirmScreen.js
raul-oliveira: comment restored

pedroferreira1 (approved):

tuliomir (approved):

PR #885, opened 2026-06-03
fix: persist registered nano contracts per network

raul-oliveira:
### Acceptance Criteria
- Persist registered nano contracts per network so they survive switching networks (testnet ↔ mainnet, and testnet → a custom network without wallet-service)
- Nano contract registration now behaves like token registration: a contract registered on a network reappears when returning to that network, and does not show up on other networks

### Security Checklist

## Summary by CodeRabbit

* **New Features**
  * Nano contracts are now automatically saved and restored on a per-network basis, ensuring your contracts persist when switching between different networks.

* **Tests**
  * Added comprehensive test suite for nano contract persistence and restoration workflows.

### DIFF
--- a/src/sagas/nanoContract.js
+++ b/src/sagas/nanoContract.js
@@ -21,7 +21,9 @@ import {
 } from 'redux-saga/effects';
 import { t } from 'ttag';
 import { NanoRequest404Error } from '@hathor/wallet-lib/lib/errors';
-import { getRegisteredNanoContracts, safeEffect } from './helpers';
+import { getGenesisHash, getRegisteredNanoContracts, safeEffect } from './helpers';
+import { WALLET_STATUS } from './wallet';
+import { STORE } from '../store';
 import {
   nanoContractBlueprintInfoFailure,
   nanoContractBlueprintInfoSuccess,
@@ -31,6 +33,7 @@ import {
   nanoContractHistorySuccess,
   nanoContractRegisterFailure,
   nanoContractRegisterSuccess,
+  nanoContractsSavedForNetwork,
   nanoContractUnregisterSuccess,
   onExceptionCaptured,
   types,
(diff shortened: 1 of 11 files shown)


src/sagas/nanoContract.js
```diff
@@ -135,6 +138,106 @@ export function* init() {
...
+          [wallet.storage, wallet.storage.registerNanoContract],
+          contract.ncId,
+          contract,
+        );
+        yield put(nanoContractRegisterSuccess({
+          entryKey: contract.ncId,
+          entryValue: contract,
+        }));
+      }
+    }
```
pedroferreira1: Maybe we could have a try...catch for each contract handling. This way if an error occurs for a contract only, the others will be normally restored. What do you think?

src/sagas/nanoContract.js
```diff
@@ -135,6 +138,106 @@ export function* init() {
...
+
+    const serverInfo = yield select((state) => state.serverInfo);
+    const genesisHash = getGenesisHash(serverInfo);
+    if (!genesisHash) return;
+
+    const savedContracts = yield call([STORE, STORE.getNanoContractsForNetwork], genesisHash);
+    if (!savedContracts) return;
+
+    const wallet = yield select((state) => state.wallet);
+    if (!wallet) return;
```
pedroferreira1: Should we check if `wallet.isReady()`?

pedroferreira1 (approved):

src/sagas/nanoContract.js
raul-oliveira: Done in 43b1fe2 — each contract is now restored inside its own try/catch, so a failure on one is logged with its `ncId` and the remaining contracts still restore.

src/sagas/nanoContract.js
raul-oliveira: Done in 43b1fe2 — the guard is now `if (!wallet || !wallet.isReady()) return;`, consistent with `updateNetworkNanoContractSnapshot`. In practice the saga runs on `START_WALLET_SUCCESS`, which is only dispatched after the wallet reaches `WALLET_STATE_READY`, but the explicit check keeps it defensive and consistent.

src/sagas/networkSettings.js
```diff
@@ -275,12 +276,22 @@ export function* persistNetworkSettings(action) {
...
   yield put(saveTokensForNetwork());
-  yield race({
-    saved: take(types.TOKENS_SAVED_FOR_NETWORK),
-    timeout: delay(5000),
-  });
+  yield put(saveNanoContractsForNetwork());
+  yield all([
+    race({
+      saved: take(types.TOKENS_SAVED_FOR_NETWORK),
+      timeout: delay(5000),
```
tuliomir, line 288: suggestion(non-blocking): Use a constant instead of a hardcoded value here.

tuliomir (approved):

PR #891, opened 2026-07-06
feat(decimal): add wallet-wide amount format preference

raul-oliveira:
### Acceptance Criteria
- New "Amount format" screen under Settings → Advanced Settings
- Users can choose **Expanded** (default) or **Compressed** amount display
- Compressed shows small values in subscript-zero notation (e.g. `0.0000005195` → `0.0₆5195`)
- The preference is wallet-wide and independent of the selected network
- The choice persists across app restarts and resets to Expanded on wallet reset
- Applies to the primary display sites: balances, token list, and transaction history/details
- Editable amount inputs are unaffected (always expanded)

Design: [Figma – Amounts Update](https://www.figma.com/design/2xRserWXaSJAgkbiQlPzhr/Amounts-Update?node-id=103-2312)

### Screenshots / videos

https://github.com/user-attachments/assets/29d06887-9de0-4e83-abb6-571b19fa86ab

### Security Checklist

## Summary by CodeRabbit

* **New Features**
  * Added an **Amount Format** screen with **Expanded** and **Compressed** modes, including a live preview.
  * Added an **“Amount format”** option in Advanced Settings (feature-toggle controlled) and persisted the choice.
  * Applied the selected format across balances and transaction/token displays, respecting configured decimal precision.
  * Updated translations for the new setting and related text.

* **Bug Fixes**
  * Restores the selected amount format after reopening/unlocking the wallet.

### DIFF
--- a/__tests__/utils/amountFormat.test.js
+++ b/__tests__/utils/amountFormat.test.js
@@ -65,34 +65,50 @@ describe('capDecimalsByMagnitude', () => {
 });
 
 describe('renderValue with amount format', () => {
+  it('renders with the network decimal places it is given', () => {
+    // 12 base units at 8 decimals => 0.00000012
+    expect(renderValue(12n, false, 8)).toBe('0.00000012');
+  });
+
+  it('falls back to the lib default decimals when unset', () => {
+    // 1234 base units at default 2 decimals => 12.34
+    expect(renderValue(1234n, false)).toBe('12.34');
+  });
+
   it('defaults to expanded (unchanged behavior)', () => {
-    expect(renderValue(1234n, false)).toBe(renderValue(1234n, false, AMOUNT_FORMAT.EXPANDED));
+    expect(renderValue(1234n, false, 8)).toBe(renderValue(1234n, false, 8, AMOUNT_FORMAT.EXPANDED));
   });
 
   it('routes COMPRESSED output through compressAmountString', () => {
     const amount = 500000n;
-    expect(renderValue(amount, false, AMOUNT_FORMAT.COMPRESSED))
-      .toBe(compressAmountString(renderValue(amount, false, AMOUNT_FORMAT.EXPANDED)));
+    expect(renderValue(amount, false, 8, AMOUNT_FORMAT.COMPRESSED))
+      .toBe(compressAmountString(renderValue(amount, false, 8, AMOUNT_FORMAT.EXPANDED)));
+  });
+
+  it('compresses sub-1 amounts once the network precision allows it', () => {
+    // 5195 base units at 8 decimals => 0.00005195 => 0.0₄5195. At the lib
+    // default of 2 decimals there is no zero run long enough to compress.
+    expect(renderValue(5195n, false, 8, AMOUNT_FORMAT.COMPRESSED)).toBe('0.0₄5195');
   });
 
   it('is a no-op for NFT amounts (no fractional part)', () => {
-    expect(renderValue(12345n, true, AMOUNT_FORMAT.COMPRESSED)).toBe('12,345');
+    expect(renderValue(12345n, true, 8, AMOUNT_FORMAT.COMPRESSED)).toBe('12,345');
   });
 });
 
 describe('renderHomeValue (magnitude rule + network decimals)', () => {
   it('renders small amounts with up to 8 decimals from the network decimal places', () => {
     // 12 base units at 8 decimals => 0.00000012 (< 1 => 8 dp)
-    expect(renderHomeValue(12n, false, AMOUNT_FORMAT.EXPANDED, 8)).toBe('0.00000012');
+    expect(renderHomeValue(12n, false, 8, AMOUNT_FORMAT.EXPANDED)).toBe('0.00000012');
   });
 
   it('caps large amounts to 2 decimals by magnitude', () => {
     // 10000012345678 at 8 decimals => 100,000.12345678 (>= 10000 => 2 dp)
-    expect(renderHomeValue(10000012345678n, false, AMOUNT_FORMAT.EXPANDED, 8)).toBe('100,000.12');
+    expect(renderHomeValue(10000012345678n, false, 8, AMOUNT_FORMAT.EXPANDED)).toBe('100,000.12');
   });
 
   it('applies Compressed notation after the magnitude cap', () => {
-    expect(renderHomeValue(12n, false, AMOUNT_FORMAT.COMPRESSED, 8)).toBe('0.0₆12');
+    expect(renderHomeValue(12n, false, 8, AMOUNT_FORMAT.COMPRESSED)).toBe('0.0₆12');
   });
 
   it('falls back to the lib default decimals when unset', () => {
--- a/__tests__/utils/amountFormat.test.js
+++ b/__tests__/utils/amountFormat.test.js
@@ -1,0 +1,133 @@
+import { describe, it, expect } from '@jest/globals';
+import {
+  compressAmountString, renderValue, getDisplayAmountFormat, normalizeAmountFormat,
+  capDecimalsByMagnitude, renderHomeValue,
+} from '../../src/utils';
+import { AMOUNT_FORMAT, AMOUNT_FORMAT_DEFAULT, AMOUNT_FORMAT_FEATURE_TOGGLE } from '../../src/constants';
+
+const stateWith = (flagOn, amountFormat) => ({
+  featureToggles: { [AMOUNT_FORMAT_FEATURE_TOGGLE]: flagOn },
+  amountFormat,
+});
+
+describe('compressAmountString', () => {
+  // Compresses only the leading fractional zero run into subscript notation and
+  // keeps every later digit verbatim (including inner and trailing zeros), so the
+  // compressed form shows the same significant digits as the Expanded form.
+  it.each([
+    ['0.0000005195', '0.0₆5195'],
+    ['0.000000012345', '0.0₇12345'],
+    ['0.00012', '0.0₃12'], // exactly 3 leading zeros (threshold)
+    ['-0.0000005195', '-0.0₆5195'], // keeps the negative sign
+    ['-0.00012', '-0.0₃12'],
+    ['0.0000005000', '0.0₆5000'], // trailing zeros are kept, not trimmed
+    ['0.00000000000050', '0.0₁₂50'], // deep run + trailing zero kept verbatim
+    ['0.000000000012', '0.0₁₀12'], // multi-digit subscript (10 zeros)
+    ['0.000000045600000123', '0.0₇45600000123'], // inner zeros kept verbatim
+  ])('compresses %s -> %s', (input, expected) => {
+    expect(compressAmountString(input)).toBe(expected);
+  });
+
+  it.each([
+    ['0.05'], // 1 leading zero, below threshold
+    ['0.0012'], // 2 leading zeros, below threshold
+    ['0.005195'], // 2 leading zeros, below threshold
+    ['1.0000005195'], // non-zero integer part
+    ['10.0000005'], // non-zero integer part
+    ['1,234.00005'], // non-zero (grouped) integer part
+    ['0.000000'], // effectively zero, nothing significant after the run
+    ['1,234'], // no fractional part
+    ['1234'], // no fractional part
+  ])('leaves %s unchanged', (input) => {
+    expect(compressAmountString(input)).toBe(input);
+  });
+});
+
+describe('capDecimalsByMagnitude', () => {
+  // < 1 -> 8, 1-999 -> 4, 1000-9999 -> 3, >= 10000 -> 2; truncated, not padded.
+  it.each([
+    ['0.12345678', '0.12345678'], // < 1, keeps up to 8
+    ['0.123456789', '0.12345678'], // < 1, truncated to 8
+    ['0.5', '0.5'], // < 1, no padding
+    ['1.12345678', '1.1234'], // 1-999 -> 4
+    ['999.98765', '999.9876'], // 999 -> 4
+    ['1000.12345', '1000.123'], // 1000-9999 -> 3
+    ['9999.9999', '9999.999'], // 9999 -> 3
+    ['10000.98765', '10000.98'], // >= 10000 -> 2
+    ['100000.12345678', '100000.12'],
+    ['9223372036854775808.12345678', '9223372036854775808.12'],
+    ['-0.123456789', '-0.12345678'], // sign preserved
+    ['1,234.56789', '1,234.567'], // grouping preserved, magnitude 1234 -> 3
+    ['1234', '1234'], // no fractional part unchanged
+  ])('%s -> %s', (input, expected) => {
+    expect(capDecimalsByMagnitude(input)).toBe(expected);
+  });
+});
+
+describe('renderValue with amount format', () => {
+  it('defaults to expanded (unchanged behavior)', () => {
+    expect(renderValue(1234n, false)).toBe(renderValue(1234n, false, AMOUNT_FORMAT.EXPANDED));
+  });
+
+  it('routes COMPRESSED output through compressAmountString', () => {
+    const amount = 500000n;
+    expect(renderValue(amount, false, AMOUNT_FORMAT.COMPRESSED))
+      .toBe(compressAmountString(renderValue(amount, false, AMOUNT_FORMAT.EXPANDED)));
+  });
+
+  it('is a no-op for NFT amounts (no fractional part)', () => {
+    expect(renderValue(12345n, true, AMOUNT_FORMAT.COMPRESSED)).toBe('12,345');
+  });
+});
+
+describe('renderHomeValue (magnitude rule + network decimals)', () => {
+  it('renders small amounts with up to 8 decimals from the network decimal places', () => {
+    // 12 base units at 8 decimals => 0.00000012 (< 1 => 8 dp)
+    expect(renderHomeValue(12n, false, AMOUNT_FORMAT.EXPANDED, 8)).toBe('0.00000012');
+  });
+
+  it('caps large amounts to 2 decimals by magnitude', () => {
+    // 10000012345678 at 8 decimals => 100,000.12345678 (>= 10000 => 2 dp)
+    expect(renderHomeValue(10000012345678n, false, AMOUNT_FORMAT.EXPANDED, 8)).toBe('100,000.12');
+  });
+
+  it('applies Compressed notation after the magnitude cap', () => {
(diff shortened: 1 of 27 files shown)


src/screens/AmountFormat.js
```diff
@@ -0,0 +1,281 @@
...
+import {
+  AMOUNT_FORMAT,
+  AMOUNT_FORMAT_DEFAULT,
+  AMOUNT_FORMAT_KEY,
+} from '../constants';
+import { setAmountFormat } from '../actions';
+import { compressAmountString } from '../utils';
+
+// decimal_places is 2 on the network, so this high-precision sample can't come
+// from a real BigInt amount. Formatting the literal through the shared helper
```
raul-oliveira: remove this comment. it's wrong.

src/utils.js
```diff
@@ -372,19 +372,106 @@ export const getLightBackground = (alpha) => {
...
+ */
+const toSubscript = (n) => String(n)
+  .split('')
+  .map((digit) => SUBSCRIPT_DIGITS[Number(digit)])
+  .join('');
+
 /**
- * Render value in a formatted way for display
+ * Collapse fractional zero runs (leading and inner) into subscript notation,
+ * e.g. "0.0000005195" -> "0.0₆5195", "0.000000045600000123" -> "0.0₇4560₅123".
```
raul-oliveira: we won't take the inner anymore. get only the first zeros after couting 3 (< 2 zeros)

src/screens/AmountFormat.js
```diff
@@ -0,0 +1,278 @@
...
+            t`Compresses leading zeros into a subscript count for shorter, easier-to-read small values (ex: 0.0₆5195).`,
+            false,
+          )}
+        </View>
+
+        <Text style={styles.previewHeader}>{t`PREVIEW`}</Text>
+        <View style={styles.previewPanel}>
+          <Text style={styles.previewLabel}>{t`Full amount`}</Text>
+          <Text style={styles.previewValue}>{`${previewValue} HTR`}</Text>
+        </View>
```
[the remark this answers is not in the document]
raul-oliveira: done

src/utils.js
```diff
@@ -372,19 +372,90 @@ export const getLightBackground = (alpha) => {
...
+ * decimal point) is compressed; inner zeros are kept verbatim
+ * (e.g. "0.000000045600000123" -> "0.0₇45600000123"). Takes prettyValue's
+ * stringified amount to reuse the lib's BigInt formatting. Returns the input
+ * unchanged when nothing qualifies: value >= 1, integer, zero, or a leading run
+ * shorter than COMPRESS_MIN_ZERO_RUN.
+ *
+ * @param {string} formatted Amount string from prettyValue
+ * @return {string}
+ */
+export const compressAmountString = (formatted) => {
```
tuliomir, line 402: issue: Missing test coverage.

This is the headline logic and it's unpinned; a regression in the zero-counter or the `0.0${subscript}` reconstruction ships green.

Suggested tests:

```js
describe('compressAmountString', () => {
  it.each([
    ['0.0000005195', '0.0₆5195'],
    ['0.000000045600000123', '0.0₇45600000123'],
    ['-0.0000005195', '-0.0₆5195'],
    ['0.005195', '0.005195'],
    ['1.0000005195', '1.0000005195'],
    ['0.00000000000050', '0.0₁₀5'],
    ['1234', '1234'],
  ])('%s -> %s', (input, expected) => expect(compressAmountString(input)).toBe(expected));
});
```

src/screens/Settings.js
```diff
@@ -173,6 +175,13 @@ export class Settings extends React.Component {
                   onPress={() => this.props.navigation.navigate('AddressMode')}
                 />
               )}
+            {this.props.amountFormatEnabled
```
tuliomir, line 178: issue: Here we gate the *menu entry* on `AMOUNT_FORMAT_FEATURE_TOGGLE`, but the saga hydrates `amountFormat` from storage regardless (`wallet.js:203-204`).

`renderValue` applies compression purely off redux state. So: user selects Compressed while the toggle is on → toggle is rolled back remotely → amounts stay compressed everywhere but the Settings entry is gone;

No way back short of a wallet reset. Is the toggle meant to gate only the entry point, or the whole feature? If the latter, gate rendering on the toggle (or skip hydration when off).

src/screens/AmountFormat.js
```diff
@@ -0,0 +1,282 @@
...
+import {
+  AMOUNT_FORMAT,
+  AMOUNT_FORMAT_DEFAULT,
+  AMOUNT_FORMAT_KEY,
+  DEFAULT_TOKEN,
+} from '../constants';
+import { setAmountFormat } from '../actions';
+import { compressAmountString } from '../utils';
+
+const PREVIEW_SAMPLE = '0.0000005195';
```
tuliomir, line 29: chore(low priority): Hardcoded value may differ from user exhibition

The preview calls `compressAmountString(PREVIEW_SAMPLE)` on the hardcoded `'0.0000005195'` (`AmountFormat.js:29,56`), so it always looks compressed. Real amounts go through `renderValue → hathorLib.numberUtils.prettyValue(amount)` (`utils.js:449`).

If `prettyValue` rounds to the token's decimal places (HTR = 2), sub-1 values never reach a ≥3 zero run and Compressed is a silent no-op for those tokens. Confirm `prettyValue` returns the full-precision decimal string here — otherwise the preview proves the feature works without exercising the
formatter production uses.

src/utils.js
```diff
@@ -372,19 +372,90 @@ export const getLightBackground = (alpha) => {
...
+  const integerPart = unsigned.slice(0, dotIndex);
+  const decimalPart = unsigned.slice(dotIndex + 1);
+
+  // Only sub-1 values have a leading-zero run worth compressing.
+  if (integerPart.replace(/,/g, '') !== '0') {
+    return formatted;
+  }
+
+  // Trailing zeros are insignificant; drop them before scanning.
+  const trimmed = decimalPart.replace(/0+$/, '');
```
tuliomir: question: Discrepancy in exhibition

`0.0000005000 → 0.0₆5` under Compressed, while the Expanded path keeps whatever `prettyValue` returns. The two modes can show different significant digits for the same amount. Intended?

src/utils.js
```diff
@@ -372,19 +372,90 @@ export const getLightBackground = (alpha) => {
...
  *
  * @param {bigint} amount The token amount as BigInt
  * @param {boolean} isInteger Whether the token is an NFT or regular token
+ * @param {string} [amountFormat] AMOUNT_FORMAT.EXPANDED (default) or COMPRESSED
  * @return {string} Formatted value for display
  */
-export const renderValue = (amount, isInteger) => {
-  if (isInteger) {
-    return hathorLib.numberUtils.prettyValue(amount, 0);
+export const renderValue = (amount, isInteger, amountFormat = AMOUNT_FORMAT.EXPANDED) => {
```
tuliomir: suggestion(non-blocking): Naming convention

Param named `isInteger`; every call site passes `isNFT` / `isNFT()`. Pre-existing, but you're
touching the signature — align the name to the concept.

src/screens/MainScreen.js
```diff
@@ -54,6 +54,7 @@ const mapStateToProps = (state) => ({
   isOnline: state.isOnline,
   wallet: state.wallet,
   tokenMetadata: state.tokenMetadata,
+  amountFormat: state.amountFormat,
```
tuliomir: suggestion(non-blocking): Selector over parameter 

Any future display site that forgets the prop silently falls back to Expanded (`renderValue`'s default). Would reading `amountFormat` from a selector inside the render helpers be more robust against drift?

src/utils.js
raul-oliveira: Added `__tests__/utils/amountFormat.test.js` pinning `compressAmountString` with an `it.each` table over these cases (plus `renderValue` routing, `getDisplayAmountFormat`, `normalizeAmountFormat`, and `capDecimalsByMagnitude`/`renderHomeValue`). Two expected values shift with the T4 change (leading-run only, no trailing trim), and one count fix — `0.00000000000050` has 12 leading zeros → `0.0₁₂50`.

src/screens/Settings.js
raul-oliveira: Gating the whole feature. Added a flag-gated selector `getDisplayAmountFormat(state)` (= stored preference while the flag is on, else Expanded) and pointed the display sites at it (MainScreen, Dashboard token list; TxDetailsModal inherits). Flag off ⇒ Expanded everywhere regardless of the stored value, so a rollback can't strand the wallet in Compressed. Gating at render (not hydration) also avoids the `featureTogglesInitialized` timing at `startWallet`. Also hardened hydration to normalize an unknown stored value to the default (`normalizeAmountFormat`).

src/screens/AmountFormat.js
raul-oliveira: Confirmed — `prettyValue` `padStart`s the fraction to exactly `DECIMAL_PLACES` (`numbers.js:59`), so it does not expose more precision than the token has, and on 2-decimal Hathor compression is a no-op. Addressed at the source rather than at the preview: the home now renders through the network's `serverInfo.decimal_places` (new `renderHomeValue`), so sub-1 values render with full precision and can actually reach the compressor on networks that expose more decimals. On 2-decimal Hathor it stays inert (flag-gated, harmless).

src/utils.js
raul-oliveira: Not intended — aligned the modes. Compression now takes only the leading zero run and keeps every later digit verbatim (removed the trailing-zero trim): `0.0000005000 → 0.0₆5000`, so it shows the same significant digits as Expanded.

src/utils.js
raul-oliveira: Renamed the param `isInteger` → `isNFT` to match every call site.

src/screens/MainScreen.js
raul-oliveira: `renderValue`/`renderHomeValue` are pure utils also used outside React (e.g. `workers/pushNotificationHandler.js`), so they can't read a selector directly. Instead I centralized the gating in one selector, `getDisplayAmountFormat(state)`, that every display site reads (MainScreen, Dashboard token list, TxDetailsModal via prop) — a single source of truth for the effective format. That also caught the Dashboard token list, which was rendering without the format at all.

src/components/TokenSelect.js
```diff
@@ -61,7 +65,7 @@ const TokenSelect = (props) => {
           <View style={styles.itemRightWrapper}>
             {tokenState === TOKEN_DOWNLOAD_STATUS.READY && (
               <Text style={styles.balanceText}>
-                {renderValue(balance, isTokenNFT(item.uid, props.tokenMetadata))}
+                {renderHomeValue(balance, isNFT, props.amountFormat, props.decimalPlaces)}
```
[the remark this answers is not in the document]
raul-oliveira: clarify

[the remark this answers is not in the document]
raul-oliveira: This is intentional. In this build the Home tab renders the multi-token list through `TokenSelect` (`IS_MULTI_TOKEN` → `DashboardStack` → `Dashboard` → `TokenSelect`, `App.js:434`; `_IS_MULTI_TOKEN = true` in `config.js`), so these rows *are* the prominent Home balance the magnitude cap targets — not a separate list.

The truncation is deliberate as well: on the Home we cap decimals by magnitude and truncate rather than round, to never overstate a balance. A sub-1 value losing its lower-order tail in Compressed mode is the accepted trade-off for a compact, consistent Home.

This is inert on the current 2-decimal network (`prettyValue` emits 2 decimals, ≤ every tier cap); it only affects networks with more decimal places. Keeping as-is.

src/screens/ConfirmImportScreen.js
```diff
@@ -71,7 +72,7 @@ const ConfirmImportScreen = () => {
       return `— ${token.symbol}`;
     }
     const available = entry.data?.available ?? 0n;
-    return `${numberUtils.prettyValue(available, constants.DECIMAL_PLACES)} ${token.symbol}`;
```
pedroferreira1, line 74: Can we remove this constant value from the code?

pedroferreira1 (dismissed):

src/screens/ConfirmImportScreen.js
raul-oliveira: Here we are using the decimal places that cames from the server (redux), it's not a removal it's a replacement.

__tests__/utils/amountFormat.test.js
```diff
@@ -0,0 +1,149 @@
+import { describe, it, expect } from '@jest/globals';
```
tuliomir, line 1: issue(non-blocking): these tests are never executed — not by CI, not locally

Pinned here for delivery; the actual concern is `.github/workflows/main.yml`, which is outside this PR's diff.

Finding №1 asked for tests because a regression in the zero-counter would ship green. These tests are good — they cover every branch that finding named — but nothing runs them, so that sentence is still true.

The CI job *named* `test (22.x)` has no jest step: its steps are checkout, node, install, gettext, check version, Linter, check_i18n. It reports green on this SHA without executing a single assertion. Locally every suite dies in `jestMockSetup.js:67`, which calls `jest.mock('unleash-proxy-client', …)` for a module that appears zero times in `package-lock.json` and is not a dependency of `@hathor/unleash-client@0.1.0` — so no `npm ci` installs it.

Both problems are pre-existing and outside this PR, so fixing them here is scope creep. But worth a follow-up issue: add `npm test` to the workflow, and either mark that mock `{ virtual: true }` or add the missing dep. Until then these assertions are documentation rather than a gate, and finding №1 should not be recorded as verified.

src/utils.js
```diff
@@ -372,21 +372,188 @@ export const getLightBackground = (alpha) => {
...
+  return `${sign}0.0${toSubscript(leadingZeros)}${significant}`;
+};
+
+/**
+ * Render value in a formatted way for display.
  *
  * @param {bigint} amount The token amount as BigInt
- * @param {boolean} isInteger Whether the token is an NFT or regular token
+ * @param {boolean} isNFT Whether the token is an NFT (rendered as an integer)
+ * @param {number} [decimalPlaces] Network decimal places (serverInfo.decimal_places).
```
tuliomir, line 440: suggestion(non-blocking): the decimalPlaces sweep skipped call sites that already have the value in scope

Pinned to this new JSDoc contract; the concrete sites are `src/screens/TokenSwap.js:170,172,332` and `src/screens/SendAmountInput.js:121`, both outside the diff.

This comment states that callers with store access should always pass `decimalPlaces`. `TokenSwap.js:100-101` selects it and passes it down to children at `:371,:401`, yet still calls `renderValue` with two arguments at `:170,:172,:332`. `SendAmountInput.js:38-39` has the same shape and misses `:121`.

`SendConfirmScreen.js:76,234,276,361`, `TokenSwapReview.js:319` and the `Reown/*` components do not select it at all — a bigger job, reasonable to defer.

None of these are regressions from this PR. But the contract added here is what makes them wrong, and the four that already have the selector in scope are a one-argument fix each.

src/screens/MainScreen.js
```diff
@@ -484,7 +505,7 @@ class TxListItem extends React.Component {
             <Text style={style.secondaryText}>{timestamp}</Text>
             <Text style={[style.secondaryText, style.bold]}>{item.getVersionInfo()}</Text>
           </View>
-          <Text style={style.balance} numberOfLines={1}>{balanceStr}</Text>
+          <Text style={style.balance}>{balanceStr}</Text>
```
tuliomir, line 508: question(non-blocking): dropping `numberOfLines={1}` lets history-row balances wrap

This line lost `numberOfLines={1}` while the style gained `maxWidth: '45%'` (`:356`), and the sibling description view got the same cap (`:342`).

With the clamp gone, a balance wider than 45% of the row now wraps to a second line and grows the row instead of ellipsizing — so a single long value makes one history row taller than its neighbours.

Was that intended? If the goal was to let the wider strings breathe, `numberOfLines={2}` bounds it; if rows should stay uniform, keep `numberOfLines={1}` alongside the new `maxWidth`.

src/utils.js
```diff
@@ -372,21 +372,188 @@ export const getLightBackground = (alpha) => {
...
+    maxDecimals = 8;
+  } else if (integerDigits.length <= 3) {
+    maxDecimals = 4;
+  } else if (integerDigits.length === 4) {
+    maxDecimals = 3;
+  } else {
+    maxDecimals = 2;
+  }
+
+  const capped = decimalPart.slice(0, maxDecimals);
```
tuliomir, line 500: nitpick: the magnitude cap can truncate a small balance to all-zeros

`capDecimalsByMagnitude` allows 8 decimals below 1 and truncates the rest, so on a network with `decimal_places > 8` an amount under 1e-8 renders as `0.00000000`. `compressAmountString` then leaves it alone, because `significant === ''` — a non-zero balance displays as zero.

Latent today (Hathor networks use 2), and truncating rather than rounding is the right call. Worth a short note on the `8` recording that it assumes network precision ≤ 8, so the assumption surfaces if that ever changes.

tuliomir (dismissed):

__tests__/utils/amountFormat.test.js
raul-oliveira: It will be addressed in the following issue #899

src/utils.js
raul-oliveira: Token swap and send token flow will be addressed in the following prs

src/screens/MainScreen.js
raul-oliveira: Yes it's intended. here is the figma link: https://www.figma.com/design/2xRserWXaSJAgkbiQlPzhr/Amounts-Update?node-id=402-3394&t=BwA0kkK6va1ayqzM-4

src/utils.js
raul-oliveira: Done in c2db0a6

src/utils/tokenSwap.js
```diff
@@ -355,9 +355,10 @@ export function selectTokenSwapContractId(state) {
  * Render amount and token symbol for UI components.
  * @param {number|bigint} amount
  * @param {TokenData} token
+ * @param {number} [decimalPlaces] Network decimal places (serverInfo.decimal_places)
  */
-export function renderAmountAndSymbol(amount, token) {
-  return `${renderValue(amount, false)} ${token.symbol}`;
+export function renderAmountAndSymbol(amount, token, decimalPlaces) {
+  return `${renderValue(amount, false, decimalPlaces)} ${token.symbol}`;
```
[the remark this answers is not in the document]
raul-oliveira: token swap references will be addressing in the following prs.

tuliomir (approved):

pedroferreira1 (approved):

PR #893, opened 2026-07-09
refactor: extract shared Radio components and restyle Address Mode

raul-oliveira:
### Acceptance Criteria
- Extract the duplicated radio selector and save button from the Amount Format and Address Mode screens into shared, reusable components (`RadioButton`, `RadioOption`, `RadioGroup`, `PreferenceSaveButton`)
- Amount Format keeps its exact look, now composed from the shared components
- Address Mode adopts the Amount Format visual style: a single bordered card with a divider, primary-colored radios, bold titles, muted descriptions and italic hints
- Address Mode preserves its existing behavior: the single-address availability check, the disabled Single option + warning banner, and the "changing address mode" saving modal
- Stacked on #891 — base branch is `feat/amount-format-settings` (not `master`)

### Screenshot
image

### Security Checklist

## Summary by CodeRabbit

- **New Features**
  - Added reusable radio-selection controls with support for titles, descriptions, badges, hints, and disabled states.
  - Added a shared “Save Preferences” button with active and disabled styling.
  - Updated address mode and amount format settings with consistent radio groups and save actions.

- **UI Improvements**
  - Added clearer introductory content and streamlined preference-selection layouts.
  - Preserved relevant warnings and selection constraints, including unavailable address modes.


__tests__/components/PreferenceSaveButton.test.js
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable */
```
raul-oliveira: remove this testfile from the pr.

__tests__/components/Radio.test.js
```diff
@@ -0,0 +1,64 @@
+/* eslint-disable */
```
raul-oliveira: remove this testfile from the pr.

__tests__/components/PreferenceSaveButton.test.js
raul-oliveira: Done — removed in f3958e0.

__tests__/components/Radio.test.js
raul-oliveira: Done — removed in f3958e0.

src/components/Radio/RadioButton.js
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * A single radio circle: outer ring plus an inner dot when selected.
+ *
+ * @param {Object} props
+ * @param {boolean} props.selected - Whether this radio is selected.
+ * @param {boolean} [props.disabled] - Greyed out; the inner dot is suppressed.
+ */
+export default function RadioButton({ selected, disabled = false }) {
+  const borderColor = selected && !disabled ? COLORS.primary : COLORS.borderColorDark;
```
tuliomir: issue(non-blocking): a selected-but-disabled radio renders as unselected, so Address Mode can show no selection at all

`selected && !disabled` gates both the ring color here and the inner dot on line 23, so a disabled option that *is* the current selection renders pixel-identical to an unselected one. The component discards state it was handed.

This is reachable, not theoretical. `detectTxOutsideFirstAddress` (`AddressMode.js:48`) returns `true` on *any* thrown error — it fails closed by design. So whenever that wallet-lib call hiccups while the wallet is already in SINGLE mode, `singleDisabled` is true for the row that is also `selected`. The card then shows **zero** radios filled, while Save is simultaneously dead because `selectedMode === currentMode`. The banner explains why Single is unavailable — not why nothing is selected. Same state is reachable without any error: a SINGLE-mode wallet that receives funds on a previously-derived index > 0.

Worth being precise: this is inherited, not introduced — the old `renderRadioIcon(isSelected, isDisabled)` did the same. What changed is its status. It was one screen's local quirk; it's now a contract binding every future consumer of the shared primitive, which makes this the moment to re-examine it.

```suggestion
  const borderColor = selected ? (disabled ? COLORS.borderColorDark : COLORS.primary) : COLORS.borderColorDark;
```

...and let the inner dot render on `selected` alone, following the same color. That keeps the greyed-out affordance while leaving the user's actual selection visible, and makes `selected` and `disabled` genuinely orthogonal — which is what the prop names already promise.

The JSDoc on line 17 mis-describes this twice, incidentally: nothing greys out (`borderColorDark` is the exact color of the normal unselected ring), and it omits that `disabled` overrides the ring even when `selected`.

src/components/PreferenceSaveButton.js
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Full-width primary save button used by the preference screens. Renders only
+ * the button; the screen supplies its own outer container/margins.
+ *
+ * @param {Object} props
+ * @param {string} props.title
+ * @param {Function} props.onPress
+ * @param {boolean} props.disabled
+ */
+export default function PreferenceSaveButton({ title, onPress, disabled }) {
```
tuliomir: suggestion: this reimplements `NewHathorButton`, in a PR whose purpose is de-duplication

`src/components/NewHathorButton.js` is already the app's primary button — **48 call sites**, including sibling settings screens (`ImportTokensScreen`, `ConfirmImportScreen`, `NanoContractRegisterScreen`, the `NetworkSettings` screens) — and already accepts `title`/`onPress`/`disabled` plus `secondary`/`danger`/`discrete` variants and `wrapperStyle`/`textStyle` escape hatches.

The overlap is near-total, and one delta is a literal no-op: `COLORS.black` (`themes.js:17`) and `NewHathorButton`'s `COLORS.textColor` (`themes.js:41`) are **both `#000`** — the same pixel via two tokens. The rest is `paddingVertical: 16` vs `height: 48`, `fontWeight: '600'` vs `bold` + `textTransform: 'uppercase'`, and the one that genuinely diverges: a **third** disabled appearance for primary buttons app-wide (`borderColorMid` `#dcdcdc` here vs `textColorShadowOpacity005` `rgba(0,0,0,0.05)` everywhere else).

Reimplementing also drops `numberOfLines={1} adjustsFontSizeToFit`, which is how `NewHathorButton` survives long translations. Currently harmless — pt-br `SALVAR PREFERÊNCIAS` still fits.

Prefer `NewHathorButton` with `wrapperStyle`/`textStyle`, or add a variant to it. If the design genuinely mandates a different disabled treatment, that belongs *in* `NewHathorButton` so all 48 call sites converge rather than diverge.

Two smaller notes. The name describes the screens that currently use it rather than what it is (the JSDoc says as much), which predicts poor reuse in both directions. And because `textTransform` was dropped, the uppercase is baked into the translatable msgid `SAVE PREFERENCES` instead of applied by style — translators must remember to uppercase, and some languages have no case distinction.

src/components/Radio/RadioOption.js
```diff
@@ -0,0 +1,108 @@
...
+export default function RadioOption({
+  selected,
+  onPress,
+  disabled = false,
+  title,
+  badge,
+  description,
+  hint,
+}) {
+  const titleColor = disabled ? COLORS.midContrastDetail : COLORS.black;
```
tuliomir: suggestion: the disabled row's text hierarchy inverts — the hint outranks the title it disables

`disabled` greys only the title. `description` (line 59) and `hint` (line 61) are unconditional, where the pre-PR Address Mode explicitly greyed the body too: `const bodyColor = isDisabled ? COLORS.midContrastDetail : COLORS.black`.

With the actual theme values that means a disabled row renders title `#9e9e9e`, description `#9e9e9e`, hint `#808080` — so the italic hint becomes the **darkest, most prominent text in a row that is supposed to read as unavailable**, and the semibold title is no longer distinguishable from its own description. That's an inverted hierarchy rather than a muted one, and it's the state Address Mode users actually hit whenever Single is unavailable.

Thread `disabled` through the text colors — a disabled variant for description/hint, or a reduced-opacity wrapper on `textWrapper` — so the row degrades as a unit.

The JSDoc on line 25 ("Non-interactive; the title greys out") should then match; it currently also omits that `disabled` is forwarded down to `RadioButton`.

src/components/Radio/RadioGroup.js
```diff
@@ -0,0 +1,59 @@
...
+ */
+export default function RadioGroup({ value, onChange, options }) {
+  return (
+    <View style={styles.card}>
+      {options.map((option, index) => (
+        <React.Fragment key={String(option.value)}>
+          {index > 0 && <View style={styles.divider} />}
+          <RadioOption
+            selected={value === option.value}
+            onPress={() => onChange(option.value)}
```
tuliomir: suggestion: `onChange` fires unconditionally — the disabled guard now has a single point of failure

The pre-PR screen enforced this twice: `disabled={isDisabled}` on the touchable *and* `if (!isDisabled)` inside the handler. The extraction kept only the outer half, so the sole remaining guard is `disabled` reaching `TouchableOpacity` at `RadioOption.js:45`.

**This isn't broken today** — RN honors `disabled` reliably. The concern is that a load-bearing wallet invariant now depends on one prop threading correctly through a generic, screen-agnostic component whose next editor has no reason to know it's load-bearing. Anything that stops `disabled` reaching the touchable (swapping to `Pressable`, wrapping the row, adding `onLongPress`) makes Single selectable when `hasTxOutside` is true — which costs the user a full wallet reload that the lib then silently reverts to `GAP_LIMIT` anyway.

Restoring it has a second payoff: it moves the invariant out of RN's touch-responder layer and into plain JS, which is the only layer this repo could actually test.

```suggestion
            onPress={() => { if (!option.disabled) onChange(option.value); }}
```

src/components/Radio/RadioOption.js
```diff
@@ -0,0 +1,108 @@
...
+  return (
+    <TouchableOpacity
```
tuliomir, line 41: suggestion: the new shared radio primitive ships with no accessibility contract

This renders a bare `TouchableOpacity` with no `accessibilityRole='radio'` and no `accessibilityState`; `RadioGroup.js:27` likewise has no `accessibilityRole='radiogroup'`. A screen reader announces these rows as unlabeled tappable text and never conveys which option is selected, or that Single is unavailable — selection is communicated purely by the inner dot, which is decorative to assistive tech.

To be fair, this isn't a violation of established practice: `accessibilityRole` appears in exactly one file today (`TokenImportBanner.js`). Raising it anyway because these are brand-new *shared primitives* whose whole purpose is reuse, and both props needed are already in scope — two lines now versus a cross-cutting retrofit later.

```suggestion
    <TouchableOpacity
      accessibilityRole='radio'
      accessibilityState={{ checked: selected, disabled }}
```

src/components/Radio/RadioGroup.js
```diff
@@ -0,0 +1,59 @@
...
+import { View, StyleSheet } from 'react-native';
+import { COLORS } from '../../styles/themes';
+import RadioOption from './RadioOption';
+
+/**
+ * A bordered card grouping radio options, separated by dividers. Controlled:
+ * the caller owns the selected `value` and receives it back via `onChange`.
+ *
+ * @param {Object} props
+ * @param {*} props.value - The currently selected option value.
```
tuliomir: nitpick: `value` is documented as `*`, but the implementation requires a primitive

`@param {*}` advertises that objects are legal, but `key={String(option.value)}` (line 29) collapses any two objects to `'[object Object]'` and collapses `1` with `'1'`, and `value === option.value` (line 32) never matches a non-referential object. Both current callers pass frozen string enums, so this is latent — it fires the moment someone builds `options` from dynamic data, which is the natural third use for this component.

Worth documenting the two relational invariants in the same block while you're here, since neither is checked: `value` must match exactly one option (otherwise every row renders unselected, silently, with no diagnostic), and `option.value` must be unique (duplicates make two rows render selected at once in a single-select control, signalled only by a React duplicate-key warning that's easy to miss in an RN log).

```suggestion
 * @param {string|number} props.value - The currently selected option value. Compared with `===` and stringified for React keys, so option values must be unique primitives, and exactly one option should match.
```

src/components/PreferenceSaveButton.js
```diff
@@ -0,0 +1,58 @@
...
+import { COLORS } from '../styles/themes';
+
+/**
+ * Full-width primary save button used by the preference screens. Renders only
+ * the button; the screen supplies its own outer container/margins.
+ *
+ * @param {Object} props
+ * @param {string} props.title
+ * @param {Function} props.onPress
+ * @param {boolean} props.disabled
```
tuliomir: nitpick: `disabled` optionality is documented three different ways across the three new components

`RadioButton` and `RadioOption` both mark it `[props.disabled]` and default it to `false` in the destructure. Here it's documented as **required** (no brackets) with no default, so an omitted `disabled` silently falls into the *active* branch on lines 24 and 28 — an enabled save button.

Behaviorally harmless today since both callers pass it explicitly, but three components landing in one PR should tell one consistent story about the same prop name.

```suggestion
 * @param {boolean} [props.disabled=false]
```

...plus `disabled = false` in the destructure on line 21.

src/components/Radio/index.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Copyright (c) Hathor Labs and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+export { default as RadioButton } from './RadioButton';
```
tuliomir: nitpick: the barrel exports two components nothing outside this folder uses

Only `RadioGroup` is imported externally (`AddressMode.js:20`, `AmountFormat.js:13`). `RadioButton` in particular has no press handling and no standalone use — exporting it publicly invites someone to hand-build a second, divergent radio row, which is the exact duplication this PR set out to remove.

Consider trimming the barrel to `RadioGroup` until something outside the folder actually needs the others.

src/screens/AddressMode.js
```diff
@@ -198,69 +151,58 @@ export default function AddressMode({ navigation }) {
...
+              },
+            ]}
+          />
+        </View>
+
+        {singleDisabled && (
+          <View style={styles.warningBanner}>
+            <Text style={styles.warningIcon}>⚠</Text>
+            <Text style={styles.warningText}>
+              {t`You can't switch to single address mode because other addresses in your wallet are already in use.`}
```
tuliomir, line 192: question: this banner asserts a cause that isn't established on the check-failure path

It states "other addresses in your wallet are already in use" as fact, but `singleDisabled` is also true when `hasTxOutsideFirstAddress()` **threw** and `detectTxOutsideFirstAddress` returned its conservative `true` (line 48). On that path we're telling the user something we never actually determined.

Entirely pre-existing behavior that this PR only relocates, so very reasonable to defer — flagging it because the lines are in the diff and the fail-closed fallback makes the path more reachable than it looks. Would a softer phrasing ("Single address mode isn't available right now") be more honest across both paths?

src/components/Radio/RadioButton.js
```diff
@@ -0,0 +1,43 @@
...
+    borderRadius: 12,
+    borderWidth: 2,
+    justifyContent: 'center',
+    alignItems: 'center',
+  },
+  inner: {
+    width: 12,
+    height: 12,
+    borderRadius: 6,
+    backgroundColor: COLORS.primary,
```
tuliomir: thought: the app now carries two unrelated selection-indicator idioms

This new ring-and-dot radio sits alongside the existing square-with-checkmark selection indicator in `TokenItem.js:112-127`, which styles itself from `COLORS.controlBorder` — a token the new Radio components don't touch.

Neither is wrong, but a PR that consolidates selection UI is the natural moment to note that the app now has two visual vocabularies for "this one is chosen", drawn from two different color tokens. Follow-up material, not this PR's job.

tuliomir (dismissed):

src/components/Radio/RadioButton.js
raul-oliveira: image

raul-oliveira: Fixed in 821216a. The two props no longer cancel each other out: `selected` decides whether the dot shows up, and `disabled` only decides what color it and the ring are drawn in. So a selected option that is also disabled keeps its dot — it's just grey instead of purple.

One thing to note: I used `borderColorDark` for the disabled color, which as you pointed out is the same grey as a normal unselected ring. That means only the dot tells the two apart. Let me know if that's too subtle and I'll move it to `midContrastDetail`.

src/components/PreferenceSaveButton.js
raul-oliveira: Agreed — I deleted `PreferenceSaveButton` and both screens now use `NewHathorButton`. I also changed the msgid to sentence case (`Save preferences`) so the uppercase comes from `textTransform`, like the other settings screens do it.

One thing to check before this merges: the disabled button now uses the app-wide grey (`rgba(0,0,0,0.05)` / `rgba(0,0,0,0.5)`) instead of the darker `#dcdcdc` / `#808080`, so it looks lighter than it did. If the design really needs the darker grey, tell me and I'll add it as a variant on `NewHathorButton` instead — that way the other 48 buttons get it too.

src/components/Radio/RadioOption.js
raul-oliveira: Good catch on the hierarchy. I now fade the whole text block with `opacity: 0.5` instead of giving each line its own disabled color, so the three keep their order and anything we add later fades with them. Also dropped `titleColor` for a plain `color: COLORS.black` on the title.

On the JSDoc I went the other way: it now just says `If true, the option is disabled.` You were right that the old wording was wrong — the whole row greys out, not just the title — but I don't think the fix is to describe the forwarding. `RadioOption` is built out of `RadioButton` plus the text, so disabling it has to disable what's inside; there's nothing extra to document. MUI, Polaris, Ant and Radix all leave it out too.

src/components/Radio/RadioGroup.js
raul-oliveira: Restored, thanks. I took your suggestion and added a short comment saying the double check is on purpose — otherwise it looks like a pointless extra `if` and someone deletes it again.

src/components/Radio/RadioOption.js
raul-oliveira: Added. I took your suggestion on `RadioOption` and also put `accessibilityRole='radiogroup'` on the `RadioGroup` card, since you mentioned that in the same comment. Agreed it's much easier to do this now than to go back and add it to every screen later.

src/components/Radio/RadioGroup.js
raul-oliveira: Changed both `value` and `options[].value` to `string|number`. `options[].value` has the same constraint through `key` and `===`, so fixing only one side would have been odd.

I did leave out the sentence from your suggestion. `key={String(option.value)}` and `selected={value === option.value}` are both a few lines below it in a 30-line file, and "React keys have to be unique" isn't specific to this component — so the sentence mostly repeated what the reader can already see, and the type does the real work. Tell me if you'd rather have it spelled out anyway and I'll put it back; for a shared component I can see the argument.

src/components/PreferenceSaveButton.js
raul-oliveira: This went away with `PreferenceSaveButton` (your other comment), so the odd third case is gone. While I was there I also made the remaining two match MUI's wording — `If true, the radio is selected.` / `If true, the option is disabled.` — so `selected` and `disabled` now read the same way in both components, not just the brackets.

src/components/Radio/index.js
raul-oliveira: Now exports `RadioGroup` only. The other two are imported by relative path inside the folder, so nothing else had to change. Keeping them private means nobody outside can build a second radio row by hand, and we can open them up later if something actually needs them.

src/screens/AddressMode.js
raul-oliveira: Falling all the way back to multi is a deliberate design decision, not an accident of the error handling. If we can't confirm the wallet has no activity past index 0, we keep it on multi — switching to single there would hide funds on the other addresses, so blocking is the right answer whether the check failed or came back positive.

That leaves the wording, which is fair. It's a copy change though, and doing it properly means telling "we checked and found activity" apart from "we couldn't check" — one boolean can't hold both. I'd rather not put that in a PR about extracting shared components; happy to open an issue for it.

src/components/Radio/RadioButton.js
raul-oliveira: They're two different components: a radio is a single choice, a checkbox is multiple choice. The shapes are different on purpose, so I'd leave both as they are.

src/screens/AddressMode.js
```diff
@@ -198,69 +151,58 @@ export default function AddressMode({ navigation }) {
...
-              styles.saveButtonText,
-              isSaveDisabled
-                ? styles.saveButtonTextDisabled
-                : styles.saveButtonTextActive,
-            ]}
-          >
-            {t`SAVE PREFERENCES`}
-          </Text>
-        </TouchableOpacity>
+        />
```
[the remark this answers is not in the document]
raul-oliveira: Fixed in 6be95a2. `disabled` is now `isSaveDisabled || saving`, and the guard at the top of `onSave` matches. I also had to add `saving` to the `useCallback` deps — without it the callback closes over a stale `saving` and the new guard never fires.

Note the button is `NewHathorButton` now, not `PreferenceSaveButton` — that component was removed in an earlier review round.

One caveat on how far this goes: `saving` is React state, so two taps inside the same frame — before `setSaving(true)` commits — would both read `false` and get through. The `disabled` prop closes it from the first re-render on, which is the case that actually happens. A `useRef` would make it airtight if you think it is worth it.

`AmountFormat.js` does not need the same change: its `onSave` is synchronous (`STORE.setItem` + `dispatch` + `goBack`), with no wallet call and no `saving` state, so there is no window.

src/screens/AddressMode.js
tuliomir: Opened issue #900 for that. Thanks!

tuliomir (approved):

pedroferreira1 (approved):

PR #894, opened 2026-07-14
feat(decimal): decimal precision on send/swap, and amount input display fix

raul-oliveira:
Builds on #891 (amount format foundation) and extends the decimal-places fix to the send and token-swap screens. Aligns the amount input and amount displays with the [Amounts Update Figma](https://www.figma.com/design/2xRserWXaSJAgkbiQlPzhr/Amounts-Update).

### Acceptance Criteria
- The amount input accepts up to 8 decimal places, independent of the selected token's precision — the value is still scaled to the token's `decimal_places` when the transaction is built.
- Long amounts shrink to stay on one line and, once they reach the minimum size, wrap across up to 3 lines instead of shrinking indefinitely.
- On the Send screen the token selector drops below the amount when it wraps, with no font-size jump at the transition.
- Read-only balances (transaction details, home available and locked) shrink and wrap the same way — previously they stayed at full size and only wrapped on iOS.
- The send flow (amount input and confirmation) and token swap (swap, review and token picker) render amounts using the network's `decimal_places` instead of the wallet-lib default of 2, extending the #891 decimal fix to these screens.

### Security Checklist

### Screenshots
Token swap
image
image
image

Send transaction

https://github.com/user-attachments/assets/3dc1e814-04d4-4213-887c-5504a6810a48

image
image

## New Features
- Added responsive amount displays that shrink or wrap values to fit available space.
- Amount inputs now support controlled font sizing and optional single-line layouts.
- Amount formatting consistently respects server-provided decimal precision across transfers, balances, fees, and swaps.

## Bug Fixes
- Improved amount layouts to prevent overflow across balance, token creation, and send screens.
- Pasted line breaks are handled more reliably, with support for up to 8 decimal places.
- External styling now applies correctly to token containers.

### DIFF
--- a/src/components/AmountTextInput.js
+++ b/src/components/AmountTextInput.js
@@ -7,23 +7,13 @@
 
 import React, { useState, useRef, useEffect, forwardRef, useImperativeHandle } from 'react';
 import { StyleSheet, TextInput } from 'react-native';
-import { getAmountParsed, getIntegerAmount } from '../utils';
+import { getAmountParsed, getIntegerAmount, computeAmountFontFit, AMOUNT_MAX_LINES } from '../utils';
+import { MAX_DECIMAL_PLACES } from '../constants';
 import { COLORS } from '../styles/themes';
 
-// Auto-shrink lower bound: never scale the font below this fraction of
-// the base size, so very long amounts stay readable even when scaled.
-const MIN_FONT_SCALE = 0.5;
-
-// Average glyph width as a fraction of fontSize for the bold sans-serif
-// AmountTextInput uses. We don't measure rendered text (an earlier
-// implementation that did caused a Folly F14Set assertion crash on
-// iOS 26.1 + Fabric due to the extra hidden <Text> re-rendering on
-// every keystroke), so we estimate width as
-// `length * fontSize * GLYPH_RATIO`. Slightly conservative for digits
-// and slightly generous for `,` / `.` — net effect is the font shrinks
-// a hair sooner than strictly necessary, which is preferable to letting
-// the value clip behind the token selector.
-const GLYPH_RATIO = 0.55;
+// Rendered line height as a fraction of the font size, reserving room for the
+// wrapped lines without clipping the bold glyphs.
+const LINE_HEIGHT_RATIO = 1.2;
 
 /**
  * Text input component specifically for handling token amounts with BigInt validation.
@@ -37,7 +27,9 @@ const GLYPH_RATIO = 0.55;
  *                                                   (no decimals)
  * @param {Object} [props.style] - Additional styles for the TextInput
  * @param {boolean} [props.autoFocus] - Whether the input should be focused on mount
- * @param {number} [props.decimalPlaces] - Number of decimal places to use
+ * @param {number} [props.decimalPlaces] - Number of decimal places for the token's value scale
+ * @param {number} [props.fontSize] - Explicit display font size in px (parent-controlled mode)
+ * @param {boolean} [props.singleLine] - Pin the input to a fixed one-line height (no wrapping)
  * @param {React.Ref} ref - Forwarded ref, exposes the focus() method
  * @returns {React.ReactElement} A formatted amount input component
  */
@@ -46,21 +38,8 @@ const AmountTextInput = forwardRef((props, ref) => {
   const [text, setText] = useState(props.value || '');
   const { decimalPlaces } = props;
 
-  // Auto-shrink-to-fit: amounts can grow long enough to overflow the
-  // input's column (e.g. `957,791,973.79`). We capture the column
-  // width via the TextInput's `onLayout` and scale `fontSize` down
-  // when the estimated text width (length × fontSize × GLYPH_RATIO)
-  // exceeds it. When the user shortens the value, the estimate drops
-  // and the font scales back up to the base size.
-  //
-  // IMPORTANT: `onLayout` measures this TextInput's OWN box, so callers
-  // MUST give it a parent-determined width — `flex: 1` when it shares a
-  // row (e.g. with a TokenBox), or `alignSelf: 'stretch'` / an explicit
-  // width when it sits alone in an `alignItems: 'center'` column. Without
-  // that the input sizes to its content; since `fontSize` (which this
-  // logic controls) also drives the content width, the measurement feeds
-  // back into itself and the font collapses to `minFontSize`, flickering
-  // on the way down.
+  // Self-measure: the input's own column width, captured via onLayout, drives the
+  // font ladder. Unused when the parent passes an explicit fontSize.
   const [containerWidth, setContainerWidth] = useState(0);
 
   // Expose the focus method to parent components
@@ -98,14 +77,18 @@ const AmountTextInput = forwardRef((props, ref) => {
       return;
     }
 
-    let parsedText = newText;
+    // The numeric keyboard shouldn't emit newlines, but a multiline input can
+    // receive them via paste; strip so they never reach the parser.
+    let parsedText = newText.replace(/[\n\r]/g, '');
     let bigIntValue;
     if (props.allowOnlyInteger) {
       // We allow only integers for NFT
       parsedText = parsedText.replace(/[^0-9]/g, '');
     }
 
-    parsedText = getAmountParsed(parsedText, decimalPlaces);
+    // Accept up to MAX_DECIMAL_PLACES typed decimals regardless of the token's
+    // precision; the value below is still scaled to the token's decimalPlaces.
+    parsedText = getAmountParsed(parsedText, MAX_DECIMAL_PLACES);
 
     // There is no NaN in BigInt, it either returns a valid bigint or throws
     // an error.
@@ -135,36 +118,40 @@ const AmountTextInput = forwardRef((props, ref) => {
     placeholder = `0.${zeros}`;
   }
 
-  const { style: customStyle, textAlign, ...restProps } = props;
+  const { style: customStyle, textAlign, fontSize, singleLine, ...restProps } = props;
 
-  // Resolve the base (unscaled) font size from the merged style chain
-  // so callers that override fontSize via `customStyle` still get
-  // correct scaling math.
-  const flatStyle = StyleSheet.flatten([style.input, customStyle]) || {};
-  const baseFontSize = flatStyle.fontSize ?? 32;
-  const minFontSize = Math.max(14, Math.floor(baseFontSize * MIN_FONT_SCALE));
+  const isControlledSize = fontSize != null;
   const displayed = text || placeholder;
-  const estimatedWidth = displayed.length * baseFontSize * GLYPH_RATIO;
-  const scaledFontSize = (containerWidth > 0 && estimatedWidth > containerWidth)
-    ? Math.max(
-      minFontSize,
-      Math.floor(baseFontSize * (containerWidth / estimatedWidth)),
-    )
-    : baseFontSize;
+  const resolvedFontSize = isControlledSize
+    ? fontSize
+    : computeAmountFontFit(displayed.length, containerWidth).fontSize;
+  const lineHeight = Math.round(resolvedFontSize * LINE_HEIGHT_RATIO);
+  const heightStyle = singleLine
+    ? { height: lineHeight }
+    : { maxHeight: lineHeight * AMOUNT_MAX_LINES };
 
   return (
     <TextInput
       ref={inputRef}
-      style={[style.input, customStyle, { fontSize: scaledFontSize }]}
+      style={[
+        style.input,
+        customStyle,
+        { fontSize: resolvedFontSize, lineHeight },
+        heightStyle,
+      ]}
       onChangeText={onChangeText}
       value={text}
+      multiline
+      scrollEnabled={false}
       textAlign={textAlign || 'center'}
-      textAlignVertical='bottom'
+      textAlignVertical='center'
       keyboardAppearance='dark'
       keyboardType='numeric'
       placeholder={placeholder}
       placeholderTextColor={COLORS.midContrastDetail}
-      onLayout={(e) => setContainerWidth(e.nativeEvent.layout.width)}
+      onLayout={isControlledSize
+        ? undefined
+        : (e) => setContainerWidth(e.nativeEvent.layout.width)}
       {...restProps}
     />
   );
@@ -172,9 +159,6 @@ const AmountTextInput = forwardRef((props, ref) => {
 
 const style = StyleSheet.create({
   input: {
-    height: 38,
-    lineHeight: 38,
-    fontSize: 32,
     fontWeight: 'bold',
     paddingVertical: 0,
     color: COLORS.textColor,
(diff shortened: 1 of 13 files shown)


src/components/AmountTextInput.js
```diff
@@ -7,26 +7,27 @@
...
+ * `multiline`, capped at AMOUNT_MAX_LINES: it shrinks to stay on one line, then
+ * wraps once the floor size can't fit it (see `computeAmountFontFit`).
+ *
+ * Sizing modes:
+ * - Self-measure (default): the input measures its own column via `onLayout` and
+ *   applies the shared ladder. Callers MUST give it a parent-determined width
+ *   (`flex: 1` in a row, `alignSelf: 'stretch'` in a centered column); otherwise the
+ *   input sizes to its content and the font-scaling measurement feeds back on itself.
+ * - Controlled (`fontSize` prop): the parent computes the size and owns any layout
+ *   changes. Used by the Send screen, which measures a stable, layout-independent
```
raul-oliveira: replace this bunch of comment with the original one: 

Text input component specifically for handling token amounts with BigInt validation.

src/components/AmountTextInput.js
```diff
@@ -135,46 +129,51 @@ const AmountTextInput = forwardRef((props, ref) => {
...
-    ? Math.max(
-      minFontSize,
-      Math.floor(baseFontSize * (containerWidth / estimatedWidth)),
-    )
-    : baseFontSize;
+  const resolvedFontSize = isControlledSize
+    ? fontSize
+    : computeAmountFontFit(displayed.length, containerWidth).fontSize;
+  const lineHeight = Math.round(resolvedFontSize * LINE_HEIGHT_RATIO);
+  // `singleLine` callers (the Send screen's inline state) get a fixed one-line
```
raul-oliveira: do we really needs this comment?

src/screens/MainScreen.js
```diff
@@ -353,6 +354,8 @@ class TxListItem extends React.Component {
     balance: {
       fontSize: 16,
       marginRight: 16,
+      // Cap the amount column and right-align it so a long value wraps within
```
raul-oliveira: remove this comment

src/screens/SendAmountInput.js
```diff
@@ -129,6 +130,23 @@ const SendAmountInput = () => {
 
   const tokenNameUpperCase = token.name.toUpperCase();
 
+  // Left ghost spacer (80, see renderGhostElement) + token box (90, see
```
raul-oliveira: do we really need this comment?

src/screens/SendAmountInput.js
```diff
@@ -129,6 +130,23 @@ const SendAmountInput = () => {
...
+  // TokenBox styles.wrapper). The measured amount-area width is layout-
+  // independent (full content width in both row and column modes), so the fit
+  // decision never feeds back on itself.
+  const TOKEN_BOX_ROW_RESERVED = 170;
+  const placeholderString = isNFT() ? '0' : `0.${'0'.repeat(decimalPlaces)}`;
+  const displayedAmount = amount || placeholderString;
+  const inlineWidth = amountAreaWidth > 0 ? amountAreaWidth - TOKEN_BOX_ROW_RESERVED : 0;
+  // Size against the width beside the token box (inlineWidth) so the amount
+  // shrinks to fit there first, then wraps. Sizing by inlineWidth rather than the
+  // full amount-area width keeps the font continuous across the row->column
```
raul-oliveira: and this batch of comments? probably it should be a line instead of all of this.

src/screens/SendAmountInput.js
```diff
@@ -140,25 +158,41 @@ const SendAmountInput = () => {
         <KeyboardAvoidingView behavior='padding' style={{ flex: 1 }} keyboardVerticalOffset={getStatusBarHeight()}>
           <View style={{ flex: 1, padding: 16, justifyContent: 'space-between' }}>
             <View>
-              <View style={{
-                flexDirection: 'row', alignItems: 'center', justifyContent: 'space-between', marginTop: 40,
-              }}
+              {/* A single View whose flexDirection switches row->column keeps the
```
raul-oliveira: remove this comment

src/utils.js
```diff
@@ -461,6 +461,61 @@ export const renderValue = (
   return formatted;
 };
 
+// Amount display sizing. Amounts can get long: the maximum integer part is
```
raul-oliveira: do we need this amount of comment here? try to simplify

src/utils.js
```diff
@@ -461,6 +461,61 @@ export const renderValue = (
...
+// 9223372036854775808 (19 digits) plus up to MAX_DECIMAL_PLACES (8) decimals and a
+// dot => ~28 chars, so the ladder below (computeAmountFontFit) shrinks the font
+// before it wraps.
+export const AMOUNT_FONT_BASE_SIZE = 32;
+export const AMOUNT_FONT_MIN_SIZE = 20;
+export const AMOUNT_MAX_LINES = 3;
+
+// Average glyph width as a fraction of fontSize for the bold amount font. We
```
raul-oliveira: same here, this coment is too verbose

src/components/AmountTextInput.js
```diff
@@ -135,46 +132,51 @@ const AmountTextInput = forwardRef((props, ref) => {
...
       placeholder={placeholder}
       placeholderTextColor={COLORS.midContrastDetail}
-      onLayout={(e) => setContainerWidth(e.nativeEvent.layout.width)}
+      onLayout={isControlledSize
+        ? undefined
+        : (e) => setContainerWidth(e.nativeEvent.layout.width)}
       {...restProps}
     />
   );
 });
```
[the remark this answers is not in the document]
raul-oliveira: Not addressing — this is pre-existing (master already leaves `value`/`decimalPlaces`/`allowOnlyInteger`/`onAmountUpdate` in `restProps`) and causes no runtime bug: every call site round-trips `value` from the state it updates via `onAmountUpdate`, so the spread `value` always equals the local `text`. React Native also silently ignores unknown props on the native `TextInput` (this isn't React DOM — there's no per-keystroke warning spam).

Separately, the proposed `_value`/`_decimalPlaces`/… destructure would introduce 4 `no-unused-vars` warnings under this repo's ESLint config (`no-unused-vars: [1, {vars: "all"}]`, no `ignoreRestSiblings`). Out of scope for this PR.

src/screens/TokenSwap.js
```diff
@@ -167,9 +167,9 @@ const TokenSwap = () => {
   useEffect(() => {
     setShowQuote(!!quote);
     if (quote) {
-      setInputTokenAmountStr(renderValue(quote.amount_in));
+      setInputTokenAmountStr(renderValue(quote.amount_in, false, decimalPlaces));
       setInputTokenAmount(quote.amount_in);
-      setOutputTokenAmountStr(renderValue(quote.amount_out));
+      setOutputTokenAmountStr(renderValue(quote.amount_out, false, decimalPlaces));
```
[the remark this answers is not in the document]
raul-oliveira: Not addressing — `decimalPlaces` comes from `serverInfo.decimal_places`, a per-network constant fixed once at wallet connect. It cannot change while the swap screen is mounted, so the `[quote]` effect can't produce stale formatting in the real user journey. Adding it to the deps would be a functional no-op.

src/screens/TokenSwap.js
```diff
@@ -423,7 +423,7 @@ const TokenSwap = () => {
                   <View style={styles.quoteRow}>
                     <Text style={styles.quoteHeader}>Conversion rate</Text>
                     <Text style={styles.quoteValue}>
-                      {renderConversionRate(quote, inputToken, outputToken)}
+                      {renderConversionRate(quote, inputToken, outputToken, decimalPlaces)}
```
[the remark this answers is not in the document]
raul-oliveira: Out of scope — `renderConversionRate` lives in `src/utils/tokenSwap.js`, which this PR does not modify (the diff only adds the `decimalPlaces` argument at the call site). The `Number()` cast is pre-existing and only loses precision above 2^53, and it affects the **informational rate label only** — the executed swap uses the BigInt `amount_in`/`amount_out` directly, so no amounts are miscalculated. Tracking as a separate follow-up rather than in this decimal-formatting PR.

src/screens/TokenSwapReview.js
```diff
@@ -303,20 +304,20 @@ const TokenSwapReview = () => {
...
-                    <Text style={styles.quoteValue}>{renderAmountAndSymbolWithSlippage('output', quote.amount_in, tokenIn, TOKEN_SWAP_SLIPPAGE)}</Text>
+                    <Text style={styles.quoteValue}>{renderAmountAndSymbolWithSlippage('output', quote.amount_in, tokenIn, TOKEN_SWAP_SLIPPAGE, decimalPlaces)}</Text>
                   </View>
                 )}
                 <View style={styles.quoteRow}>
                   <Text style={styles.quoteHeader}>{t`Network Fee`}</Text>
                   <Text style={styles.quoteValue}>
                     {networkFee != null && networkFee > 0n
-                      ? `${renderValue(networkFee, false)} ${hathorLib.constants.DEFAULT_NATIVE_TOKEN_CONFIG.symbol}`
+                      ? `${renderValue(networkFee, false, decimalPlaces)} ${hathorLib.constants.DEFAULT_NATIVE_TOKEN_CONFIG.symbol}`
```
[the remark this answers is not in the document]
raul-oliveira: Addressed — `nativeSymbol` now reads `serverInfo.native_token.symbol`, falling back to `DEFAULT_NATIVE_TOKEN_CONFIG.symbol` when server info isn't loaded, matching the existing pattern in `AmountFormat.js`. Applied in both `SendConfirmScreen.js` and `TokenSwapReview.js`.

src/constants.js
```diff
@@ -159,6 +159,10 @@ export const AMOUNT_FORMAT_DEFAULT = AMOUNT_FORMAT.EXPANDED;
 // 'wallet:' prefix: STORE.clearItems(true) sweeps it on resetWallet, so it resets to Expanded.
 export const AMOUNT_FORMAT_KEY = 'wallet:amount_format';
 
+// Max decimal places the amount input accepts, independent of the token's own
+// precision — the value is still scaled to the token's decimal_places downstream.
+export const MAX_DECIMAL_PLACES = 8;
```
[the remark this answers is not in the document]
raul-oliveira: Not addressing — this is the intended change. The requirement was updated: the amount input now accepts up to **8** decimal places (the PR description/AC was updated to match). Reverting to 18 would contradict the current requirement.

src/screens/SendConfirmScreen.js
```diff
@@ -268,12 +269,12 @@ const SendConfirmScreen = () => {
...
     if (networkFee === null) {
-      return <Text style={{ color: COLORS.textColorShadow }}>{t`Loading...`}</Text>;
+      return <Text style={[styles.summaryValue, { color: COLORS.textColorShadow }]}>{t`Loading...`}</Text>;
     }
     if (networkFee > 0n) {
       return (
-        <Text>
-          {renderValue(networkFee, false)} {nativeSymbol}
+        <Text style={styles.summaryValue}>
+          {renderValue(networkFee, false, decimalPlaces)} {nativeSymbol}
```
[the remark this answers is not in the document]
raul-oliveira: Addressed — `nativeSymbol` now reads `serverInfo.native_token.symbol`, falling back to `DEFAULT_NATIVE_TOKEN_CONFIG.symbol` when server info isn't loaded, matching the existing pattern in `AmountFormat.js`. Applied in both `SendConfirmScreen.js` and `TokenSwapReview.js`.

src/components/AmountTextInput.js
```diff
@@ -98,14 +77,18 @@ const AmountTextInput = forwardRef((props, ref) => {
...
     let bigIntValue;
     if (props.allowOnlyInteger) {
       // We allow only integers for NFT
       parsedText = parsedText.replace(/[^0-9]/g, '');
     }
 
-    parsedText = getAmountParsed(parsedText, decimalPlaces);
+    // Accept up to MAX_DECIMAL_PLACES typed decimals regardless of the token's
+    // precision; the value below is still scaled to the token's decimalPlaces.
+    parsedText = getAmountParsed(parsedText, MAX_DECIMAL_PLACES);
```
tuliomir, line 91: issue(blocking): typed decimals beyond the token's precision are silently truncated

`getAmountParsed(parsedText, MAX_DECIMAL_PLACES)` keeps 8 typed decimals, but line 97 scales with the token's `decimalPlaces` and `getIntegerAmount` slices the excess away.

On a 2-decimal network, typing `1.12345678` displays that string and sends `112n` (1.12); `0.00000001` yields `0n`, which `isValid` accepts, so the field shows a value while Continue stays disabled with no error.

Previously the text was clamped, so what you saw was what you sent. Consider `getAmountParsed(parsedText, decimalPlaces ?? MAX_DECIMAL_PLACES)`, which keeps the undefined-`serverInfo` fallback without discarding entered precision.

src/utils.js
```diff
@@ -461,6 +461,56 @@ export const renderValue = (
...
+ * Because the font is already at the floor when `wraps` flips to true, callers
+ * that switch layout at that point (Send) see no font-size jump at the switch.
+ *
+ * @param {number} textLength Length of the displayed amount (value or placeholder)
+ * @param {number} availableWidth Single-line width available for the amount, in px
+ * @param {number} [baseFontSize] Starting (largest) font size; defaults to the amount base
+ * @param {number} [minFontSize] Floor font size below which it wraps instead of shrinking
+ * @return {{ fontSize: number, wraps: boolean }}
+ */
+export const computeAmountFontFit = (
```
tuliomir, line 495: issue(non-blocking): new pure util `computeAmountFontFit` ships with no unit tests

Three-branch ladder (base / shrink / floor + wrap) with a boundary at `minFontSize`, and its `wraps` flag drives the Send screen's row→column switch. Flipping a `>=` to `>` at either boundary silently changes when the token box drops below the amount.

The function is pure and takes plain numbers — a small input/output table next to `__tests__/utils/amountFormat.test.js` pins the ladder cheaply. Worth noting the suite is a local gate only: `.github/workflows/main.yml` has no `npm test` step, so nothing here runs in CI.

src/screens/TokenSwap.js
```diff
@@ -167,9 +167,9 @@ const TokenSwap = () => {
...
     setShowQuote(!!quote);
     if (quote) {
-      setInputTokenAmountStr(renderValue(quote.amount_in));
+      setInputTokenAmountStr(renderValue(quote.amount_in, false, decimalPlaces));
       setInputTokenAmount(quote.amount_in);
-      setOutputTokenAmountStr(renderValue(quote.amount_out));
+      setOutputTokenAmountStr(renderValue(quote.amount_out, false, decimalPlaces));
       setOutputTokenAmount(quote.amount_out);
     }
   }, [quote]);
```
tuliomir, line 175: nitpick: effect reads `decimalPlaces` but doesn't list it as a dependency

If `serverInfo` ever resolved after a quote, the two amount strings would keep the precision they were formatted with. In practice `serverInfo` lands during wallet start, well before this screen mounts, so it's latent rather than live — but adding it costs nothing.

PR #896, opened 2026-07-23
feat: passkey core — PRF key derivation, external signer & xpub-only storage

pedroferreira1:
### Summary

The self-contained passkey module that a passkey wallet is built on — **first of two stacked PRs**. It's gated behind the `passkey-onboarding.rollout` flag (off) and **not yet wired into any screen**, so it ships inert; the onboarding UI and PIN-less flows that consume it land in the follow-up (`feat/passkey-integration`).

What's here:
- **`passkeyService`** — the WebAuthn PRF ceremony (create / sign in with a passkey) and `user.id` encoding (UTF-8 byte-budgeted so any wallet name stays within WebAuthn's 64-byte cap).
- **`passkeySigner`** — `derivePasskeyXpub` (PRF seed → account xpub), the external tx signer (`makePasskeyTxSigner`) that **reuses wallet-lib's `transactionUtils.signTxInputs`** instead of duplicating the signing loop, unlock verification, and the `PasskeyCancelled` / `PasskeyXpubMismatch` / `PasskeyBusy` (single-flight) error types. Error messages are plain strings here — the `ttag` wrapping + translations land in the integration PR.
- **`store`** — `initPasskeyStorage` / `walletMeta` / `isPasskeyWallet`: an xpub-only wallet persists only `{ walletType, xpub, passkeyLabel, credentialId? }` — **no seed, no PIN, no encrypted key**.
- **`constants`** — RP id/name, PRF salt, and the feature toggle (default off).
- **`react-native-passkey`** dependency + a jest virtual-mock fix (`jestMockSetup`) that unblocks the suite, plus `passkeySigner` unit tests.

### Acceptance Criteria
- Module compiles and `__tests__/passkey/passkeySigner.test.js` passes (9 tests: cancelled / xpub-mismatch / busy single-flight / happy path).
- `derivePasskeyXpub(words)` deterministically yields the account-path (`m/44'/280'/0'`) xpub.
- The external tx signer produces signatures via `transactionUtils.signTxInputs` (no duplicated signing logic).
- `initPasskeyStorage` persists **only** the xpub + label + `walletType` marker — verify no seed / encrypted key is written.
- `user.id` encoding stays ≤ 64 bytes for a CJK/emoji wallet name (byte-budget truncation, not string-length).
- The jest suite starts (previously died in `jestMockSetup` on uninstalled modules).
- **No behavior change to the app**: nothing imports the module from a screen/saga yet; the feature flag is defined but off.

### Security Checklist

## Summary by CodeRabbit

- **New Features**
  - Added passkey-based onboarding and sign-in using PRF-enabled passkeys.
  - Enabled passkey-based transaction signing and unlock verification for xpub-only wallets.
  - Persist wallet metadata for passkey wallets, including `credentialId` backfill and passkey wallet detection.
  - Added passkey onboarding feature toggle and PRF configuration (default off).
- **Tests**
  - Added deterministic passkey signing/unlock tests covering cancellation, xpub mismatch, missing metadata, credentialId backfill, and single-flight concurrency.
- **Chores**
  - Added `react-native-passkey` and adjusted Jest transform settings for affected packages.

### DIFF
--- a/__tests__/passkey/passkeySigner.test.js
+++ b/__tests__/passkey/passkeySigner.test.js
@@ -14,7 +14,9 @@ jest.mock('../../src/constants', () => ({ NETWORK_MAINNET: 'mainnet' }));
 jest.mock('../../src/store', () => ({
   STORE: {
     getWalletMeta: jest.fn(),
-    updateWalletMeta: jest.fn(),
+    // Returns a promise: the signer treats the credentialId backfill as fire-and-forget and
+    // attaches a .catch(), so the mock must be thenable or that call would throw.
+    updateWalletMeta: jest.fn(() => Promise.resolve()),
   },
 }));
 
@@ -75,9 +77,14 @@ const makeMeta = (overrides = {}) => ({
   ...overrides,
 });
 
-// A fake account-root xpriv whose derivation chain never actually runs in these tests
-// (signTxInputs is mocked and does not invoke the resolver), but is shaped correctly just in case.
+// A fake account-root xpriv shaped like the real one: BOTH derivation methods are stubbed so the
+// resolver's 'spend' branch (deriveChild = compliant) and 'legacy' branch (deriveNonCompliantChild)
+// can each run without throwing. The default happy-path tests mock signTxInputs and never invoke
+// the resolver; the dedicated resolver test below builds its own asserting root.
 const makeRoot = () => ({
+  deriveChild: jest.fn(() => ({
+    deriveChild: jest.fn(() => ({})),
+  })),
   deriveNonCompliantChild: jest.fn(() => ({
     deriveNonCompliantChild: jest.fn(() => ({})),
   })),
@@ -198,6 +205,45 @@ describe('makePasskeyTxSigner', () => {
     expect(transactionUtils.signTxInputs.mock.calls[0][1]).toBe(fakeStorage);
     expect(result).toEqual({ inputSignatures: [], ncCallerSignature: null });
   });
+
+  test('the resolver derives the spend chain COMPLIANTLY and the legacy chain NON-compliantly', async () => {
+    // signTxInputs is delegated to, but the per-chain key derivation lives in the resolver it gets
+    // as its 3rd arg -- the deriveChild/deriveNonCompliantChild logic that was previously buggy.
+    // The real signTxInputs invokes that resolver DURING signing (while the in-memory root is still
+    // alive, before the finally nulls it), so drive it the same way via mockImplementation.
+    const SPEND_LEAF = { chain: 'spend-leaf' };
+    const LEGACY_LEAF = { chain: 'legacy-leaf' };
+    const spendAcct = { deriveChild: jest.fn(() => SPEND_LEAF) };
+    const legacyAcct = { deriveNonCompliantChild: jest.fn(() => LEGACY_LEAF) };
+    const root = {
+      deriveChild: jest.fn(() => spendAcct),
+      deriveNonCompliantChild: jest.fn(() => legacyAcct),
+    };
+    walletUtils.getXPrivKeyFromSeed.mockReturnValue(root);
+
+    const derived = {};
+    transactionUtils.signTxInputs.mockImplementation(async (_tx, _storage, resolver) => {
+      derived.spend = await resolver('spend');
+      derived.legacy = await resolver('legacy');
+      return { inputSignatures: [], ncCallerSignature: null };
+    });
+
+    const signer = makePasskeyTxSigner();
+    await signer(fakeTx, fakeStorage);
+
+    // 'spend' (shielded) uses the COMPLIANT path: deriveChild(SHIELDED path) then deriveChild(0).
+    // Called exactly once => the shielded branch never touched the legacy (non-compliant) path.
+    expect(derived.spend).toBe(SPEND_LEAF);
+    expect(root.deriveChild).toHaveBeenCalledTimes(1);
+    expect(root.deriveChild).toHaveBeenCalledWith("m/44'/280'/2'");
+    expect(spendAcct.deriveChild).toHaveBeenCalledWith(0);
+
+    // 'legacy' (P2PKH) uses the NON-COMPLIANT path: deriveNonCompliantChild(P2PKH path) then (0).
+    expect(derived.legacy).toBe(LEGACY_LEAF);
+    expect(root.deriveNonCompliantChild).toHaveBeenCalledTimes(1);
+    expect(root.deriveNonCompliantChild).toHaveBeenCalledWith("m/44'/280'/0'");
+    expect(legacyAcct.deriveNonCompliantChild).toHaveBeenCalledWith(0);
+  });
 });
 
 describe('verifyPasskeyForUnlock', () => {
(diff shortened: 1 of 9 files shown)


src/passkey/passkeySigner.js
```diff
@@ -0,0 +1,183 @@
...
+
+      // Delegate the input-signing loop to wallet-lib so input selection, shielded-spend
+      // handling, the nano/OCB caller signature and encoding all live in one place and don't
+      // drift as the lib evolves. The resolver returns the change-path xpriv for each chain
+      // ('legacy' = m/44'/280'/0'/0, 'spend' = the shielded spend chain m/44'/280'/2'/0).
+      return await transactionUtils.signTxInputs(tx, storage, async (chain) => {
+        const acctPath = chain === 'spend'
+          ? hathorConstants.SHIELDED_SPEND_ACCT_PATH
+          : hathorConstants.P2PKH_ACCT_PATH;
+        return root.deriveNonCompliantChild(acctPath).deriveNonCompliantChild(0);
```
tuliomir: question: Shouldn't we use `deriveChild()` for non-legacy addresses?

pedroferreira1: Nice catch!! Fixed https://github.com/HathorNetwork/hathor-wallet-mobile/pull/896/changes/2ce7c1c72a35baa241a1e4cb2c52ce2a3fc665fb

raul-oliveira: @coderabbitai add this to your memory for future reviews

src/passkey/passkeySigner.js
```diff
@@ -0,0 +1,207 @@
...
+
+        // Wallets signed in before credentialId was captured (or on another device) learn it
+        // here, so the next ceremony can skip the picker too.
+        if (credentialId && credentialId !== meta.credentialId) {
+          STORE.updateWalletMeta({ credentialId });
+        }
+
+        // Derive the account root xpriv once; wallet-lib's signTxInputs derives the per-input
+        // keys from the chain xprivs it requests through the resolver below.
+        root = walletUtils.getXPrivKeyFromSeed(words, { networkName: NETWORK_MAINNET });
```
raul-oliveira, line 154: Why this networkName param is using a fixed value?

jestMockSetup.js
```diff
@@ -90,7 +90,7 @@ jest.mock('unleash-proxy-client', () => ({
     UPDATE: 'UPDATE',
     IMPRESSION: 'IMPRESSION',
   },
-}));
+}), { virtual: true });
```
raul-oliveira: why do we need it here?
virtual is supposed to be used when the module doesn't exist locally

src/passkey/passkeyService.js
```diff
@@ -0,0 +1,293 @@
...
+// never generated here.
+const randomBytes = (n) => {
+  const b = new Uint8Array(n);
+  // react-native-get-random-values makes global.crypto.getRandomValues available.
+  global.crypto.getRandomValues(b);
+  return b;
+};
+
+// react-native-passkey is a NATIVE dependency. Lazy-require it so the JS bundle still builds for
+// anyone who hasn't run `yarn && pod install` yet; the clear error only fires if passkey is used.
```
raul-oliveira: why should we keep this in consideration? 
Every new release of the app is based on a new entire fresh build, isnt?

raul-oliveira (dismissed):

tuliomir (dismissed):

src/passkey/passkeySigner.js
pedroferreira1: `networkName` only sets the version-byte prefix used when an xpriv/xpub is serialized to a string (mainnet vs testnet) — it doesn't affect the derived key material, since BIP32 derivation from the seed is network-independent. The signer uses root purely to derive child private keys and sign, so the signatures come out identical on any network.

jestMockSetup.js
pedroferreira1: Exactly right — virtual: true is for a module that isn't resolvable locally, which unleash-proxy-client is (not installed, not in package-lock, not imported anywhere in src/ — the app uses @hathor/unleash-client). Without it, jest.mock('unleash-proxy-client', …) throws Cannot find module at setup and kills the whole suite; that's what I was unblocking.
▎
▎ But you've spotted something better: this mock is just dead (leftover from before the @hathor/unleash-client migration). So instead of virtual-mocking it, I removed it — along with react-native-modal and react-native-animatable, which are dead in exactly the same way, plus their stale transformIgnorePatterns entries. Verified the suite still runs with zero Cannot find module and passkeySigner still 9/9. Pushed in the follow-up commit.

src/passkey/passkeyService.js
pedroferreira1: Good point, you're right that releases (and any post-pod install build) always have the dependency, so this never fires in production. It's purely a dev-onboarding safety net: a teammate who runs the app before yarn && cd ios && pod install gets a clear "run pod install" error only if they hit a passkey flow, instead of an obscure native failure.

 I reworded the comment to say that accurately (the old "so the JS bundle still builds" was imprecise, Metro bundles it either way; the real benefit is the clear error). I kept the lazy-require itself, since it's what lets getPasskey surface that clear, actionable error, which is exactly the behavior a previous review asked us to harden (preserve the original error as cause, assert the export). Happy to go all the way to a plain top-level import if you'd rather drop the defensive code, but I leaned toward keeping it to satisfy both reviews.

__tests__/passkey/passkeySigner.test.js
```diff
@@ -0,0 +1,225 @@
...
+  ...overrides,
+});
+
+// A fake account-root xpriv whose derivation chain never actually runs in these tests
+// (signTxInputs is mocked and does not invoke the resolver), but is shaped correctly just in case.
+const makeRoot = () => ({
+  deriveNonCompliantChild: jest.fn(() => ({
+    deriveNonCompliantChild: jest.fn(() => ({})),
+  })),
+});
```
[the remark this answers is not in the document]
pedroferreira1: Agreed — and the most valuable of the batch, since the spend/legacy derivation is the piece that had actually been buggy and had zero coverage. Added a test that drives the resolver the way the real signTxInputs does — via mockImplementation, invoking it while the in-memory root is still alive. (Capturing the resolver and calling it afterwards doesn't work: the signer's finally nulls root once signing resolves.) It asserts 'spend' → deriveChild(SHIELDED_SPEND_ACCT_PATH).deriveChild(0) (compliant) and 'legacy' → deriveNonCompliantChild(P2PKH_ACCT_PATH).deriveNonCompliantChild(0) (non-compliant), with toHaveBeenCalledTimes(1) on each root method so any cross-over fails. Also extended makeRoot() to stub deriveChild, as suggested.

src/passkey/passkeyService.js
```diff
@@ -0,0 +1,294 @@
...
+ * Use for first-time onboarding — this mints a NEW passkey (and thus a NEW wallet) every call.
+ *
+ * @param {string} userName label shown by the OS passkey UI
+ * @returns {Promise<{ words: string, label: string, credentialId: string|null }>}
+ */
+export async function createWalletWordsFromPasskey(userName = 'Hathor Wallet') {
+  const credentialId = await registerPasskey(userName); // enables PRF on the new passkey
+  const { prf } = await getPrfViaAssertion(credentialId); // Face ID again to evaluate PRF
+  return { ...wordsFromPrf(prf), label: userName, credentialId: credentialId ?? null };
+}
```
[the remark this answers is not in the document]
pedroferreira1: Correct that a cancelled/failed second assertion after registerPasskey leaves an orphaned platform credential, and retries accumulate more (the random user.id suffix is deliberate — it stops the authenticator from silently replacing another wallet's passkey when two share a name). As you note, WebAuthn has no client API to delete a credential, so there's no rollback available — the only real mitigation is a distinct "a passkey may have been created; remove it in Settings if you don't finish" message, and the two-prompt split itself is required (iOS returns the PRF secret only at assertion). We're accepting this as a known WebAuthn limitation at this stage rather than adding that UX surface; tracking for later. No code change here.

src/passkey/passkeySigner.js
```diff
@@ -0,0 +1,207 @@
...
+
+        if (!meta?.xpub || derivePasskeyXpub(words) !== meta.xpub) {
+          throw new PasskeyXpubMismatchError(meta?.passkeyLabel);
+        }
+
+        // Wallets signed in before credentialId was captured (or on another device) learn it
+        // here, so the next ceremony can skip the picker too.
+        if (credentialId && credentialId !== meta.credentialId) {
+          STORE.updateWalletMeta({ credentialId });
+        }
```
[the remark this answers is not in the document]
pedroferreira1: Valid concern, fixed — though the mechanism differs slightly from the description: STORE.updateWalletMeta doesn't actually return a promise today; it creates the setItem promise but never returns it, so await-ing the call would be a no-op. The floating promise is inside updateWalletMeta. So the fix is one level down: updateWalletMeta now returns its setItem promise, and both backfill sites attach a logging .catch(). I kept it fire-and-forget rather than await — the persisted credentialId is only an optimization to skip the OS picker next time, so it must never block or fail the ceremony; logging a rare storage failure is enough.

tuliomir (dismissed):

raul-oliveira (dismissed):

tuliomir (approved):

raul-oliveira (approved):

PR #897, opened 2026-07-23
feat: passkey wallet — onboarding UI & PIN-less integration

pedroferreira1:
### Summary

Wires the passkey core into the app — **stacked on `feat/passkey-core`** (review/merge that first). A passkey wallet replaces both the seed phrase and the PIN: create or sign in with a passkey, persist only the account xpub + label, and run every operation that used to need a PIN through a passkey ceremony (Face ID / fingerprint) that derives keys in memory, signs, and discards them.

What's here:
- **Onboarding & unlock UI** — the passkey onboarding button + a dedicated passkey lock screen; `InitWallet`/`App` mounting and routing (kill/relaunch → lock screen, not the PIN pad).
- **Wallet start & sagas** — xpub-only read-only start + external-signer registration; push-notification and wallet-service gating for passkey wallets.
- **PIN-less flows** — send / create-token / swap / reown pass **no** PIN to the lib; the Security screen hides "biometry instead of PIN" and "Change PIN" for passkey wallets.
- **i18n** — re-adds the `ttag` wrapping to `passkeySigner`'s error strings and adds the extracted translations (pt-br filled; da/ru-ru English fallback).

All behind the `passkey-onboarding.rollout` flag (off by default).

### Acceptance Criteria
- Create a passkey wallet: one tap + biometric lands on the dashboard — no ChoosePinScreen, no seed-words backup screen.
- Kill/relaunch → passkey lock screen → biometric unlock → balances load.
- Send, create token, and swap trigger a biometric ceremony instead of the PIN pad; **no PIN is passed to the lib**; the transaction confirms.
- Choosing the wrong passkey shows a clear error naming the expected wallet; no silent wallet switch.
- Reset + sign in with the same passkey restores the same wallet (same first address).
- Reown: `htr_sendTransaction` / `htr_sendNanoContractTx` work with one ceremony each; `htr_signWithAddress` / `htr_signOracleData` are cleanly rejected ("not yet supported") pending the external private-key follow-up; a cancelled ceremony is retryable.
- Security screen hides "biometry instead of PIN" and "Change PIN" for passkey wallets.
- A cancelled or failed send/create re-enables its button.
- Passkey error strings are translated (pt-br); no fuzzy entries; `msgfmt -c` passes.
- **Regression:** normal seed+PIN wallets are unaffected (PIN screen everywhere, biometry toggle present, push notifications available).
- Feature flag off by default — nothing appears unless enabled.

### Security Checklist

### DIFF
--- a/locale/pt-br/texts.po
+++ b/locale/pt-br/texts.po
@@ -73,7 +73,7 @@ msgstr "CRIAR TOKEN DE DEPÓSITO"
 msgid "CREATE FEE TOKEN"
 msgstr "CRIAR TOKEN DE TAXA"
 
-#: src/screens/CreateTokenAmount.js:36 src/screens/CreateTokenConfirm.js:87
+#: src/screens/CreateTokenAmount.js:36 src/screens/CreateTokenConfirm.js:88
 #: src/screens/CreateTokenDepositNotice.js:42
 #: src/screens/CreateTokenTypeNotice.js:94 src/utils.js:654
 msgid "CREATE TOKEN"
@@ -148,11 +148,11 @@ msgid "This app is developed by Hathor Labs and is distributed for free."
 msgstr ""
 "Este aplicativo foi desenvolvido pela Hathor Labs e é distribuído de graça."
 
-#: src/screens/About.js:99 src/screens/InitWallet.js:65
+#: src/screens/About.js:99 src/screens/InitWallet.js:66
 msgid "This wallet is connected to the **mainnet**."
 msgstr "Esta wallet está conectada à **mainnet**."
 
-#: src/screens/About.js:102 src/screens/InitWallet.js:68
+#: src/screens/About.js:102 src/screens/InitWallet.js:69
 msgid ""
 "A mobile wallet is not the safest place to store your tokens.\n"
 "So, we advise you to keep only a small amount of tokens here, such as pocket "
@@ -164,8 +164,8 @@ msgstr ""
 
 #: src/screens/About.js:107
 msgid ""
-"For further information, check out the |link1:Terms of Service| and |"
-"link2:Privacy Policy|, or our website |link3:https://hathor.network/|."
+"For further information, check out the |link1:Terms of Service| and |link2:"
+"Privacy Policy|, or our website |link3:https://hathor.network/|."
 msgstr ""
 "Para mais informações, veja os |link1:Termos de Serviço| e |link2:Política "
 "de Privacidade|, ou visite nosso site |link3:https://hathor.network/|."
@@ -332,6 +332,7 @@ msgstr "Você está prestes a adicionar esses tokens à sua carteira:"
 msgid "Import tokens"
 msgstr "Importar tokens"
 
+#: src/components/PasskeyOnboardingButton.js:176
 #: src/screens/ConfirmImportScreen.js:129
 #: src/screens/Reown/RegisterTokenAfterSuccess.js:80
 #: src/screens/Reown/SuccessFeedbackScreen.js:39
@@ -366,13 +367,13 @@ msgid "Amount of ${ name } (${ symbol })"
 msgstr "Quantidade de ${ name } (${ symbol })"
 
 #: src/screens/CreateTokenAmount.js:211 src/screens/CreateTokenName.js:67
-#: src/screens/CreateTokenSymbol.js:86 src/screens/InitWallet.js:229
-#: src/screens/InitWallet.js:382 src/screens/SendAddressInput.js:66
+#: src/screens/CreateTokenSymbol.js:86 src/screens/InitWallet.js:233
+#: src/screens/InitWallet.js:386 src/screens/SendAddressInput.js:66
 #: src/screens/SendAmountInput.js:169
 msgid "Next"
 msgstr "Próximo"
 
-#: src/screens/CreateTokenConfirm.js:31
+#: src/screens/CreateTokenConfirm.js:32
 msgid ""
 "You chose to create a **Deposit-Based Token**, which requires a 1% HTR "
 "deposit."
@@ -380,7 +381,7 @@ msgstr ""
 "Você escolheu criar um **Token Baseado em Depósito**, que requer um depósito "
 "de 1% em HTR."
 
-#: src/screens/CreateTokenConfirm.js:33
+#: src/screens/CreateTokenConfirm.js:34
 msgid ""
 "You chose to create a **Fee-Based Token**, so a small fee will be applied to "
 "each future transaction of this token."
@@ -388,55 +389,55 @@ msgstr ""
 "Você escolheu criar um **Token com Taxa**, então uma pequena taxa será "
 "aplicada a cada transação futura deste token."
 
-#: src/screens/CreateTokenConfirm.js:145
+#: src/screens/CreateTokenConfirm.js:146
 msgid "Creating token"
 msgstr "Criando um token"
 
-#: src/screens/CreateTokenConfirm.js:151
+#: src/screens/CreateTokenConfirm.js:152
 msgid "Building the transaction"
 msgstr "Criando a transação"
 
-#: src/screens/CreateTokenConfirm.js:166
+#: src/screens/CreateTokenConfirm.js:174
 msgid "Enter your 6-digit pin to create your token"
 msgstr "Digite seu PIN de 6 dígitos para criar o seu token"
 
-#: src/screens/CreateTokenConfirm.js:167
+#: src/screens/CreateTokenConfirm.js:175
 msgid "Authorize token creation"
 msgstr "Autorizar criação de token"
 
-#: src/screens/CreateTokenConfirm.js:169 src/screens/SendConfirmScreen.js:196
-#: src/screens/TokenSwapReview.js:221
+#: src/screens/CreateTokenConfirm.js:177 src/screens/SendConfirmScreen.js:204
+#: src/screens/TokenSwapReview.js:237
 msgid "Building transaction"
 msgstr "Criando transação"
 
-#: src/screens/CreateTokenConfirm.js:210
+#: src/screens/CreateTokenConfirm.js:218
 #, javascript-format
 msgid "**${ name }** created successfully"
 msgstr "**${ name }** criado com sucesso"
 
-#: src/screens/CreateTokenConfirm.js:222
+#: src/screens/CreateTokenConfirm.js:235
 msgid "Amount of ${ name }"
 msgstr "Quantidade de ${ name }"
 
-#: src/screens/CreateTokenConfirm.js:236
+#: src/screens/CreateTokenConfirm.js:249
 msgid "Token name"
 msgstr "Nome do token"
 
-#: src/screens/CreateTokenConfirm.js:242 src/screens/CreateTokenSymbol.js:71
+#: src/screens/CreateTokenConfirm.js:255 src/screens/CreateTokenSymbol.js:71
 msgid "Token symbol"
 msgstr "Símbolo do token"
 
 #: src/components/Reown/CreateTokenRequest.js:87
 #: src/components/Reown/CreateTokenRequest.js:129
-#: src/screens/CreateTokenConfirm.js:249
+#: src/screens/CreateTokenConfirm.js:262
 msgid "Deposit"
 msgstr "Depósito"
 
-#: src/screens/CreateTokenConfirm.js:257
+#: src/screens/CreateTokenConfirm.js:270
 msgid "Network fee"
 msgstr "Taxa de Rede"
 
-#: src/screens/CreateTokenConfirm.js:266
+#: src/screens/CreateTokenConfirm.js:279
 msgid "Create token"
 msgstr "Criar um token"
 
@@ -553,8 +554,8 @@ msgid ""
 "Once selected, the token type cannot be changed later. |link:Learn more "
 "about deposits and fees here|"
 msgstr ""
-"Uma vez selecionado, o tipo de token não pode ser alterado depois. |"
-"link:Saiba mais sobre depósitos e taxas aqui|"
+"Uma vez selecionado, o tipo de token não pode ser alterado depois. |link:"
+"Saiba mais sobre depósitos e taxas aqui|"
 
 #: src/screens/CreateTokenTypeNotice.js:133
 msgid "DEPOSIT TOKEN"
@@ -614,18 +615,18 @@ msgstr "Tokens encontrados"
 msgid "Continue"
 msgstr "Continuar"
 
-#: src/screens/InitWallet.js:62
+#: src/screens/InitWallet.js:63
 msgid "Welcome to Hathor Wallet!"
 msgstr "Bem vindo à Hathor Wallet!"
 
-#: src/screens/InitWallet.js:73
+#: src/screens/InitWallet.js:74
 msgid ""
-"For further information, check out our website |link:https://"
-"hathor.network/|."
+"For further information, check out our website |link:https://hathor."
+"network/|."
 msgstr ""
 "Para mais informações, visite nosso site |link:https://hathor.network/|."
 
-#: src/screens/InitWallet.js:86
+#: src/screens/InitWallet.js:87
 msgid ""
 "I agree with the |link1:Terms of Service| and |link2:Privacy Policy| and "
 "understand the risks of using a mobile wallet"
@@ -633,19 +634,19 @@ msgstr ""
 "Eu concordo com os |link1:Termos de Serviço| e |link2:Política de "
 "Privacidade| e entendo os riscos de usar uma wallet de celular"
 
-#: src/screens/InitWallet.js:98
+#: src/screens/InitWallet.js:99
 msgid "Start"
 msgstr "Iniciar"
 
-#: src/screens/InitWallet.js:115
+#: src/screens/InitWallet.js:116
 msgid "To start,"
 msgstr "Para começar,"
 
-#: src/screens/InitWallet.js:117
+#: src/screens/InitWallet.js:118
 msgid "You need to **initialize your wallet**."
 msgstr "Você precisa **inicializar sua wallet**."
 
-#: src/screens/InitWallet.js:120
+#: src/screens/InitWallet.js:121
 msgid ""
 "You can either **start a new wallet** or **import a wallet** that already "
 "exists."
@@ -653,23 +654,23 @@ msgstr ""
 "Você pode tanto **iniciar uma nova wallet** ou **importar uma wallet** que "
 "já existe."
 
-#: src/screens/InitWallet.js:123
+#: src/screens/InitWallet.js:124
 msgid "To import a wallet, you will need to provide your seed words."
 msgstr "Para importar uma wallet, você deve entrar com as suas palavras."
 
-#: src/screens/InitWallet.js:128
-msgid "Import Wallet"
-msgstr "Importar Wallet"
-
-#: src/screens/InitWallet.js:134
+#: src/screens/InitWallet.js:129
 msgid "New Wallet"
 msgstr "Nova Wallet"
 
-#: src/screens/InitWallet.js:220
+#: src/screens/InitWallet.js:134
+msgid "Import Wallet"
+msgstr "Importar Wallet"
+
+#: src/screens/InitWallet.js:224
 msgid "Your wallet has been created!"
 msgstr "Sua wallet foi criada!"
 
-#: src/screens/InitWallet.js:222
+#: src/screens/InitWallet.js:226
 msgid ""
 "You must **do a backup** and save the words below **in the same order they "
 "appear**."
@@ -677,11 +678,11 @@ msgstr ""
 "Você deve **realizar o backup** e salvar as palavras abaixo **na mesma ordem "
 "em que elas aparecem**."
 
-#: src/screens/InitWallet.js:343
+#: src/screens/InitWallet.js:347
 msgid "To import a wallet,"
 msgstr "Para importar sua wallet,"
 
-#: src/screens/InitWallet.js:345
+#: src/screens/InitWallet.js:349
 #, javascript-format
 msgid ""
 "You need to **write down the ${ this.numberOfWords } seed words** of your "
@@ -690,11 +691,11 @@ msgstr ""
 "Você precisa **digitar as ${ this.numberOfWords } palavras** da sua wallet, "
 "separadas por espaço."
 
-#: src/screens/InitWallet.js:348
+#: src/screens/InitWallet.js:352
 msgid "Words"
 msgstr "Palavras"
 
-#: src/screens/InitWallet.js:353
+#: src/screens/InitWallet.js:357
 msgid "Enter your seed words separated by space"
 msgstr "Digite suas palavras separadas por espaços"
 
@@ -727,8 +728,8 @@ msgstr "**${ loadedData.addresses } endereços** encontrados"
 msgid "There's been an error connecting to the server."
 msgstr "Ocorreu um erro ao conectar com o servidor."
 
-#: src/screens/LoadWalletErrorScreen.js:28 src/screens/PinScreen.js:318
-#: src/screens/Settings.js:177
+#: src/screens/LoadWalletErrorScreen.js:28 src/screens/PasskeyLockScreen.js:139
+#: src/screens/PinScreen.js:318 src/screens/Settings.js:177
 msgid "Reset wallet"
 msgstr "Resetar Wallet"
 
@@ -783,6 +784,23 @@ msgstr "Registrar Nano Contract"
 #: src/screens/NanoContractRegisterQrCodeScreen.js:62
 msgid "Scan the nano contract ID QR code"
 msgstr "Escaneie o QR code do ID do Nano Contract"
+
+#: src/screens/PasskeyLockScreen.js:61
+msgid "Unlock cancelled."
+msgstr "Desbloqueamento cancelado."
+
+#: src/screens/PasskeyLockScreen.js:103
+#, javascript-format
+msgid "Unlock with the passkey \"${ passkeyLabel }\""
+msgstr "Desbloquear com passkey \"${ passkeyLabel }\""
+
+#: src/screens/PasskeyLockScreen.js:104
+msgid "Unlock with your passkey"
+msgstr "Desbloquear com sua passkey"
+
+#: src/screens/PasskeyLockScreen.js:124
+msgid "Unlock"
+msgstr "Desbloquear"
 
 #: src/screens/PinScreen.js:272
 msgid "Incorrect PIN Code. Try again."
@@ -975,15 +993,15 @@ msgstr "Nenhuma biometria suportada"
 msgid "Use ${ this.supportedBiometry }"
 msgstr "Usar ${ this.supportedBiometry }"
 
-#: src/screens/Security.js:191
+#: src/screens/Security.js:195
 msgid "SECURITY"
 msgstr "SEGURANÇA"
 
-#: src/screens/Security.js:213
+#: src/screens/Security.js:221
 msgid "Change PIN"
 msgstr "Mudar PIN"
 
-#: src/screens/Security.js:218
+#: src/screens/Security.js:226
 msgid "Lock wallet"
 msgstr "Bloquear a wallet"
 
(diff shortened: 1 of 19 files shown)


src/screens/SendConfirmScreen.js
```diff
@@ -188,6 +189,13 @@ const SendConfirmScreen = () => {
     // Disable the button before opening the PinScreen so it can't be tapped
     // again while we return from it and build the feedback modal.
     setIsSending(true);
+    if (STORE.isPasskeyWallet()) {
```
raul-oliveira: nit: this passkey shortcut — the `if (STORE.isPasskeyWallet()) { ... execute...(); return; }` guard plus its 3-line comment — is duplicated verbatim here, in `CreateTokenConfirm.js:165` and `TokenSwapReview.js:223` (and the `setIsSending(false)` re-enable + comment is duplicated again in the two `onDismissError`s). Could extract a small shared helper/hook (e.g. `authorizeTx({ isPasskey, execute, navigation, pinParams })`) so the PIN-less contract and its rationale live in one place. Non-blocking.

src/screens/PasskeyLockScreen.js
```diff
@@ -0,0 +1,154 @@
...
+        dispatch(startWalletRequested({ walletType: 'passkey' }));
+      }
+      dispatch(unlockScreen());
+    } catch (e) {
+      setError(e instanceof PasskeyCancelledError ? t`Unlock cancelled.` : e.message);
+      setVerifying(false);
+    }
+  };
+
+  useEffect(() => {
```
raul-oliveira, line 84: nit: Component Anatomy — this `useEffect` is declared after the `unlock` handler and the `passkeyLabel` derivation; the convention groups lifecycles right after state. Moving the effect up (optionally wrapping `unlock` in `useCallback`) keeps the state → effects → behavior → render ordering used elsewhere.

raul-oliveira (approved):
Approving — the integration diff itself looks solid and internally consistent. Two non-blocking nits inline.

src/screens/PasskeyLockScreen.js
pedroferreira1: Fair point on the convention, but I've left this one as-is. Moving the effect above unlock (a const it calls) trips no-use-before-define, and the clean workaround - wrapping unlock in useCallback so it can be referenced first - actually produces state -> behavior -> effect, not the state -> effect -> behavior the convention describes, while adding a hook and a deps array to maintain for no functional gain. The current order (state -> derived passkeyLabel -> unlock -> mount effect) reads cleanly, so I'd rather not add the indirection. Happy to revisit if you feel strongly about matching the house ordering exactly

src/screens/SendConfirmScreen.js
pedroferreira1: Makes total sense, done!! https://github.com/HathorNetwork/hathor-wallet-mobile/pull/897/commits/624b6a3f21c0e8de317e6fc1e7194827acc4d72e

src/sagas/wallet.js
```diff
@@ -291,6 +309,11 @@ export function* startWallet(action) {
         },
     };
     wallet = new HathorWallet(walletConfig);
+    if (isPasskeyWallet) {
+      // Must happen BEFORE start(): flips isSignedExternally so the read-only guards allow
+      // sending, with signatures produced by the passkey ceremony instead of a stored key.
+      wallet.setExternalTxSigningMethod(makePasskeyTxSigner());
```
tuliomir, line 323: issue(blocking): the pinned wallet-lib has none of the APIs this line switches on

Pinned here for delivery — the actual concern is `package.json:29`, which pins `@hathor/wallet-lib: 3.1.1` and is not touched by this PR.

I unpacked the published 3.1.1 tarball: `transactionUtils.signTxInputs` — what `makePasskeyTxSigner` delegates to — exists nowhere in the package. Separately, every signing entry point in 3.1.1 rejects a missing PIN unconditionally, with no external-signer escape: `signTx`, `prepareCreateNewToken`, `sendManyOutputsSendTransaction`, `SendTransaction.signTx`. `isSignedExternally` short-circuits `isReadonly()` only, never the PIN check.

So on a clean `npm install` every AC involving a transaction fails: Send shows "Pin code is required to sign a transaction", Create Token shows "Pin is required.", and no biometric sheet ever appears.

To be clear about where the fault is: the app code is right. I checked the lib branches, and `feat/pin-optional-external-signer` implements exactly this relaxation at exactly these call sites — it just is not released. Land the `@hathor/wallet-lib` bump with this PR rather than as a follow-up; the dependency contract belongs with the code that needs it.

src/screens/TokenSwapReview.js
```diff
@@ -213,14 +220,22 @@ const TokenSwapReview = () => {
...
   const onSwapButtonPress = () => {
-    const pinParams = {
-      cb: executeSend,
-      canCancel: true,
-      screenText: t`Enter your 6-digit pin to authorize operation`,
-      biometryText: t`Authorize operation`,
-      biometryLoadingText: t`Building transaction`,
-    };
-    navigation.navigate('PinScreen', pinParams);
+    authorizeTransaction({
```
tuliomir, line 237: issue(blocking): the swap path is unreachable for a passkey wallet, even after the lib bump

Pinned here for delivery — the actual concern is `TokenSwapReview.js:97` (outside the diff), the `wallet.createNanoContractTransaction` call at screen mount.

That method guards on `await this.storage.isReadonly()` — the raw storage flag — instead of the wallet-level `this.isReadonly()` that short-circuits on `isSignedExternally`. `generateAccessDataFromXpub` sets `WALLET_FLAGS.READONLY` unconditionally, so the guard is always true for a passkey wallet. I confirmed this is still the case in the newest published 4.1.0, so the bump in my other comment does not fix it.

The call runs with `signTx: false` well before any authorization, so the screen dies in `PHASE.ERROR` and this whole `authorizeTransaction` block plus the new `isSending` state never execute. The reown `htr_sendNanoContractTx` / `htr_createNanoContractCreateTokenTx` paths fail identically, which also makes part of the new `PinConfirmationPrompt` branch dead.

`feat/pin-optional-external-signer` already converts both lib call sites to `this.isReadonly()`. This PR needs the release that carries it, or the swap AC should come out until it lands.

src/sagas/reown.js
```diff
@@ -652,6 +678,18 @@ export function* processRequest(action) {
   } catch (e) {
     log.error('Error in processRequest:', e);
 
+    // A cancelled passkey ceremony is "not now", not an error (the rpc-handler re-wraps our
+    // typed error, so we use a consumable flag — same pattern as pinWasCancelled). Retry the
+    // request: the dapp consent modal shows again and the user can accept or reject.
+    if (consumePasskeySigningCancelled()) {
```
tuliomir, line 694: issue(blocking): a stale `signingCancelled` flag silently hijacks unrelated failures here

`signingCancelled` (`passkeySigner.js:83`) is module-global. It is set by any cancelled signing ceremony and cleared only by the start of the *next* ceremony, or by `consumePasskeySigningCancelled()` — whose sole caller is this line.

So an in-app cancel on SendConfirmScreen / CreateTokenConfirm / TokenSwapReview leaves it `true` indefinitely. The next dapp request that fails for an unrelated reason before any ceremony runs — insufficient funds, invalid params, `PromptRejectedError`, network — reads that stale `true` right here, before the `switch (e.constructor)` below. The saga retries instead of reporting: the user's explicit reject is ignored, the consent modal reappears for no visible reason, and the `InsufficientFundsModal` / `REQUEST_ERROR` modal that should have explained the failure never renders.

Reset `signingCancelled = false` at the top of `processRequest`, before `handleRpcRequest`, so the flag can only ever reflect a ceremony from this attempt.

src/sagas/reown.js
```diff
@@ -652,6 +678,18 @@ export function* processRequest(action) {
...
+      // Cancel the poll fork from this attempt before recursing — the early return below skips the
+      // cleanup at the end of processRequest, so without this the fork leaks (and a new one is
+      // forked per retry). Mirrors the SendNanoContractTxError retry path.
+      yield cancel(pendingPollTask);
+      const result = yield* processRequest(action);
```
tuliomir, line 709: issue(blocking): the retry lands on a screen the user cannot act on

This breaks the "a cancelled ceremony is retryable" AC rather than fulfilling it.

The rpc-handler fires its `*LoadingTrigger` *before* signing (`sendTransaction.js:141` then `:146`), so by the time the passkey ceremony runs, `setSendTxStatusLoading()` has already dispatched via `reown.js:1070`. This branch returns before the `switch` below, so unlike every sibling retry path — `setNewNanoContractStatusFailure()` before the recursion at `:725`, `setSendTxStatusFailure()` at `:794` — it resets nothing.

`SendTransactionRequest.js:485` then renders the full-screen "Sending transaction / Please wait." spinner and hides its own main content at `:497` for as long as the status is LOADING. The recursion pushes a fresh consent prompt onto a screen stuck on a spinner that will never resolve, so the user can neither see nor accept the retry.

That is the concrete difference from the `pinWasCancelled` pattern this mirrors: a PIN cancel happens during the prompt phase, before any loading status exists. Dispatch the matching `*StatusFailure` for `params.request.method` before recursing.

src/components/PasskeyOnboardingButton.js
```diff
@@ -0,0 +1,225 @@
...
+        passkeyLabel: label || 'Passkey wallet',
+        credentialId: credentialId ?? null,
+      });
+      setModalVisible(false);
+      // Same sequence ChoosePinScreen uses: unlock before entering the main stack,
+      // then request the wallet start (the saga re-reads the persisted walletMeta).
+      dispatch(unlockScreen());
+      dispatch(startWalletRequested({ walletType: 'passkey' }));
+      NavigationService.resetToMain();
+    } catch (e) {
```
tuliomir, line 94: issue(blocking): a failed write here strands the user on a PIN screen for a PIN-less wallet

`STORE.initPasskeyStorage` writes in two awaited steps — `storage.saveAccessData(accessData)` then `setItem(WALLET_META_KEY, ...)` (`store.js:362-366`) — and this catch performs no rollback.

If the second write rejects, `walletIsLoaded()` returns true (access data exists) but `isPasskeyWallet()` returns false (no meta), so `App.js:754` renders PinScreen instead of PasskeyLockScreen. The app is permanently stuck asking for a PIN that was never set, on access data holding no encrypted words. The only escape is "Reset wallet" — while the alert the user just saw said "Passkey unavailable", implying nothing was created.

Worth noting the timing: `setItem` populates the memory cache before awaiting AsyncStorage, so the current session behaves normally and the breakage only appears on the next cold start.

Make `initPasskeyStorage` atomic — write the meta first, or clear the access data on failure — and give this case its own message.

src/screens/TokenSwapReview.js
```diff
@@ -183,6 +189,7 @@ const TokenSwapReview = () => {
   // listener releases reserved UTXOs along the way.
   const exitOnError = () => {
     setModal(null);
+    setIsSending(false);
```
tuliomir, line 194: issue: cancelling the ceremony on this screen silently destroys the reviewed transaction

Pinned here for delivery — the actual concern is `executeSend`'s catch at `TokenSwapReview.js:216-219` (outside the diff), which calls this `exitOnError`.

The catch is pre-existing, but the PR routes brand-new *routine* failures into it. On the PIN path a cancellation never reached this code — PinScreen simply never invoked `cb`. On the passkey path, `PasskeyCancelledError`, `PasskeyBusyError` (double-tap) and `PasskeyXpubMismatchError` all land here.

So the user dismisses the Face ID sheet and is popped off the screen with no message at all — UTXOs released, quote refetched, only a `console.error` that goes nowhere in a release build. Indistinguishable from a crash, and contrary to the AC that a cancelled flow re-enables its button.

Special-case the cancel and busy errors: reset `isSending` and stay put. Route everything else through the feedback modal the way SendConfirmScreen does, plus `onExceptionCaptured`.

src/sagas/reown.js
```diff
@@ -599,6 +601,30 @@ export function* processRequest(action) {
...
   };
 
+  // Message/oracle signing calls wallet.signMessageWithAddress, which needs the stored
+  // private key and bypasses the external signer — impossible for xpub-only passkey
+  // wallets. Reject cleanly instead of crashing inside the rpc-handler.
+  const passkeyUnsupportedMethods = [
+    AVAILABLE_METHODS.HATHOR_SIGN_MESSAGE,
+    AVAILABLE_METHODS.HATHOR_SIGN_ORACLE_DATA,
+  ];
+  if (STORE.isPasskeyWallet() && passkeyUnsupportedMethods.includes(params.request.method)) {
```
tuliomir, line 618: question(non-blocking): should the session proposal filter these two as well?

The per-request rejection here is exactly what the description promises, so this is a scope question rather than a defect.

It is about the layer above: `availableMethods` at `:1401` is the unfiltered `values(AVAILABLE_METHODS)`, so the mandatory-method check at `:1404` passes for a dapp that declares `htr_signWithAddress` as **mandatory**. It connects successfully and only breaks at first use, after the user has granted the session.

Filtering at proposal time would surface the incompatibility at connect — but it also makes such a dapp unconnectable for passkey users rather than partially usable. That reads like a product call; worth stating which behavior you want while the external private-key follow-up is pending.

src/sagas/reown.js
```diff
@@ -599,6 +601,30 @@ export function* processRequest(action) {
...
+    log.debug(`Rejecting ${params.request.method}: not supported for passkey wallets.`);
+    yield call(() => walletKit.respondSessionRequest({
+      topic: payload.topic,
+      response: {
+        id: payload.id,
+        jsonrpc: '2.0',
+        error: {
+          code: ERROR_CODES.USER_REJECTED_METHOD,
```
tuliomir: suggestion(non-blocking): this code tells the dapp the user declined

`USER_REJECTED_METHOD` (5002) means the user declined. They didn't — the wallet structurally cannot comply, and that stays true no matter how many times they retry.

The practical difference: the dapp renders "user rejected" and typically offers a retry that can never succeed, whereas an unsupported-method code lets it disable the action outright.

`UNAUTHORIZED_METHODS` (3001), or a `-32601`-style unsupported code, would make the deferral legible on the dapp side. Cheap change, and it does not affect the deliberate "reject for now" behavior itself.

src/passkey/authorizeTransaction.js
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @param {object} args
+ * @param {() => void} args.execute - The passkey-path action, run directly (no PIN) for passkey
+ *   wallets. NOT called for PIN wallets — those authorize through pinParams.cb instead. It is
+ *   often the same function as pinParams.cb, but need not be (a caller may wrap it).
+ * @param {{ navigate: Function }} args.navigation - Navigation used to open the PinScreen.
+ * @param {object} args.pinParams - Params forwarded to the PinScreen (cb, screenText, ...); cb is
+ *   invoked with the entered PIN. Ignored for passkey wallets.
+ */
+export function authorizeTransaction({ execute, navigation, pinParams }) {
```
tuliomir: suggestion: the duplicated callback has already diverged at one of the three call sites

At `SendConfirmScreen.js:192` and `CreateTokenConfirm.js:165`, `execute` and `pinParams.cb` are literally the same identifier. At `TokenSwapReview.js:226` they are deliberately different — `execute` wraps with `setIsSending(true)`, `cb` does not.

That is exactly the drift the duplication invites: a future edit to the send path's `cb` will silently not apply to the passkey path, and nothing catches it.

Collapsing to one required callback with optional wrapping would make the shared case impossible to diverge by accident:

```js
export function authorizeTransaction({ navigation, cb, onPasskey = cb, ...pinScreenParams }) {
  if (STORE.isPasskeyWallet()) {
    onPasskey();
    return;
  }
  navigation.navigate('PinScreen', { cb, ...pinScreenParams });
}
```

That also lets the JSDoc drop the "typically the same function as `pinParams.cb`" paragraph.

src/sagas/wallet.js
```diff
@@ -187,6 +188,18 @@ export function* startWallet(action) {
...
+  // Passkey (xpub-only) wallets carry no seed and no PIN: the saga re-reads the persisted
+  // metadata instead of trusting action payloads with sensitive data.
+  const walletMeta = STORE.getWalletMeta();
+  const isPasskeyWallet = walletMeta?.walletType === 'passkey';
+
+  // A passkey wallet boots read-only from this xpub; if it is missing/corrupted, fail fast with a
+  // clear message instead of handing `{ xpub: undefined }` to HathorWallet and surfacing an opaque
+  // wallet-lib error later. (makePasskeyTxSigner guards the same case at sign time.)
+  if (isPasskeyWallet && !walletMeta?.xpub) {
+    throw new Error('Passkey wallet metadata is missing its xpub.');
```
tuliomir: suggestion: this throw is unreported and surfaces to the user as a network error

`errorHandler` (`sagas/helpers.js:115`) does `console.error` + `put(startWalletFailed())` and — unlike the sibling catch at `wallet.js:380` — never calls `onExceptionCaptured`. For a wallet whose only persisted identity is that xpub, this is the most damaging corruption possible, and nothing reaches Sentry.

What the user sees is worse: `LoadWalletErrorScreen` renders the hardcoded "There's been an error connecting to the server.", blaming the network for lost data. "Try again" then routes to PasskeyLockScreen, whose `verifyPasskeyForUnlock` folds "no stored xpub" into `PasskeyXpubMismatchError` at `passkeySigner.js:203` and tells them "This passkey opens a different wallet." They will hunt for a nonexistent second passkey instead of doing the one thing that works — reset, then sign in with the same passkey.

Report this via `onExceptionCaptured`, and split "metadata missing" from "xpub mismatch" at `passkeySigner.js:141`/`:203`. The comment three lines up saying the signer "guards the same case" is what hides the difference — the two guards produce different diagnoses.

src/screens/PasskeyLockScreen.js
```diff
@@ -0,0 +1,154 @@
...
+    try {
+      await verifyPasskeyForUnlock();
+      if (!wallet) {
+        // App boot (or wallet stopped): start the xpub-only wallet. The saga re-reads the
+        // persisted walletMeta; no secret travels through redux.
+        dispatch(startWalletRequested({ walletType: 'passkey' }));
+      }
+      dispatch(unlockScreen());
+    } catch (e) {
+      setError(e instanceof PasskeyCancelledError ? t`Unlock cancelled.` : e.message);
```
tuliomir: suggestion: this catch has no logging and can render an empty error

The file imports no logger at all. Compare `PinScreen.js:162`, whose lock-screen catch does both `log.error(e)` and `onExceptionCaptured(e, true)`. A missing native module, a `react-native-passkey` shape regression, or an AsyncStorage failure at unlock leaves no diagnostic anywhere — on the app's only entry point for these wallets.

`e.message` is also unguarded: a non-`Error` rejection or an empty native message yields `setError(undefined)`, the `{error && ...}` guard renders nothing, and the user stares at an Unlock button that appears to do nothing when tapped.

The `[BadConfiguration]`-style `.error` code that `PasskeyOnboardingButton.js:97` deliberately surfaces is dropped here too, so the same misconfiguration is actionable during onboarding and opaque at unlock. Add `log.error`, dispatch `onExceptionCaptured` for non-cancel errors, and fall back to a translated generic message with the `e?.error` prefix.

src/components/PasskeyOnboardingButton.js
```diff
@@ -0,0 +1,225 @@
...
+      // react-native-passkey throws a PasskeyError with a `.error` code — surface it so a
+      // generic "unknown error" becomes actionable (e.g. BadConfiguration = domain not verified).
+      console.warn('[passkey] failed:', e?.error, e?.message, e);
+      const code = e?.error ? `[${e.error}] ` : '';
+      Alert.alert(t`Passkey unavailable`, `${code}${e?.message ?? String(e)}`);
```
tuliomir, line 104: suggestion: a deliberate cancellation is reported as "Passkey unavailable"

Dismissing the OS passkey sheet lands in this catch and raises an alarming modal claiming the feature is broken, when the user simply backed out.

The right predicate already exists and is used correctly two files over: `isPasskeyCancel` (`passkeyService.js:95`, used at `passkeySigner.js:134` and `:197`). It is just not applied here.

Two smaller things in the same block: `console.warn` instead of `logger('passkey')`/Sentry, inconsistent with everything else in `src/passkey/`; and the `try` also covers `unlockScreen()` / `startWalletRequested()` / `resetToMain()` above, which run *after* the wallet is persisted — so a navigation failure there tells the user the passkey was unavailable for a wallet that was in fact created.

src/sagas/wallet.js
```diff
@@ -291,6 +309,11 @@ export function* startWallet(action) {
         },
     };
     wallet = new HathorWallet(walletConfig);
+    if (isPasskeyWallet) {
+      // Must happen BEFORE start(): flips isSignedExternally so the read-only guards allow
```
tuliomir: suggestion: "Must happen BEFORE start()" is an invariant the lib does not enforce

`setExternalTxSigningMethod` assigns `isSignedExternally` synchronously and `isReadonly()` re-evaluates it per call. `start()` never reads it, and `clearSensitiveData()` never clears `storage.txSignFunc` — so calling the setter after `start()` would behave identically today.

The real constraint is "before the first send". A hard "Must" the library doesn't impose either gets cargo-culted into unrelated code, or gets disproved and discredits the surrounding block — which is otherwise accurate and worth keeping.

Same overreach in the base branch's header at `passkeySigner.js:17`; worth fixing in the same pass.

src/components/PasskeyOnboardingButton.js
```diff
@@ -0,0 +1,225 @@
...
+  };
+
+  const onCreate = () => {
+    const name = walletName.trim() || 'Hathor Wallet';
+    run(() => createWalletWordsFromPasskey(name));
+  };
+
+  // A passkey wallet has no seed phrase and no PIN — its recovery is the OS syncing the passkey to
+  // the user's password manager. Name it per platform so the disclosure below is concrete.
+  const passwordManager = Platform.OS === 'ios' ? 'iCloud Keychain' : 'Google Password Manager';
```
tuliomir: suggestion: this backup promise is not something the app can verify

`passwordManager` is chosen purely from `Platform.OS`, but `passkeyService.js:212` requests `residentKey: 'required'` — which yields a *discoverable* credential, not a *synced* one.

The credential may live in 1Password, Dashlane, Samsung Pass, or a hardware key. On Android with a non-Google provider, the user-facing string below names the wrong manager outright.

Since the PRF secret *is* the wallet, "keep it backed up by syncing it to X" is a recovery promise with real money behind it. Naming the provider generically ("your password manager") would be honest, with a note in the comment that the provider is not detectable.

src/sagas/pushNotification.js
```diff
@@ -377,6 +377,16 @@ export function* setAvailablePushNotification(action) {
  * load the wallet service, otherwise it will use the wallet already loaded in the redux store.
  */
 export function* loadWallet() {
+  // Passkey (xpub-only) wallets have no PIN and no stored seed words, which this flow
+  // requires. Push notifications are already gated off for them at startWallet; this guard
+  // only protects against unexpected entry paths.
+  if (STORE.isPasskeyWallet()) {
+    yield put(pushLoadWalletFailed({
```
tuliomir, line 388: suggestion(non-blocking): this `Error` is constructed and then discarded

The only consumer of `PUSH_WALLET_LOAD_FAILED` is the `race` at `:490`, which treats it purely as a signal and never reads `payload.error`. There is no reducer case for it.

So if this "unexpected entry path" ever fires, the user gets the generic `pushRegisterFailed()` state with no explanation, and there is no log line to say why.

The guard itself is correct defense-in-depth and worth keeping — just add a `log.error` alongside it so it leaves a trace when it does fire.

src/screens/InitWallet.js
```diff
@@ -123,16 +124,19 @@ class InitialScreen extends React.Component {
             {t`To import a wallet, you will need to provide your seed words.`}
           </Text>
           <View style={this.style.buttonView}>
+            <NewHathorButton
```
tuliomir: question(non-blocking): is the New Wallet / Import Wallet reorder in scope?

The PR swaps the two pre-existing buttons — New Wallet was last and primary, Import Wallet first and secondary; now reversed — and restyles them. Appending `<PasskeyOnboardingButton />` below the existing pair would not have required touching either.

This changes the default onboarding affordance for every user, including with the feature flag off, and it is not mentioned in the description or the ACs.

Intentional? If so it deserves its own line in the PR body, since it ships regardless of the rollout flag. If not, worth splitting out.

src/components/PasskeyOnboardingButton.js
```diff
@@ -0,0 +1,225 @@
...
+      const xpub = derivePasskeyXpub(words);
+      await STORE.initPasskeyStorage(xpub, {
+        passkeyLabel: label || 'Passkey wallet',
+        credentialId: credentialId ?? null,
+      });
+      setModalVisible(false);
+      // Same sequence ChoosePinScreen uses: unlock before entering the main stack,
+      // then request the wallet start (the saga re-reads the persisted walletMeta).
+      dispatch(unlockScreen());
+      dispatch(startWalletRequested({ walletType: 'passkey' }));
```
tuliomir: nitpick: `walletType: 'passkey'` is never read

`startWallet` destructures only `{ words, pin }` from the payload and derives `isPasskeyWallet` from `STORE.getWalletMeta()` — deliberately, per the comment at `sagas/wallet.js:191`.

The payload field therefore reads as if it drives the branch when it does not. Drop it (same at `PasskeyLockScreen.js:57`), or label it diagnostic-only.

src/sagas/reown.js
```diff
@@ -599,6 +601,30 @@ export function* processRequest(action) {
     chain: get(requestSession.namespaces, 'hathor.chains[0]', ''),
   };
 
+  // Message/oracle signing calls wallet.signMessageWithAddress, which needs the stored
```
tuliomir: nitpick: this names the wrong lib function for one of the two methods

`htr_signWithAddress` does call `wallet.signMessageWithAddress`, but `htr_signOracleData` goes through `nanoUtils.getOracleSignedDataFromUser` -> `getOracleInputData` and never touches it.

The conclusion holds — both need a private key the external signer cannot provide — but the function name will mislead anyone who greps for it. Something like "both resolve an address private key directly, bypassing the external tx signer" covers both cases.

src/components/PasskeyOnboardingButton.js
```diff
@@ -0,0 +1,225 @@
...
+
+  const run = async (deriveWords) => {
+    setBusy(true);
+    try {
+      const { words, label, credentialId } = await deriveWords();
+      // The words exist only in this scope: derive the account xpub, persist it with the
+      // wallet metadata, and start the wallet read-only. No PIN, no stored secret.
+      const xpub = derivePasskeyXpub(words);
+      await STORE.initPasskeyStorage(xpub, {
+        passkeyLabel: label || 'Passkey wallet',
```
tuliomir: nitpick: these default names are user-facing but not translated

`'Passkey wallet'` here and `'Hathor Wallet'` at `:105` both reach the user: the label is rendered by PasskeyLockScreen ("Unlock with the passkey ...") and by `PasskeyXpubMismatchError`, and the name becomes the OS passkey displayName in the sign-in picker.

Wrapping both in `t` would bring them in line with the rest of the PR's i18n, which is otherwise complete.

src/screens/PasskeyLockScreen.js
```diff
@@ -0,0 +1,154 @@
...
+ * lock-screen mode. There is no PIN and no stored secret: unlocking runs a passkey ceremony,
+ * re-derives the xpub in memory and compares it to the stored one. Verification failure is
+ * never fatal — the user can always retry or fall back to resetting the wallet.
+ */
+const PasskeyLockScreen = () => {
+  const dispatch = useDispatch();
+  const wallet = useSelector((state) => state.wallet);
+  const [verifying, setVerifying] = useState(false);
+  const [error, setError] = useState(null);
+  // Ceremony must fire exactly once on mount even if the component re-renders (the OS shows
```
tuliomir: nitpick: the stated reason for `startedRef` is not what it guards

A `useEffect(..., [])` does not re-run on re-render, so "even if the component re-renders" is not the hazard.

What the ref actually protects against is React 18 StrictMode's double-mount and Fast Refresh remounts.

Small thing, but a maintainer who correctly reasons "empty deps already cover that" and deletes `startedRef` reintroduces a double credential sheet in dev.

src/components/PasskeyOnboardingButton.js
```diff
@@ -0,0 +1,225 @@
...
+    if (Platform.OS !== 'ios') return undefined;
+    const show = Keyboard.addListener('keyboardWillShow', (e) => setKbHeight(e.endCoordinates.height));
+    const hide = Keyboard.addListener('keyboardWillHide', () => setKbHeight(0));
+    return () => {
+      show.remove();
+      hide.remove();
+    };
+  }, []);
+
+  if (!enabled) return null;
```
tuliomir, line 72: praise: this toggle design avoids the rollback trap we hit before

Gating only wallet *creation* on the flag, and keying every runtime branch afterwards off the persisted `walletType` rather than the toggle, is the right call.

It is the opposite of what bit us in the amount-format work, where a toggle rollback stranded users with the behavior but no UI to change it. Here a rollback cannot leave an existing passkey user facing a PIN prompt they can never satisfy.

The i18n is also complete, for what it's worth: all new strings extracted to `.pot`, propagated to all three `.po` files, pt-br filled, and zero fuzzy entries across every catalog.

src/screens/SendConfirmScreen.js
raul-oliveira: tks

src/screens/PasskeyLockScreen.js
raul-oliveira: ok. I'll raise a discussion for it. tks

locale/pt-br/texts.po
```diff
@@ -1663,25 +1681,47 @@ msgstr "Transação"
...
+#, javascript-format
+msgid ""
+"This passkey opens a different wallet. Use the passkey named \"${ label }\"."
+msgstr ""
+"Essa passkey é de uma carteira diferente. Use a passkey \"${ label }\"."
+
+#: src/passkey/passkeySigner.js:46
+msgid ""
+"This passkey opens a different wallet. Use the passkey that created this "
+"wallet."
```
tuliomir, line 1730: suggestion(non-blocking): Translate this word
```suggestion
"carteira."
```

locale/pt-br/texts.po
```diff
@@ -1663,25 +1681,47 @@ msgstr "Transação"
...
+msgstr ""
+"Essa passkey é de uma carteira diferente. Use a passkey que criou essa wallet"
```
tuliomir: suggestion(non-blocking): Translate this word
```suggestion
"Essa passkey é de uma carteira diferente. Use a passkey que criou essa carteira"
```

PR #898, opened 2026-07-27
fix(decimal): render Reown/WalletConnect amounts with the network's decimal places

raul-oliveira:
Third slice of the amount-decimals fix, on top of #891 (merged to master). Extends the network `decimal_places` fix to the Reown/WalletConnect request screens. Independent of #894 — based directly on master.

### Acceptance Criteria
- Reown/WalletConnect request screens render token amounts using the network's `decimal_places` (from `serverInfo`) instead of the wallet-lib default of 2
- Covers: create-token request, get-utxos request, transaction fees, send-transaction request, and the nano-contract action/argument displays
- No behavior change on a 2-decimal network; the fix is observable only where the network reports a different precision

### Screenshots

image
image
image

https://github.com/user-attachments/assets/6dbe0011-355d-4f40-9867-2dd9fcc02a3b

### Security Checklist

### DIFF
--- a/src/components/Reown/NanoContract/NanoContractActions.js
+++ b/src/components/Reown/NanoContract/NanoContractActions.js
@@ -23,6 +23,7 @@ import { isTokenNFT, renderValue } from '../../../utils';
 import { AlertUI, COLORS } from '../../../styles/themes';
 import { WarnTextValue } from '../../WarnTextValue';
 import { CircleError } from '../../Icons/CircleError.icon';
+import { DEFAULT_ICON_SIZE } from '../../Icons/constants';
 import {
   getActionTitle,
   isAuthorityAction,
@@ -143,6 +144,8 @@ const ActionItem = ({ action, title, isNft, tokens, isTokenRegistered, showToken
     value: [commonStyles.text, commonStyles.field],
     addressSection: {
       marginTop: 8,
+      marginLeft: DEFAULT_ICON_SIZE + 16,
+      paddingRight: 16,
     },
     contentWrapper: {
       flex: 1,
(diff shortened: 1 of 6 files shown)


src/components/Reown/NanoContract/NanoContractActions.js
```diff
@@ -154,88 +157,91 @@ const ActionItem = ({ action, title, isNft, tokens, isTokenRegistered, showToken
...
+        {action.type !== NanoContractActionType.GRANT_AUTHORITY
+          && action.type !== NanoContractActionType.ACQUIRE_AUTHORITY
+          && action.amount != null && action.amount !== undefined && (
+            <Amount amount={action.amount} isNft={isNft} />
         )}
+      </View>
 
-        {/* ACQUIRE_AUTHORITY: Show only address (send the authority and create the output) */}
-        {action.type === NanoContractActionType.ACQUIRE_AUTHORITY && action.address && (
+      {/* WITHDRAWAL: Show only address (address to send the amount and create the output) */}
```
tuliomir, line 194: suggestion(non-blocking): the layout restyle isn't covered by the acceptance criteria

This action item is restructured (addresses moved out of the icon row, indented by `DEFAULT_ICON_SIZE + 16`), and four files gain `alignItems: 'flex-start'` / `textAlign: 'right'` / `maxWidth`. Those render differently on a 2-decimal network too, so AC bullet 3 ("no behavior change on a 2-decimal network") no longer holds. Presumably it's making room for longer amounts — worth saying so in the description, so a reviewer knows to check the layout at the old precision as well.

src/components/Reown/NanoContract/NanoContractActions.js
```diff
@@ -249,13 +255,16 @@ const ActionItem = ({ action, title, isNft, tokens, isTokenRegistered, showToken
  * @param {boolean} props.isNft
  */
 const Amount = ({ amount, isNft }) => {
-  const amountToRender = renderValue(amount, isNft);
+  const decimalPlaces = useSelector((state) => state.serverInfo?.decimal_places);
```
tuliomir, line 258: question(non-blocking): should these amounts honour the compressed/expanded preference?

`renderValue`'s fourth argument is the wallet-wide amount format added in PR 891; `TxDetailsModal` passes `getDisplayAmountFormat(state)`, the Reown screens don't, so they stay EXPANDED. Deliberate — full precision on a screen you're signing from — or just out of this slice's scope?

